PR #2, opened 2024-02-12
theme: Add vscode-dark-plus

d1y:

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -1,3 +1,7 @@
 [beancount]
 path = "extensions/beancount"
 version = "0.0.1"
+
+[vscode-dark-plus]
+path = "extensions/vscode-dark-plus"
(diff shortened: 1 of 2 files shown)


extensions.toml
```diff
@@ -1,3 +1,7 @@
 [beancount]
 path = "extensions/beancount"
 version = "0.0.1"
+
+[vscode-dark-plus-theme]
```
maxdeviant, line 5: Let's just call this `vscode-dark-plus`.

extensions/vscode_dark_plus.zed
maxdeviant, line 1: Can we make this directory name kebab-cased?

```
extensions/vscode-dark-plus
```

d1y:
OK, force push

PR #3, opened 2024-02-12
Add git_rebase highlight extenstion

d1y:
https://github.com/zed-industries/zed/pull/7222

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "extensions/beancount"]
 	path = extensions/beancount
 	url = https://github.com/zed-extensions/beancount
+[submodule "extensions/git_rebase"]
+	path = extensions/git_rebase
+	url = git@github.com:d1y/git_rebase.zed.git
--- a/extensions.toml
+++ b/extensions.toml
@@ -1,3 +1,7 @@
 [beancount]
 path = "extensions/beancount"
 version = "0.0.1"
+
+[git_rebase]
+path = "extensions/git_rebase"
+version = "0.0.1"


d1y:
Okay, it seems that Zed will implement these features in the future. I look forward to better git (VCS) support in Zed and will close this PR ❤️

SomeoneToIgnore:
I was more interested in the actual discussion and to know whether there are better design approaches, but this way will work too, I guess.
Thank you for the attempt.

ck37:
## Update: Recommend Migration to tree-sitter-quarto

### Current Issue with tree-sitter-pandoc-markdown

We've identified a fundamental limitation with the dual grammar (block + inline with injection) approach: **headings containing hyphens don't receive heading color**.

**Root cause:** When the inline grammar injects into heading content, it parses hyphens creating text nodes with no scope, which lose the parent heading color. The inline grammar has no context about being inside a heading due to tree-sitter injection isolation.

**Attempted fixes:** Multiple approaches failed (see `docs/hyphenated-headings-issue.md`):
- Capturing inline children in block grammar → blocks injection
- Adding text captures in inline grammar → overrides all text colors
- Excluding headings from injection → breaks bold/italic in headings
- Scope and priority adjustments → doesn't solve injection context loss

This is an **architectural limitation** of using separate grammars with injection.

### tree-sitter-quarto Status: Ready for Integration

Checking the latest tree-sitter-quarto repo, the grammar is now **production-ready**:

✅ **parser.c generated** (was missing before)
✅ **58/58 tests passing** (100% coverage)  
✅ **WASM compilation verified**
✅ **All core Quarto features implemented**
✅ **Uses standard `@markup.*` scopes** (can be remapped to Zed's `@text.*`)

**Key advantage:** tree-sitter-quarto uses a **unified grammar** (not dual grammar), which avoids the injection context loss issue entirely.

### Recommendation

**Switch to tree-sitter-quarto** to:
1. Fix the hyphenated heading issue
2. Get proper Quarto-specific parsing (chunk options, cross-references, callouts)
3. Use a grammar maintained specifically for Quarto

### Migration Steps

1. Update `extension.toml` to point to tree-sitter-quarto repo
2. Create scope remapping from `@markup.*` to `@text.*` (Zed compatibility)
3. Import/adapt query files (highlights, injections, folds)
4. Test with existing fixtures
5. Update documentation

**Estimated effort:** 2-4 hours for initial integration, 1-2 days for testing and refinement.

### Decision Point

Should we:
- **Option A:** Continue trying to fix hyphenated headings in dual grammar (high complexity, uncertain success)
- **Option B:** Switch to tree-sitter-quarto (cleaner solution, better long-term)

I recommend **Option B**. The grammar is ready, and we'll get better Quarto support overall.

Thoughts?

PR #6, opened 2024-02-14
Dracula theme

Otterpocket:
Adds Dracula theme


Otterpocket:
The repo's top level directory isn't extensions/ so my assumption is that I either have to restructure https://github.com/dracula/zed so that the extension.json is in the top level dir or tell the submodule/toml to look there?

maxdeviant:
@Otterpocket Thank you!

PR #9, opened 2024-02-14
Update extensions.toml for Dracula

Otterpocket:
Correct directory structure


maxdeviant:
@Otterpocket We'll need to bump the submodule's commit as well, so that we pull down the version with the right structure.

PR #11, opened 2024-02-14
Update `catppuccin` theme

ssaunderss:
Updated the git submodule after finishing porting / polishing the rest of the catppuccin themes.


ssaunderss (commented):
revert version update

ssaunderss:
@JosephTLyons @maxdeviant I reverted the version bump and it looks like the CICD is passing now. What's the process for versioning extensions?

maxdeviant:
@ssaunderss Hmm, you should just need to bump the version in `extensions.toml` and in the `extension.json` in the theme repository.

ssaunderss:
@maxdeviant that was it, thanks!

PR #46, opened 2024-02-17
msun-dark theme

mikesun:
Minimalist dark theme for loosely based on the [rsms Sublime Text theme](https://github.com/rsms/sublime-theme).

https://github.com/mikesun/msun-dark-zed


JosephTLyons:
Not quite understanding why it isn't liking being alphabetized here...

JosephTLyons:
Awesome! Let's give it a shot! Sorry for the friction here, we will improve things at some point where it won't require pull requests and such to add extensions.

PR #50, opened 2024-02-17
Add Synthwave extension

DanielMSchmidt:


DanielMSchmidt:
@maxdeviant I added an author field, the `theme_importer` did not emit it unfortunately :D

maxdeviant:
@DanielMSchmidt It looks like the structure of the theme file still isn't quite right.

It looks like the theme is top-level instead of being nested in the `themes` array (a theme file can contain multiple themes).

If you add this field to the top-level of the theme file it should give you JSON schema validation warnings:

```json
"$schema": "https://zed.dev/schema/themes/v0.1.0.json"
```

Here's an example of what the theme structure looks like: https://github.com/zed-industries/zed/blob/db0eaca2e5a1bbea2bf0192429ab7f3320e38a60/assets/themes/one/one.json

JosephTLyons:
Hey @DanielMSchmidt, I opened up a PR to help here. This should resolve the schema errors.

https://github.com/DanielMSchmidt/zed-synthwave/pull/1

PR #52, opened 2024-02-18
Update msun-dark theme with new msun-dark-6000k variation

mikesun:


JosephTLyons:
Hey @mikesun, I think we will want to bump the version number on this extension, in both of these places, so that it will be flagged as an upgrade to anyone who has downloaded it:

https://github.com/mikesun/msun-dark-zed/blob/103d03a691c6a9156f7460f9fcb9c1cdc22c56a0/extension.json#L3

---

https://github.com/zed-industries/extensions/blob/20f73eaeeaa6653dc8a32d822a0baa6eca4433dc/extensions.toml#L35

PR #56, opened 2024-02-18
Update macos-classic theme for fix error, warning, info color in dark mode

huacnlee:
https://github.com/huacnlee/zed-theme-macos-classic/issues/1


maxdeviant:
@huacnlee Thanks!

PR #57, opened 2024-02-18
Java support for Zed

samuser107:


samuser107:
@cla-bot check

extensions/java/README.md
maxdeviant, line 1: @samuser107 The extension should be provided by a separate repository and registered as a Git submodule in the extensions repository.

We don't want to include the source for the extension inline like this.

See the docs on publishing extensions: https://github.com/zed-industries/extensions/blob/main/AUTHORING_EXTENSIONS.md#publishing-your-extension

maxdeviant (changes requested):
The extension needs to referenced via a Git submodule.

maxdeviant:
@samuser107 Hmm, the submodule reference doesn't look quite right.

There should be a separate repository containing just the contents of the `java` extension, and that is what needs to be added as a Git submodule.

For example, here's how we do it with our Nushell extension: https://github.com/zed-extensions/nu

samuser107:
I think it's alright now

maxdeviant:
Looks like we're failing to build the WASM grammar now.

Is there a reason you went with https://github.com/serenadeai/java-tree-sitter as the Tree-sitter grammar instead of this one from the Tree-sitter org?

https://github.com/tree-sitter/tree-sitter-java

maxdeviant:
> Looks like we're failing to build the WASM grammar now.
> 
> Is there a reason you went with https://github.com/serenadeai/java-tree-sitter as the Tree-sitter grammar instead of this one from the Tree-sitter org?
> 
> https://github.com/tree-sitter/tree-sitter-java

Actually, it looks like https://github.com/serenadeai/java-tree-sitter isn't even a Tree-sitter grammar for Java, it's Java bindings _to_ Tree-sitter.

maxdeviant:
Could we also change the extension name to just "Java"?

https://github.com/samuser107/zed-java-extension/blob/2482bd4760c39a275a003e70e8a1bb272cf9d99c/extension.json#L2

The "for Zed" isn't necessary, as this is a Zed extension.

samuser107:
> > Looks like we're failing to build the WASM grammar now.
> > Is there a reason you went with https://github.com/serenadeai/java-tree-sitter as the Tree-sitter grammar instead of this one from the Tree-sitter org?
> > https://github.com/tree-sitter/tree-sitter-java
> 
> Actually, it looks like https://github.com/serenadeai/java-tree-sitter isn't even a Tree-sitter grammar for Java, it's Java bindings _to_ Tree-sitter.

yes sorry for that. fixed it with the name too

maxdeviant (approved):
Thanks for working through this!

PR #58, opened 2024-02-18
Add log extension

evrsen:


evrsen:
despite the problems with testing extensions locally, this extension seems to be valid and works as expected when directly integrated to zed's src code. I would assume it's safe to merge this extension now

maxdeviant:
@evrsen Thanks for working through it!

PR #70, opened 2024-02-20
Improve package-extensions script to update extension by `extensions.toml`.

huacnlee:
Now every extension version update, we just need to update `extensions.toml`.

1. Modify `extensions.toml` to special the new version or a git commit.
2. Create a Pull Request.
3. The `package-extensions.js` will update the submodule by `extensions.toml`.


maxdeviant:
Hey @huacnlee, this isn't something we plan to merge.

We want to be able to know the current state of the extension registry just by reading the `extensions.toml`, so we don't want to have to checkout the Git submodules unless we need to publish a new version.

PR #104, opened 2024-02-23
Bump nu extension to 0.0.2

JosephTLyons:


fdncred:
@JosephTLyons any updates here?

maxbrunsfeld:
Sorry for the delay here. The CI failure is caused by a bug in Tree-sitter itself, where certain grammars take too much time and memory to compile to WASM. Something about the Nu grammar triggers this condition. We'll fix the issue in Tree-sitter when we get a chance.

fdncred:
Thanks @maxbrunsfeld for the update. If you discover something that we should do differently in the grammar, please let us know at https://github.com/nushell/tree-sitter-nu. We'd like to continue to be included with zed.

maxbrunsfeld:
@fdncred I don't fully understand the problem yet, but it may have to do with [this `_unquoted_rule`](https://github.com/nushell/tree-sitter-nu/blob/c5b7816043992b1cdc1462a889bc74dc08576fa6/grammar.js#L1396-L1500) helper. The regexes that it generates aren't currently handled very efficiently by Tree-sitter: it generates a lot of helper functions for matching them, because there are many character sets to match that are all slightly different. The relevant code that we need to improve in Tree-sitter's code generation is in [this area](https://github.com/tree-sitter/tree-sitter/blob/master/cli/src/generate/render.rs#L736-L755).

maxbrunsfeld:
Did some work on Tree-sitter to help with this: https://github.com/tree-sitter/tree-sitter/pull/3234

PR #106, opened 2024-02-23
try fix 0.0.4 swift

samuser107:


maxbrunsfeld:
@samuser107 Sorry for the confusion here, this isn't your fault, there seems to be a bug in Zed (actually, Tree-sitter) that is triggered by the Swift parser. I am investigating right now. In the meantime, we are shipping a patch release of Zed that will prevent this faulty extension from being loaded.

PR #129, opened 2024-02-25
Bump base16 to v0.0.3

bswinnerton:
This commit bumps the base16 theme to `v0.0.3`, which adds specific colors for:

- The background of tabs when they are being dragged
- The heading of search results
- Predictive text

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -4,7 +4,7 @@ version = "0.0.1"
 
 [base16]
 path = "extensions/base16"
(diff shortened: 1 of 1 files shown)


maxdeviant:
@bswinnerton We also need to bump the version in `extensions.toml`.

bswinnerton:
@maxdeviant good catch! Thank you 🙇‍♂️

PR #130, opened 2024-02-26
Helm (`helm`) extension support

hjoshi123:
I'm fairly new to language configuration, lsp and tree sitter and was working on Helm charts so decided to create an extension for go templates.

Kindly let me know if something is missing.


hjoshi123:
@cla-bot check

hjoshi123:
@maxdeviant should the repo name for grammar be tree-sitter-helm? because its not that

maxdeviant:
> @maxdeviant should the repo name for grammar be tree-sitter-helm? because its not that

Hmm, I'm not 100% sure how the name of the resulting Wasm grammar is determined.

@maxbrunsfeld Do you know?

hjoshi123:
@maxbrunsfeld how should I handle this? Should I change the name of the extension and folders so that it matches the repository of tree sitter?

.gitmodules
```diff
@@ -85,6 +85,10 @@
 [submodule "extensions/graphql"]
 	path = extensions/graphql
 	url = https://github.com/11bit/zed-extension-graphql.git
+  
+[submodule "extensions/go-template"]
+	path = extensions/go-template
+	url = https://github.com/hjoshi123/helm.zed.git
```
maxbrunsfeld, line 91: Would you mind naming the submodule folder `helm`? Also, in your extension, let's call the language `Helm` instead of `go-template`. Though it uses `go-template` for the grammar, the fact that YAML is injected seems to be Helm-specific (not just general Go template support).

.gitmodules
```diff
@@ -312,4 +316,4 @@
 
 [submodule "extensions/zedspace"]
 	path = extensions/zedspace
-	url = https://github.com/Brunowilliang/zedspace.git
+	url = git@github.com:Brunowilliang/zedspace.git
```
maxbrunsfeld, line 319: Can you revert this submodule URL change? We need to use https for all submodules for our CI.

extensions.toml
```diff
@@ -94,6 +94,10 @@ version = "0.0.2"
 path = "extensions/horizon"
 version = "0.0.1"
 
+[go-template]
```
maxbrunsfeld, line 97: Let's call this `helm`.

maxbrunsfeld (changes requested):

maxdeviant:
> @hjoshi123 please do the needful

@cabrinha Please stop spamming this PR with messages like this. At this point we're aware of your desire for this extension.

Comments like this are just noise and are not doing anything in terms of making this work proceed any faster. If you would like to see this work proceed faster, consider helping out with the necessary changes.

maxdeviant:
Closing due to inactivity.

PR #216, opened 2024-02-26
Bump Ultimate Dark Neo to 0.0.2

rubjo:


rubjo:
Hm not sure why this still fails looking for 0.0.1, theme repo is updated.

Imgkl:
@rubjo the submodule is still pointing to the old version. 

The check commit id under `https://github.com/rubjo/extensions/tree/main/extensions`. That's needs to be updated to the latest commit.

PR #221, opened 2024-02-26
Update Swift to 0.1.0 to fix crash

samuser107:


samuser107:
@maxdeviant Is it normal that the extension cannot be used even after this update that is supposed to fix the crash experienced?

maxdeviant:
> @maxdeviant Is it normal that the extension cannot be used even after this update that is supposed to fix the crash experienced?

Are you still seeing the crash with the latest version of the Swift extension?

samuser107:
Okay! Let me know if i can help and when it will be available again!

PR #231, opened 2024-02-27
Add caius/zed-make for Makefile support

caius:
Initial attempt, allows `Makefile` language to show up and hard tabs be configured in the settings to make editing the files easier. Doesn't contain any syntax highlighting yet as I can't get a valid `highlights.scm` file in place so far.

Haven't looked if the language toml or extension toml can set hard tabs or tab width settings, it's just in the readme for now for people to add to their Zed settings by hand.

This is extracted from https://github.com/zed-industries/zed/pull/7402 as that didn't land before Extensions came into the world 😻

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -151,3 +151,6 @@
 [submodule "extensions/zedspace"]
 	path = extensions/zedspace
 	url = git@github.com:Brunowilliang/zedspace.git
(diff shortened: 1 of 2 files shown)


maxdeviant:
@caius Thank you!

PR #246, opened 2024-02-28
Add Dotenv extension

evrsen:


evrsen:
Ah sorry, the tree-sitter seems to not even be a working one, I'll reopen this issue when there exists a working tree-sitter grammar for dotenv files.

mrnugget:
Just so you're aware: we already treat `.env` as Bash files here https://github.com/zed-industries/zed/blob/f8959834c4fec9a4848e008af810fd32feb42881/crates/languages/src/bash/config.toml#L3 Shouldn't block anything you're doing, but wanted to call it out because I discovered that yesterday :)

evrsen:
aha that's interesting, thanks for sharing! I'm not really familiar with bash, are .env files a subset of bash or is this a quick hack that just works for .env files for the time being? if the former then an external extension for this might not be needed at all 🤷 I was thinking of doing something similar to the ``.env`` plugin the JetBrains IDEs have, well, that's when zed adds more capabilities to extensions.

mrnugget:
As far as I know `.env` files are `KEY=value` lines that can contain bash-y variable substitution, so treating as Bash seems like a safe bet to me for now.

PR #276, opened 2024-03-02
New One Dark Pro Theme

EvanZhouDev:
I have completely redesigned the One Dark Pro theme, starting from Zed's original One Dark theme's JSON.
No theme porter was used, and each value was manually adjusted to integrate the One Dark Pro theme from VSCode into Zed. It was created to maintain the modern background of Zed but with the beautiful, contrasting syntax theme from the VSCode One Dark Pro theme.

Here's what it looks like.

one-dark-pro-demo

This theme has already been created by @MordFustang21. However, when it was ported using the Theme Porter, some issues remained despite its wonderful design.

I started my current implementation off, as mentioned, Zed's own One Dark, instead of @MordFustang21's, and thus I have created a new Extension repo, instead of simply contributing to his.

If @MordFustang21 would like me to create a new extension instead of overriding his, I would be glad to, but I believe that this extension has been more adapted for VSCode One Dark Pro users. I will also be glad to maintain this in the future.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -152,7 +152,7 @@
 
 [submodule "extensions/one-dark-pro"]
 	path = extensions/one-dark-pro
-	url = git@github.com:MordFustang21/zed-one-dark-pro.git
+	url = git@github.com:EvanZhouDev/zed-one-dark-pro.git
 
 [submodule "extensions/one-hunter"]
 	path = extensions/one-hunter
--- a/extensions.toml
+++ b/extensions.toml
@@ -152,7 +152,7 @@ version = "0.0.1"
 
 [one-dark-pro]
 path = "extensions/one-dark-pro"
-version = "0.0.2"
+version = "0.0.1"
 
 [one-hunter]
 path = "extensions/one-hunter"


EvanZhouDev:
@cla-bot check

EvanZhouDev:
@maxdeviant @MordFustang21 What do you think?

Moshyfawn:
My 2 cents: I'd rather see a contribution to the already referenced repo that fixes the existing problems, rather than just removing the original contribution.

If/when the extensions store is created, you'll be able to create your own standalone extensions, even if they extend existing themes/functionality, but for now I'd like to see open source guidelines followed.

Also, even with the extension store, I'd prefer to avoid duplicating existing extensions, at least to avoid complications for the end user.

EvanZhouDev:
@Moshyfawn I do agree with your points. However, I have completely redesigned the extension, sharing with the original only the idea, and have hardly based it on the original repo. Thus, a PR into the extension doesn't make a ton of sense, especially given the fact that the original was simply a theme port, while I manually designed mine.

MordFustang21:
@EvanZhouDev What you've described above sounds more like a remake. `I have completely redesigned the extension, sharing with the original only the idea`. Now if you were porting the existing theme and trying to make it as compatible as possible then sure you might have a case but then you'd be better served improving the existing extension instead of trying to nuke someone else's extension in favor of your own.

 Also going to this comment `given the fact that the original was simply a theme port, while I manually designed mine.` It's not strictly based on the theme importer. There's been quite a few changes made and more are needed, the theme importer isn't perfect.
 
 Overall I agree with @Moshyfawn upstream fixes should be preferred and duplicate themes should be avoided unless they're substantially different which yours may be.

maxdeviant:
@EvanZhouDev It seems like there are two options here:

1. Upstream fixes/changes to the current theme
2. Publish your version of One Dark Pro under a different name

maxdeviant:
> @EvanZhouDev It seems like there are two options here:
> 
> 1. Upstream fixes/changes to the current theme
> 2. Publish your version of One Dark Pro under a different name

Going to close this in favor of one of the options cited above.

PR #284, opened 2024-03-03
Add Modest Dark theme

timcole:
A modest zed theme based on one dark but with brighter colours on a darker background


maxdeviant:
@timcole Thank you!

PR #286, opened 2024-03-04
Add base16-zed extension

timcki:
The existing base16 extension is limited to 2 themes and lacks the ability to add new themes using a template. This PR adds support for all standard base16 themes and a template to allow for adding new ones.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -253,3 +253,6 @@
 [submodule "extensions/zedspace"]
 	path = extensions/zedspace
 	url = git@github.com:Brunowilliang/zedspace.git
+[submodule "extensions/base16-zed"]
+	path = extensions/base16-zed
+	url = git@github.com:timcki/base16-zed.git
--- a/extensions.toml
+++ b/extensions.toml
@@ -6,6 +6,10 @@ version = "0.0.1"
 path = "extensions/base16"
 version = "0.0.4"
 
+[base16-zed]
+path = "extensions/base16-zed"
+version = "0.0.1"
+
 [beancount]
 path = "extensions/beancount"
 version = "0.0.1"


maxdeviant:
@timcki Have you checked with the author of the base16 package to see if there is interest in extending that one with the themes you're missing?

timcki:
@maxdeviant I did consider contributing to the existing base16 package repository however in that repo there were only 2 themes added by 2 different contributors (not based on templates). I decided it's easier to start a new project following other base16 implementations (like here for example: https://github.com/aarowill/base16-alacritty).
My concern was that by submitting a pull request to the existing repository, I would essentially override both of the themes that already existed and add new unrelated folders.

maxdeviant:
> @maxdeviant I did consider contributing to the existing base16 package repository however in that repo there were only 2 themes added by 2 different contributors (not based on templates). I decided it's easier to start a new project following other base16 implementations (like here for example: https://github.com/aarowill/base16-alacritty). My concern was that by submitting a pull request to the existing repository, I would essentially override both of the themes that already existed and add new unrelated folders. EDIT: I realised I didn't address the main question; my main motivation wasn't only to add the one base16 variant I was missing but to provide all the other color configuration people might expect from a `base16` extension.

I don't want to speak on behalf of @bswinnerton, but I think given that the existing `base16` extensions is meant to be a port of base16, I would suspect that any changes to bring more parity in that direction would be welcome.

But I feel pretty strongly that at this stage we shouldn't have two extensions that are both looking to provide the base 16 theme that will just leave Zed users confused as to which one they should use.

bswinnerton:
Hi @timcki! I'd be happy to support a more template based approach in the base16-zed repo. I had started off with a manual approach that was easier to fine tune but as more themes are added I'm sure we'll need a more holistic approach. 

Happy to add you as a collaborator on that repo so you can move fast and we can have a central Zed extension for all of base16.

timcki:
Hi @bswinnerton thanks for responding (: I'd be happy to open the same pr in your repo!

PR #289, opened 2024-03-04
reservation for xcode-color-scheme extension

evrsen:


evrsen:
Yep have seen it and tried it, none of the themes it provides matches the default colour scheme of Xcode to 100%, so I'd like to maintain my own \:)

PR #293, opened 2024-03-04
reservation for xcode-text-fonts extension

evrsen:
This PR is for reserving a Xcode Text Fonts extension for Zed when https://github.com/zed-industries/zed/issues/8838 is stabilized.


maxdeviant:
We don't need to squat package names on the registry right now.

PR #295, opened 2024-03-04
reservation for xcode-color-schemes extension

evrsen:


maxdeviant:
We don't need to squat package names on the registry right now.

PR #306, opened 2024-03-06
Add Ruby Extension

rzukic:
Add Ruby Extension
Bildschirmfoto 2024-03-05 um 08 11 01


huacnlee:
We are already have built in Ruby support in Zed.

PR #307, opened 2024-03-06
Add git-firefly

d1y:
Prev: [#7048](https://github.com/zed-industries/zed/issues/7048), Let's roll git highlighting back into the extension

> Because tree-sitter-gitcommit slows down the compilation of zed 😅 (https://github.com/zed-industries/zed/pull/7147#issuecomment-1937722323 / https://github.com/zed-industries/zed/issues/7048#issuecomment-1979620553)

Support syntax highlighting 

- git_commit
- git_config
- git_rebase
- git_diff(.diff)
- gitignore
- gitattributes

cc @williamdes @SomeoneToIgnore

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -54,6 +54,10 @@
 	path = extensions/gdscript
 	url = https://github.com/grndctrl/zed-gdscript
 
+[submodule "extensions/git-firefly"]
+	path = extensions/git-firefly
+	url = git@github.com:d1y/git_firefly.git
+
 [submodule "extensions/github-dark-default"]
 	path = extensions/github-dark-default
 	url = git@github.com:MordFustang21/zed-github-dark.git
@@ -265,6 +269,3 @@
 [submodule "extensions/zedspace"]
 	path = extensions/zedspace
 	url = git@github.com:Brunowilliang/zedspace.git
-[submodule "extensions/git-firefly"]
-	path = extensions/git-firefly
-	url = git@github.com:d1y/git_firefly.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -265,3 +265,6 @@
 [submodule "extensions/zedspace"]
 	path = extensions/zedspace
 	url = git@github.com:Brunowilliang/zedspace.git
+[submodule "extensions/git-firefly"]
+	path = extensions/git-firefly
+	url = git@github.com:d1y/git_firefly.git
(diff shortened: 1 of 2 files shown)


d1y:
Here's a fun fact: I didn't build wasm locally and debug it, because I got stuck compiling git_commit.wasm, which was really slow 😢
Repo: https://github.com/d1y/git_firefly
And, when the extension is completed, the gitcommit related in zed can be deleted, but before that, we need to ensure that the extension can work

JosephTLyons:
Excited for this one!

williamdes:
Will that mean that the git highlight will not be needed in the main repo ?

d1y:
The CI action run failed, Is this because compilation is too slow 🤔? @maxdeviant 

image

d1y:
> Will that mean that the git highlight will not be needed in the main repo ?

Yep. Falling back to extensions is a good option :)

d1y:
Next. Can we merge it?

maxdeviant:
> The CI action run failed, Is this because compilation is too slow 🤔? @maxdeviant
> 
> image

It seems like it's getting stuck building the `git_commit` grammar.

When I run it locally it starts building `git_commit` and just never seems to complete.

It could be some issue with the grammar itself?

SomeoneToIgnore:
It is actually able to complete building it, but takes an eternity.
So, technically, no issues, just a very peculiar build process.

maxdeviant:
> It is actually able to complete building it, but takes an eternity. So, technically, no issues, just a very peculiar build process.

Ah, I probably just haven't let it run long enough.

Although it still seems like a potential problem with the grammar if it takes that long.

SomeoneToIgnore:
It is, and this PR masks it by moving it away out of Zed repo to here, so if we fail to find a proper one of fix this grammar, I would still love to move on with this PR somehow, to move things away out of Zed source tree.

d1y:
The helix git-commit used https://github.com/the-mikedavis/tree-sitter-git-commit

I recommend trying this?

https://github.com/helix-editor/helix/blob/b93fae9c8b955e11f427979134e3494294e8e2e0/languages.toml#L1529-L1540

d1y:
Pong! The CI run successfully

> I used https://github.com/the-mikedavis/tree-sitter-git-commit

JosephTLyons:
Super nice to see this, and super meta!

SCR-20240307-mllc

PR #313, opened 2024-03-07
Stay build In order to local build

d1y:
Convenient to observe the generated files for debugging


maxdeviant:
Thanks, but I don't think we really want this as a permanent fixture.

If you're debugging locally you can just remove the part where we delete the build directory.

PR #327, opened 2024-03-11
Add schema for icon extensions

evrsen:
This PR proposes a schema for icon extensions (https://github.com/zed-industries/zed/issues/8843), I'm not sure if this is done correctly and probably is missing some important configuration options, would be glad to get some feedback and suggestions :)


maxdeviant:
We aren't using JSON schema to validate extension files anymore.

We can revisit the schema when we pick up work on providing icons via extensions.

PR #330, opened 2024-03-12
colorizer theme for zed

tamimhasandev:
I added **colorizer** theme for **zed** which will help users to write better code with a better look


tamimhasandev:
I have configured my mail with github please check

PR #335, opened 2024-03-14
colorizer theme for zed

tamimhasandev:
I created a new theme called Colorizer for Zed


maxdeviant:
@tamimhasandev Thank you!

PR #361, opened 2024-03-22
Add initial Unison support

zetashift:
Let me know if anything is missing!

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -309,6 +309,3 @@
 [submodule "extensions/zedspace"]
 	path = extensions/zedspace
 	url = https://github.com/Brunowilliang/zedspace.git
-[submodule "extensions/unison-zed"]
-	path = extensions/unison-zed
-	url = https://github.com/zetashift/unison-zed
--- a/.gitmodules
+++ b/.gitmodules
@@ -301,3 +301,6 @@
 [submodule "extensions/zedspace"]
 	path = extensions/zedspace
 	url = https://github.com/Brunowilliang/zedspace.git
+[submodule "extensions/unison-zed"]
+	path = extensions/unison-zed
+	url = https://github.com/zetashift/unison-zed
--- a/.gitmodules
+++ b/.gitmodules
@@ -309,3 +309,7 @@
 [submodule "extensions/zedspace"]
 	path = extensions/zedspace
 	url = https://github.com/Brunowilliang/zedspace.git
+
+[submodule "extensions/unison"]
+	path = extensions/unison
+	url = https://github.com/zetashift/unison-zed
--- a/extensions.toml
+++ b/extensions.toml
@@ -274,8 +274,8 @@ version = "0.0.2"
 path = "extensions/ultimate-dark-neo"
 version = "0.0.2"
 
-[unison-zed]
-path = "extensions/unison-zed"
+[unison]
+path = "extensions/unison"
 version = "0.0.1"
 
 [vesper]
--- a/extensions.toml
+++ b/extensions.toml
@@ -481,6 +481,7 @@ submodule = "extensions/ultimate-dark-neo"
 version = "0.0.4"
 
 [unison]
+submodule = "extensions/unison"
 path = "extensions/unison"
 version = "0.0.1"
 
--- a/extensions.toml
+++ b/extensions.toml
@@ -555,7 +555,6 @@ version = "0.0.4"
 
 [unison]
 submodule = "extensions/unison"
-path = "extensions/unison"
 version = "0.0.1"
 
 [unoflat]
--- a/extensions.toml
+++ b/extensions.toml
@@ -270,6 +270,10 @@ version = "0.0.2"
 path = "extensions/ultimate-dark-neo"
 version = "0.0.2"
 
+[unison-zed]
+path = "extensions/unison-zed"
+version = "0.0.1"
+
 [vesper]
 path = "extensions/vesper"
 version = "0.0.1"


zetashift:
@cla-bot check

zetashift:
@cla-bot check

maxdeviant:
@zetashift Can we drop the `-zed` suffixes from the extension ID and name?

Since these are all Zed extensions, there isn't a need to specify.

https://github.com/zetashift/unison-zed/blob/92ebe2290a432d4312c98e5e1b42b8d7237bb3a9/extension.toml#L1-L2

It should be:

```toml
id = "unison"
name = "Unison"
```

zetashift:
@maxdeviant I've updated the `extension.toml`: https://github.com/zetashift/unison-zed/blob/main/extension.toml
Do I have to do anything else to update the `git submodule`?

maxdeviant:
> @maxdeviant I've updated the `extension.toml`: https://github.com/zetashift/unison-zed/blob/main/extension.toml Do I have to do anything else to update the `git submodule`?

You need to bump the submodule commit to the new one.

Typically this is done by going into the submodule directory (`extensions/unison`, in this case), and then checking out `main` / `master`, and then you should see a commit change in the `extensions` repo that needs to be committed.

zetashift:
> > @maxdeviant I've updated the `extension.toml`: https://github.com/zetashift/unison-zed/blob/main/extension.toml Do I have to do anything else to update the `git submodule`?
> 
> You need to bump the submodule commit to the new one.
> 
> Typically this is done by going into the submodule directory (`extensions/unison`, in this case), and then checking out `main` / `master`, and then you should see a commit change in the `extensions` repo that needs to be committed.

Ah thank you! I also saw that I named the path wrong, e.g. it should be `unison` and not `unison-zed`, though that makes me wonder what would happen if another unison extensions pops up?

maxdeviant:
> > > @maxdeviant I've updated the `extension.toml`: https://github.com/zetashift/unison-zed/blob/main/extension.toml Do I have to do anything else to update the `git submodule`?
> > 
> > 
> > You need to bump the submodule commit to the new one.
> > Typically this is done by going into the submodule directory (`extensions/unison`, in this case), and then checking out `main` / `master`, and then you should see a commit change in the `extensions` repo that needs to be committed.
> 
> Ah thank you! I also saw that I named the path wrong, e.g. it should be `unison` and not `unison-zed`, though that makes me wonder what would happen if another unison extensions pops up?

Could you say more about this? I'm not sure I follow.

I would think any improvements to the Unison extension should go towards your extension, as opposed to creating new ones.

zetashift:
> I would think any improvements to the Unison extension should go towards your extension, as opposed to creating new ones.

Ah this was exactly what my question was about. I'll probably share that along then.

```
node:internal/errors:932
  const err = new Error(message);
              ^

Error: Command failed: ./zed-extension --scratch-dir ./scratch --source-dir extensions/unison --output-dir output
[2024-03-25T15:49:32Z INFO  zed_extension] loading extension manifest
[2024-03-25T15:49:32Z INFO  zed_extension] compiling extension
[2024-03-25T15:49:32Z INFO  extension::extension_builder] downloading wasi-sdk to /home/runner/work/extensions/extensions/scratch/wasi-sdk
[2024-03-25T15:49:35Z INFO  extension::extension_builder] checking out unison parser
[2024-03-25T15:49:35Z INFO  extension::extension_builder] compiling unison parser
[2024-03-25T15:49:57Z INFO  extension::extension_builder] finished compiling extension /home/runner/work/extensions/extensions/extensions/unison
Error: Failed to instantiate wasm module: invalid import 'strcat'


    at ChildProcess.exithandler (node:child_process:422:12)
    at ChildProcess.emit (node:events:518:28)
    at maybeClose (node:internal/child_process:1105:16)
    at ChildProcess._handle.onexit (node:internal/child_process:305:5) {
  code: 1,
  killed: false,
  signal: null,
  cmd: './zed-extension --scratch-dir ./scratch --source-dir extensions/unison --output-dir output'
}

Node.js v20.11.1
 ELIFECYCLE  Command failed with exit code 1.
 ```
 I'm not sure what this error is about, is my extension configured wrong for wasm building?

**Comment by maxdeviant** [comment]
> https://github.com/zed-industries/extensions/actions/runs/8392887233/job/23063430084?pr=361#step:9:19
>  I'm not sure what this error is about, is my extension configured wrong for wasm building?

It's failing due to this error:

> Error: Failed to instantiate wasm module: invalid import 'strcat'

We don't expose `strcat` to Wasm builds of Tree-sitter grammars. We're going to add `strncat`, and then the grammar will need to be updated to use that instead.

**Comment by maxbrunsfeld** [comment]
We opened an issue on the Unison grammar repo: https://github.com/kylegoetz/tree-sitter-unison/issues/57.

**Comment by maxdeviant** [comment]
> > We're going to add `strncat`
> 
> Is there an issue or something to see when this lands in tree-sitter, and when Zed uses that specific version? Else I might forget this PR :P

Tree-sitter already supports `strncat` on the latest `master`: https://github.com/tree-sitter/tree-sitter/commit/0fc92c9a7d0ddb417bd74bf7f533bb8f3042dbe3

Right now we're blocked on the upstream grammar resolve this issue: https://github.com/kylegoetz/tree-sitter-unison/issues/57 (or if you want to fork it to make the necessary adjustments).

**Comment by zetashift** [comment]
> Tree-sitter already supports strncat on the latest master: https://github.com/tree-sitter/tree-sitter/commit/0fc92c9a7d0ddb417bd74bf7f533bb8f3042dbe3

Does Zed( or the extensions part) use the master or is it pinned to a specific version? 

> Right now we're blocked on the upstream grammar resolve this issue
I have a PR up https://github.com/kylegoetz/tree-sitter-unison/pull/58 ! Let's see if it gets merged anytime soon :)

**Comment by maxdeviant** [comment]
> > Tree-sitter already supports strncat on the latest master: [tree-sitter/tree-sitter@0fc92c9](https://github.com/tree-sitter/tree-sitter/commit/0fc92c9a7d0ddb417bd74bf7f533bb8f3042dbe3)
> 
> Does Zed( or the extensions part) use the master or is it pinned to a specific version?

We patch our `tree-sitter` crate to a specific SHA:

https://github.com/zed-industries/zed/blob/cd5ddfe34bdda1626d9702cf2087b093acfa744c/Cargo.toml#L391-L392

And right now we're pinned to https://github.com/tree-sitter/tree-sitter/commit/7f21c3b98c0749ac192da67a0d65dfe3eabc4a63, so Zed already supports Wasm extensions using `strncat`.

**Comment by zetashift** [comment]
```
[2024-04-16T20:54:38Z INFO  extension::extension_builder] finished compiling extension /home/runner/work/extensions/extensions/extensions/unison
Error: Failed to instantiate wasm module: invalid import 'fwrite'
```

Is this WASM error from the parser or something external? I grepped tree-sitter-unison and it doesn't seem to use `fwrite`?

**Comment by maxbrunsfeld** [comment]
Maybe it calls `fprintf` or similar? If so, maybe that can be wrapped in an ifdef, disabling it for wasm builds? or remove it completely?

**Comment by zetashift** [comment]
> Maybe it calls `fprintf` or similar? If so, maybe that can be wrapped in an ifdef, disabling it for wasm builds? or remove it completely?

Seems it does: https://github.com/zed-industries/extensions/pull/361#issuecomment-2062441610
They seem to be enabled only if it's DEBUG and `log_level` is enabled.

**Comment by kylegoetz** [comment]
@maxbrunsfeld could you or zetashift check this PR? https://github.com/kylegoetz/tree-sitter-unison/pull/68

I'm conditionally excluding the fprintf stuff if `__EMSCRIPTEN__` is defined in the preprocessor. Based on Emscripten docs, all WASM builds using Emscripten define it.

**Comment by maxdeviant** [comment]
> @maxbrunsfeld could you or zetashift check this PR? [kylegoetz/tree-sitter-unison#68](https://github.com/kylegoetz/tree-sitter-unison/pull/68)
> 
> I'm conditionally excluding the fprintf stuff if `__EMSCRIPTEN__` is defined in the preprocessor. Based on Emscripten docs, all WASM builds using Emscripten define it.

We've used the `__wasm32__` directive in other instances: https://github.com/6cdh/tree-sitter-racket/pull/8

Emscripten defines it as well, I believe, but it's not Emscripten-specific.

**Comment by zetashift** [comment]
@kylegoetz this indeed fixes the `fwrite` error, but now the following error pops up for me locally:

```
extensions on  add-unison [!?]
❯ ./zed-extension  --scratch-dir ./scratch --source-dir extensions/unison --output-dir output
Error: Failed to instantiate wasm module: invalid import 'atof'
```

**Comment by kylegoetz** [comment]
@maxbrunsfeld  or @maxdeviant , any ideas for why this is happening, and if teh list of what is available is overly strict? If `atof` is not available to parse a string to a double, is `strtod`? Is there a definitive list of what Zed makes available to TSGs? I don't want to make poor @zetashift play telephone whack-a-mole with me as I fix one of these and they re-build and get another invalid import, and we repeat this a bunch.

**Comment by kylegoetz** [comment]
@maxdeviant and those are the *only* things I can import in my TSG that aren't also included in my repository? For example, here are my imports:

```c
#include "tree_sitter/parser.h"
#include <assert.h>
#include <stdio.h>
#include <inttypes.h> // needed for portability of PRId64
#include "jtckdint.h" // needed to prevent integer overflow in get_whole
#ifdef DEBUG
#include <assert.h>
#endif
#include <string.h>
#include <wctype.h>
#include <stdlib.h>
#include <math.h>
#include <ctype.h>
#include "maybe.c"
```

So no math, no wctype, no string, no assert, and no ctype code is usable? Let alone most things from stdlib?

kylegoetz:
@maxdeviant 

Am I able to use `assert.h`? How about `inttypes.h`? `string.h`? `wctype.h`? `math.h`? `ctype.h`? I'm trying to figure out how substantially I'd need to refactor `scanner.c` so it can be compiled into Zed.

maxdeviant:
> @maxdeviant
> 
> Am I able to use `assert.h`? How about `inttypes.h`? `string.h`? `wctype.h`? `math.h`? `ctype.h`? I'm trying to figure out how substantially I'd need to refactor `scanner.c` so it can be compiled into Zed. I think I use inttypes.h because it's the only way to reliably ensure cross-platform long size. Otherwise Windows and Linux do not have the same length for long, and then logging the variables isn't possible because the format code for an int64_t for example is different depending on platform. (In fact, I added it recently because the tree-sitter Github action for building and deploying to NPM was failing on Windows, and the fix was to use this lib.

I'll defer to @maxbrunsfeld here, as he's more familiar with Tree-sitter.

maxbrunsfeld:
@kylegoetz All types should be fine to use, but there are only certain libc functions available when running Tree-sitter parsers via wasm. The core library itself needs to provide the libc functions. If you run `tree-sitter build --wasm` in your grammar repo, it will tell you if you are using any functions that are unavailable in wasm, and output a list of the functions that *are* available.

kylegoetz:
Okay, the only issue is `atof`, which I can't just copy-paste into scanner.c because it's platform-dependent code as per the C standard. So I opened a ticket and filed a PR for it in the `tree-sitter` repo. I don't know how to update `wasm-stdlib.h`, though.

maxbrunsfeld:
@kylegoetz do you need to store the actual `float` value represented by a string, or do you just need to check if it *is* a float?

If the latter, I think it would be better to parse the float manually, since any libc function like `atof` or `strtof` is going have locale-specific details that may not match the actual syntax of the Unison language. `atof` is particularly problematic, because it just returns zero on failure AFAICT.

> it's platform-dependent code as per the C standard

To be clear, that's a reason **not** to use `atof` right? Your unison parser probably shouldn't behave differently on different platforms.

kylegoetz:
Thanks for the advice, everyone! One issue was Unison's "REPL" would accept out-of-bounds `Float`s and pretty print them as `Infinity`. I've confirmed with Unison devs that `Infinity` is still of type `Float`, so there's no issue about needing to determine float or infinity tokens by bounds checking.

So I've removed bounds checking from the scanner, which has dramatically simplified numeric parsing!

@zetashift mind checking to see if you can build now? `tree-sitter build-wasm` works without error for me.

maxdeviant:
@zetashift @kylegoetz Thank you both for all your work getting this one published!

kylegoetz:
> @zetashift @kylegoetz Thank you both for all your work getting this one published!

Thanks for the patience to your and @maxbrunsfeld for patiently explaining some stuff to me.

zetashift:
Thanks for getting this through the line! It's @kylegoetz who did the hard part! :1st_place_medal:

PR #363, opened 2024-03-22
Added Nord light theme

mikasius:
nord-light

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -180,7 +180,7 @@ version = "0.0.4"
 
 [nord]
 path = "extensions/nord"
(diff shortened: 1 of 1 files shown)


mikasius:
@cla-bot check

mikasius:
Hm, can't understand what I'm doing wrong...

maxdeviant:
> Hm, can't understand what I'm doing wrong... Could it be because the first commit was made from a different email?

Yea, the first commit doesn't have the right author information.

I think you'll need to amend the commit and then force push the branch.

PR #366, opened 2024-03-22
Add Vesper Pro theme

webhooked:


Moshyfawn:
Would it make sense to combine the Pro and Base versions into one theme family since the later already exists as a Zed theme extension?

webhooked:
@Moshyfawn The two themes aren't really associated. This is basically a "new" theme heavily inspired by the original Vesper theme. Would you prefer if I renamed it and opened a new PR?

PR #373, opened 2024-03-24
Add markdown-oxide language server

Feel-ix-343:


maxbrunsfeld:
It looks like the submodule itself needs to be added to the git repo.

maxdeviant:
> It looks like the submodule itself needs to be added to the git repo.

I added the submodule in the latest commit.

Feel-ix-343:
The other thing is that the extension does not automatically download the language server. The user can simply install through cargo. 

The extension is too pre-beta imo to have binaries, so the alternative would be a cargo install through the extension. This however would be a more complex installation than the regular lsp extension. Also, either the latest commit or latest tag should be installed - there would be frequent updates. What do you make of this @maxbrunsfeld ? How would the extension manage the installation or should it? (Its quite common in neovim for a LS to be installed from a normal package manager)

maxdeviant:
> The other thing is that the extension does not automatically download the language server. The user can simply install through cargo.
> 
> The extension is too pre-beta imo to have binaries, so the alternative would be a cargo install through the extension. This however would be a more complex installation than the regular lsp extension. Also, either the latest commit or latest tag should be installed - there would be frequent updates. What do you make of this @maxdeviant ? How would the extension manage the installation or should it? (Its quite common in neovim for a LS to be installed from a normal package manager)

For now I think we can expect the user to have it installed already, and report an error if we can't find it.

Here's how we do something similar in the Uiua extension: https://github.com/zed-industries/zed/blob/5361a4d72dfe8f286581896e9aa7feb232672493/extensions/uiua/src/uiua.rs#L15-L17

Feel-ix-343:
I'll add that

Feel-ix-343:
Is there a way to show the error message in the UI?

Feel-ix-343:
I updated the extension. Do you need to update submodules on here now?

Feel-ix-343:
> Not currently, but we'll probably add something like that in the future.

I can help! I assume the LanguageServerInstallStatus::Failed(...) would be used for the UI right?

Feel-ix-343:
Beyond that, I think this is ready to merge

maxdeviant:
@Feel-ix-343 Thanks!

Feel-ix-343:
@maxdeviant how are extensions updated? I just pushed a commit to the repo

PR #377, opened 2024-03-25
Update markdown oxide extension

Feel-ix-343:


Feel-ix-343:
I just changed the extension description

PR #383, opened 2024-03-26
Update Paraíso to extension.toml format

treffynnon:
Updates the extension to use the [freshly documented](https://github.com/zed-industries/extensions/commit/6ab7d4253b6de77025a586e6afc808f668098eac) extension.toml format instead of extension.json.


maxdeviant:
@treffynnon Just a heads-up that currently on Zed versions v1.129.x and higher will be able to see extensions using the `extension.toml` format. That version will be released to Preview tomorrow and then Stable the following week.

PR #393, opened 2024-03-27
Add Vitesse theme

d1y:


d1y:
I'll be right back to fix this in a moment #358

maxdeviant:
@d1y Thank you! This is the theme I use myself, so I've been waiting for someone to add it 😄

PR #405, opened 2024-03-29
Upgrade Make to 1.0.0

caius:
Now contains Makefile syntax highlighting!

Also upgraded to the new `extensions.toml` configuration and tidied up some bits in the repository.

Loaded as a Dev Extension in Zed Preview and it functions as expected locally.


caius:
Thanks!

PR #406, opened 2024-03-29
Update log extension

evrsen:


maxdeviant:
@evrsen You'll also need to bump the extension version in your repo and in the `extensions.toml` in this repo.

maxdeviant (changes requested):
See my comment above.

maxdeviant (approved):

Issue #28, opened 2023-05-04
Oceanic Next


jchaing:
### Check for existing issues

- [X] Completed

### Describe the feature

Kindly requesting to add Oceanic Next Theme support 🤞

https://github.com/voronianski/oceanic-next-color-scheme

### If applicable, add mockups / screenshots to help present your vision of the feature

![image](https://user-images.githubusercontent.com/9937696/236305481-cd5e0f67-3bec-4901-bfd9-d63920b2c85f.png)

JosephTLyons:
Hello @jchaing, I've transferred your issue into the `extensions` repository and have reopened it.

JosephTLyons:
A [Oceanic Next theme extension](https://github.com/rkunev/oceanic-next) now exists.

<img width="720" alt="SCR-20240716-mktk" src="https://github.com/user-attachments/assets/12cadfb7-85b0-4596-af23-3963de13d7cb">

PR #409, opened 2024-03-29
Add Oceanic Next

rkunev:
Port of [Oceanic Next Color Scheme](https://github.com/voronianski/oceanic-next-color-scheme) with slight changes. Personally my favorite dark theme and it seems that others might agree too #28!

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -210,6 +210,10 @@
 	path = extensions/nu
 	url = https://github.com/zed-extensions/nu.git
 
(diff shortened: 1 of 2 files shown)


rkunev:
@cla-bot check

maxdeviant:
@rkunev Thanks, this looks great!

Just to note, because you're using the `extension.toml` format you'll need to be on the Preview version of Zed (v0.129.x) in order to see the extension. That version will be going out to Stable next week.

PR #412, opened 2024-03-29
Add Basher

d1y:
Pick up https://github.com/zed-industries/zed/pull/9902

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -18,6 +18,10 @@
 	path = extensions/base16
 	url = https://github.com/bswinnerton/base16-zed.git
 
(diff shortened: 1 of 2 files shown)


maxdeviant (changes requested):
@d1y Would you mind changing the extension ID to something other than `bash`? I'd like to leave it open for when we decide what our plan is when we eventually move Bash support out into an extension.

d1y:
Now that I have some time, I'm going to change it to basher, just a minute

maxdeviant (approved):
@d1y Thank you!

PR #418, opened 2024-03-30
Update sql extension

evrsen:
* Support prettier formatting (https://github.com/zed-industries/zed/issues/9537)

I'm unable to test the extensions locally and I'm not sure why, hoping this change is enough for enabling prettier sql formatting.


evrsen:
i dont understand, what's  wrong?

maxdeviant:
> i dont understand, what's wrong?

I think the problem is you have the `grammars/sql` committed in your extension repo as a submodule.

We just need the `sql.toml` file in that directory, as we fetch the source files for the grammar ourselves when packaging.

PR #419, opened 2024-03-30
Add asciidoc 0.0.1

andreicek:
This adds support for AsciiDoc syntax highlighting.

The tree sitter is still WIP, but it works to some extent and it's good to see at least a bit of color in the editor.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -457,3 +457,7 @@
 [submodule "extensions/zedwaita"]
 	path = extensions/zedwaita
 	url = https://github.com/someone13574/zed-adwaita-theme.git
+
+[submodule "extensions/zed-asciidoc"]
+	path = extensions/zed-asciidoc
+	url = https://github.com/andreicek/zed-asciidoc.git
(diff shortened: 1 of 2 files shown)


andreicek:
@cla-bot check

maxdeviant (changes requested):
Thanks for creating this, @andreicek!

Could we drop the `-zed` suffix from the extension ID? https://github.com/andreicek/zed-asciidoc/blob/b940fc1dff376f6eae860c577b4b0435a5bf9fb7/extension.toml#L1

Since these are all Zed extensions we don't need to clarify.

andreicek:
@maxdeviant sorry for keeping this PR waiting. I updated the name, can you please check everything is in order now?

.gitmodules
```diff
@@ -457,3 +457,7 @@
 [submodule "extensions/zedwaita"]
 	path = extensions/zedwaita
 	url = https://github.com/someone13574/zed-adwaita-theme.git
+
+[submodule "extensions/zed-asciidoc"]
```
huacnlee, line 461: Maybe you also need to remove the `zed-` prefix for submodule path.

maxdeviant (approved):
Thanks for this, @andreicek!

PR #420, opened 2024-03-30
Use the binary release for Markdown Oxide

Feel-ix-343:
I mostly copied the gleam extension, and I believe this works well, however I would love some feedback on the extension's code. 

And how important is it to support aarch platforms? I only support x86_64 right now

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -161,7 +161,7 @@ version = "1.0.0"
 
 [markdown-oxide]
 submodule = "extensions/markdown-oxide"
-version = "0.0.2"
+version = "0.0.3"
 
 [material-dark]
 submodule = "extensions/material-dark"


Feel-ix-343:
Hey also @maxdeviant I noticed that (as of about a week ago) the language server plugins are not available yet in Zed (macos release). Is this expected?

maxdeviant:
> And how important is it to support aarch platforms? I only support x86_64 right now

I would say it's a nice-to-have, but it's only going to become more common with the rising number of Apple Silicon Macs.

If the upstream language server distributes aarch64 binaries then I'd say we should definitely use them.

Feel-ix-343:
Ill add it to the LS then update here.

Feel-ix-343:
So right now I have:
image

How does that look for support? Anything I'm missing?

maxdeviant:
> So right now I have: image
> 
> How does that look for support? Anything I'm missing?

Looks good!

PR #428, opened 2024-04-01
Bump catppuccin to 0.2.4

tecandrew:
- closes https://github.com/zed-industries/zed/issues/10024
- use `extension.toml` instead of `extension.json`

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -33,7 +33,7 @@ version = "0.0.1"
 
 [catppuccin]
 submodule = "extensions/catppuccin"
-version = "0.2.3"
+version = "0.2.4"
 
 [colorizer]
 submodule = "extensions/colorizer"


tecandrew:
I'm finding issues with the `extension.toml` file: defining it instead of the `extension.json` file prevents me from using the theme in Development?


using extension.toml

https://github.com/zed-industries/extensions/assets/21319730/45e200bd-eb14-47fb-9894-357ec5487f67


using extension.json

https://github.com/zed-industries/extensions/assets/21319730/0c535e6c-af76-4469-8e6c-4abcd8b626dd


on a similar note, MacOS Classic themes seems to be gone and it also defines the [`extension.toml`](https://github.com/huacnlee/zed-theme-macos-classic/blob/main/extension.toml) file.

Moshyfawn:
@tecandrew which version of Zed are you using? Only Zed v0.129.x and up can see extensions using `extension.toml`.

tecandrew:
@Moshyfawn ahh, okay. should have stated that. i'm using `Zed 0.128.3` (latest stable/non-preview)

PR #429, opened 2024-04-02
Update slint extension

flukejones:


maxdeviant:
@flukejones Just a heads-up that v0.0.5 of the `zed_extension_api` is not yet supported in a released version of Zed. Compatibility will be added in the Preview release that is going out tomorrow.

PR #432, opened 2024-04-02
Add Roc extension

h2000:


maxbrunsfeld:
```
Error: language Roc provides query /home/runner/work/extensions/extensions/extensions/roc/languages/roc/textobjects.scm but no grammar
```

Seems like a grammar is missing?

h2000:
@maxbrunsfeld  thanks for the finding. I'll take a look.

h2000:
@maxbrunsfeld what validation/test tools can I use to analyse that the extension (language server/tree sitter) works? Loading into to Zed via "Install Dev Extension" works, Roc is then shown as installed plugin and the roc file type is detected. But no syntax highlighting or language server activity is visible. I looked into the ~/Library/Logs/Zed/Zed.log, which shows no errors or other info.

maxbrunsfeld:
I think the roc language’s config toml needs to list roc as its grammar.

maxbrunsfeld:
We need to a better job surfacing those problems. The CI failed with an error indicating it, but it is not checked at all when running Install Dev Extension.

JosephTLyons:
@rtfeldman 👀

h2000:
@maxbrunsfeld thanks a lot for your help, now I have the plugin working locally with syntax highlighting and language server support. All changes are pushed and ready to be merged. 
image

maxbrunsfeld:
Thanks @h2000 - this should be in the extension store shortly.

PR #437, opened 2024-04-03
Bump D extension to 0.0.3

gdamore:
This adds support for LSP via serve-d.  LSP configuration is not yet present, but this already provides useful guidance from DScanner, as well as rudimentary Find References, Rename symbol and similar.

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -59,7 +59,7 @@ version = "0.0.1"
 
 [d]
 submodule = "extensions/d"
-version = "0.0.2"
+version = "0.0.3"
 
 [dockerfile]
 submodule = "extensions/dockerfile"


maxdeviant (changes requested):
@gdamore Could you update the version number in `extensions.toml` as well?

I can get to it in a bit, if you don't get to it first.

gdamore:
Done.

gdamore:
(I force pushed it so it only looks like when commit when it merges.)

maxdeviant:
> (I force pushed it so it only looks like when commit when it merges.)

We squash commits on merge, so it's all the same at the end of the day 😄

maxdeviant (approved):

PR #443, opened 2024-04-04
Add Helm extension

cabrinha:
Support for https://helm.sh

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -110,7 +110,7 @@
 	path = extensions/graphql
 	url = https://github.com/11bit/zed-extension-graphql.git
 
-[submodule "helm"]
+[submodule "extensions/helm"]
 	path = extensions/helm
 	url = https://github.com/hjoshi123/helm.zed.git
 
--- a/.gitmodules
+++ b/.gitmodules
@@ -110,10 +110,6 @@
 	path = extensions/graphql
 	url = https://github.com/11bit/zed-extension-graphql.git
 
-[submodule "extensions/helm"]
-	path = extensions/helm
-	url = https://github.com/cabrinha/helm.zed.git
-
 [submodule "extensions/horizon"]
 	path = extensions/horizon
 	url = https://github.com/ayn2op/zed-horizon.git
@@ -381,3 +377,6 @@
 [submodule "extensions/zedspace"]
 	path = extensions/zedspace
 	url = https://github.com/Brunowilliang/zedspace.git
+[submodule "helm"]
+	path = extensions/helm
+	url = https://github.com/cabrinha/helm.zed.git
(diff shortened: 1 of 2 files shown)


maxdeviant:
It looks like there are still some problems when trying to compile the `go-template` grammar to Wasm:

```
[2024-04-07T12:09:22Z INFO  extension::extension_builder] compiling go-template parser
Error: failed to compile extension

Caused by:
    0: failed to compile grammar 'go-template'
    1: failed to compile go-template parser with clang: wasm-ld: warning: creating shared libraries, with -shared, is not yet stable
       wasm-ld: error: symbol exported via --export not found: tree_sitter_go-template
       clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

maxbrunsfeld:
>  `wasm-ld: error: symbol exported via --export not found: tree_sitter_go-template`

The function is probably called `go_template` (with an underscore, not a dash), so that needs to the name of the grammar. @cabrinha I think that means we need to change the directory name from `grammars/go-template` to `grammars/go_template`.

Also, since this is a new extension, it'd be good to be using the latest extension schema and API. Could you copy the structure of the extensions in here? https://github.com/zed-industries/zed/tree/main/extensions

maxbrunsfeld:
That would only be needed if there is a Helm language server.

maxbrunsfeld:
The `Cargo.toml` is only needed if you're providing Rust code to configure a language server.

maxbrunsfeld:
It looks like the grammar actually is called  `gotmpl` (not `go_template`):  https://github.com/ngalaiko/tree-sitter-go-template/blob/master/grammar.js#L3

maxbrunsfeld:
Yeah, apparently `yaml_command` is not a node in the grammar. Where did that injection query come from?

mikayla-maki:
Discussed in discord, author has decided to discontinue this effort, that said extension is very close, just needs to improve the association between heml and yaml.

qvalentin:
> That would only be necessary if there was a Helm language server.

Well, there is https://github.com/mrjosh/helm-ls 😄 


Perhaps I can provide some information on how helm is currently handled in Neovim:

- There is a specific file type that is set for files that match the pattern `*/templates/*.{yaml,yml},*/templates/*.tpl,*.gotmpl,helmfile*.{yaml,yml}`.
- The helm dialogue in the https://github.com/ngalaiko/tree-sitter-go-template repo is only used to be able to include both gotmpl and helm in https://github.com/nvim-treesitter/nvim-treesitter (currently it would not matter which grammar zed would include).
- Injections should use combined injections if possible (as discussed in https://github.com/ngalaiko/tree-sitter-go-template/pull/11), I think you were on the right track here @cabrinha

Issue #388, opened 2024-03-26
Biome


luckydye:
### Check for existing issues

- [X] Completed

### Language

Biome

### Language server link

https://github.com/biomejs/biome/tree/main/packages/%40biomejs/biome

### Misc notes

I tried to implement the extension with the following observations:
- [ ] "language" in 'extension.toml' only accepts a single language, but additional variations like "TypeScript" as "TSX" are required.
- [ ] Formatting over LSP
- [x] ~~quick fixes through LSP do not work yet (?)~~
- [x] ~~lsp server does not see biome.json in the project root~~
- [x] Diagnostics are displayed correctly 👏

PR #444, opened 2024-04-04
Add Biome extension

luckydye:
Adds support for the Biome Language Server. (https://github.com/zed-industries/extensions/issues/388)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -14,6 +14,10 @@
 	path = extensions/beancount
 	url = https://github.com/zed-extensions/beancount.git
 
+[submodule "extensions/biome"]
+	path = extensions/biome
+	url = https://github.com/biomejs/biome-zed.git
+
 [submodule "extensions/blanche"]
 	path = extensions/blanche
 	url = https://github.com/kwonoj/zed-blanche.git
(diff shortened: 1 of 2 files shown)


maxdeviant (changes requested):
Thanks for opening this @luckydye!

Would you mind filling out the `authors` field for the extension? https://github.com/biomejs/biome-zed/blob/a3eda2789ca75c40e613d9c905671cd31d2b2ff5/extension.toml#L6

luckydye:
My bad, I changed the url..

maxdeviant (approved):
Thank you!

quinncomendant:
Thanks for this, @luckydye! I’m really excited to have Biome in Zed.

Quick question about how this was implemented: After installing the Biome Zed extension, is it necessary to set any configuration to disable ESLint and Prettier, since their functionality overlaps with Biome? I’m worried they might both be running at the same time, creating conflicts.

luckydye:
@quinncomendant In my oppinion Zed still lacks a little bit there.
As far as I know, the only solution is the formatter config field to configure the behaviour. 

https://zed.dev/docs/reference/all-settings#formatter

Linting can be used side-by-side.

PR #446, opened 2024-04-05
Add Typst extension

WeetHet:
Adds support for typst (typst.app) language with tinymist language server


maxdeviant:
@WeetHet Thanks!

PR #448, opened 2024-04-06
Add Glazier themes

airgap:
1. Glazier | Phantom
2. Glazier | Deep Space
3. Glazier | Redrum

Early prototype

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -440,3 +440,7 @@ version = "0.0.1"
 submodule = "extensions/zed"
 path = "extensions/zig"
 version = "0.0.1"
+
+[glazier]
+submodule = "extensions/glazier"
+version = "0.0.0"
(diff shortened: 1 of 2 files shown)


airgap:
@cla-bot check

extensions.toml
```diff
@@ -440,3 +440,7 @@ version = "0.0.1"
 submodule = "extensions/zed"
 path = "extensions/zig"
 version = "0.0.1"
+
+[glazier]
+submodule = "extensions/glazier"
+version = "0.0.0"
```
maxdeviant, line 446: Could we use a non-zero version here?

maxdeviant (changes requested):
Would you mind adjusting the version number so it doesn't have all zero components?

It looks like the Git submodule is missing as well (we have a `.gitmodules` entry, but no actual submodule committed).

maxdeviant (approved):
@airgap Thank you!

PR #454, opened 2024-04-08
Add support for Lox language

arian81:
Lox is a toy language introduced in the book [Crafting Interpreters](https://craftinginterpreters.com/). There was already a Treesitter schema written for it so I thought it would be cool to add it to zed. Hopefully future generations of compiler enthusiasts reading that book can use zed to learn how to make compilers from scratch.


maxdeviant:
@arian81 Thank you!

PR #456, opened 2024-04-09
Add CSV support

huacnlee:
image

Ref issue: https://github.com/zed-industries/zed/issues/9653

----

The negative number is no highlight match, I have made a change to tree-sitter-csv: https://github.com/arnau/tree-sitter-csv/pull/1

I think we can release version v0.0.1 first to use, and then if the tree-sitter-csv accept that PR, I can release a new version.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -42,13 +42,13 @@
 	path = extensions/cosmos
 	url = https://github.com/nauvalazhar/cosmos.git
 
-[submodule "extensions/cue"]
-	path = extensions/cue
-	url = https://github.com/jkasky/zed-cue.git
-
 [submodule "extensions/csv"]
 	path = extensions/csv
 	url = https://github.com/huacnlee/zed-csv.git
+
+[submodule "extensions/cue"]
+	path = extensions/cue
+	url = https://github.com/jkasky/zed-cue.git
 
 [submodule "extensions/d"]
 	path = extensions/d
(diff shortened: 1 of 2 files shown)


huacnlee:
Hi @maxdeviant 

I saw you said in https://github.com/zed-industries/extensions/pull/412#pullrequestreview-1971377446

If you mind that the best extension name `csv` is taken by a third party, I can change the name.

However, you may also consider using this for the time being. If Zed Team wants to use a unified way to manage extensions such as `csv`, you can choose to directly override this extension (I don't mind this, you can do that directly).

I am currently adding this support to meet my current need to use Zed to open CSV files in my work. It would be better if there is highlighting. This only adds syntax support.

maxdeviant:
> Hi @maxdeviant
> 
> I saw you said in [#412 (review)](https://github.com/zed-industries/extensions/pull/412#pullrequestreview-1971377446)
> 
> If you mind that the best extension name `csv` is taken by a third party, I can change the name.
> 
> However, you may also consider using this for the time being. If Zed Team wants to use a unified way to manage extensions such as `csv`, you can choose to directly override this extension (I don't mind this, you can do that directly).
> 
> I am currently adding this support to meet my current need to use Zed to open CSV files in my work. It would be better if there is highlighting. This only adds syntax support.

I think in this case it's fine. We don't currently have CSV support built-in to Zed, so I think it's fair for you to claim the `csv` extension ID 🙂

PR #478, opened 2024-04-11
Add Justfile support

### LINKED ISSUE #156: Justfile
### Check for existing issues

- [X] Completed

### Language

Justfile

### Tree Sitter parser link

https://github.com/IndianBoy42/tree-sitter-just

### Language server link

_No response_

### Misc notes

Similar enough to makefiles to maybe just use make syntax.

jackTabsCode:
Fixes #156

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -130,6 +130,10 @@
 	path = extensions/java
 	url = https://github.com/samuser107/zed-java-extension.git
 
+[submodule "extensions/just"]
+	path = extensions/just
+	url = https://github.com/jackTabsCode/zed-just.git
+
 [submodule "extensions/kiselevka"]
 	path = extensions/kiselevka
 	url = https://github.com/kdubrovsky/kiselevka.git
--- a/extensions.toml
+++ b/extensions.toml
@@ -175,6 +175,10 @@ version = "0.1.1"
 submodule = "extensions/java"
 version = "0.0.3"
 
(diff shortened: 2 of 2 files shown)


jackTabsCode:
Not entirely sure why the workflow failed, can anyone point me in the right direction?

maxdeviant:
> Not entirely sure why the workflow failed, can anyone point me in the right direction?

It looks like the Git submodule wasn't committed to the repo.

There is an entry in `.gitmodules`, but we're missing the actual submodule reference at `extensions/just`.

jackTabsCode:
> > Not entirely sure why the workflow failed, can anyone point me in the right direction?
> 
> It looks like the Git submodule wasn't committed to the repo.
> 
> There is an entry in `.gitmodules`, but we're missing the actual submodule reference at `extensions/just`.

I think it should be fixed now? I have very limited experience with submodules.

maxdeviant:
Looks like it's failing now because the version number in the extension's `extension.toml` doesn't match the one listed in `extensions.toml`: https://github.com/jackTabsCode/zed-just/blob/c35a129a1f0f8bdc0e9b7305aea06c7b7b5f3e09/extension.toml#L4

jackTabsCode:
Ok, let's try again?

maxdeviant:
@jackTabsCode Thanks!

PR #479, opened 2024-04-11
Add min theme

jacobdalamb:
Help needed!
Very early start.


maxdeviant:
@jacobdalamb Is this extension ready to be published?

If you're not done with it yet that's perfectly fine, but in that case I would like to close out the PR so it isn't sitting idle in the list.

We can also go ahead and publish it in its current state, if that is what you want.

Issue #163, opened 2023-03-29
Twig


codyhopper:
### Check for existing issues

- [X] Completed

### Language

Twig

### Tree Sitter parser link

https://github.com/gbprod/tree-sitter-twig

### Misc notes

Twig is a modern template engine for PHP and the primary language for Craft CMS development. 

Also bumped PHP support: https://github.com/zed-industries/zed/issues/5652

corneliusio:
Looks to be added with https://github.com/zed-industries/extensions/pull/575

JosephTLyons:
There is a [Twig extension](https://github.com/YussufSassi/zed-twig) now.

<img width="719" alt="SCR-20240716-maih" src="https://github.com/user-attachments/assets/c91416d3-cd7f-451d-a8a3-9fb3f4200f9f">

PR #486, opened 2024-04-12
Add Twig extension

gopeter:
Extension to enable syntax highlighting for Twig files in Zed (#163).

Unfortunately, the extension does not support formatting. The available Prettier plugins didn't worked well, so I've decided to skip this. If you want to format your Twig files, install [https://www.djlint.com/docs/getting-started/](djLint) and configure Zed/your project like this:

```json
{
  "language_overrides": {
    "Twig": {
      "format_on_save": {
        "external": {
          "command": "bash",
          "arguments": [
            "-c",
            "cat {buffer_path} | djlint - --profile=nunjucks --reformat"
          ]
        }
      }
    }
  }
}
```

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -476,5 +476,4 @@ version = "0.0.1"
 
 [zig]
 submodule = "extensions/zed"
-path = "extensions/zig"
 version = "0.0.1"
--- a/extensions.toml
+++ b/extensions.toml
@@ -416,6 +416,11 @@ submodule = "extensions/zed"
 path = "extensions/toml"
 version = "0.0.2"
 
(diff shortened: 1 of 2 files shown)


extensions.toml
```diff
@@ -416,6 +416,11 @@ submodule = "extensions/zed"
 path = "extensions/toml"
 version = "0.0.2"
 
+[twig]
+submodule = "extensions/twig"
+path = "extensions/twig"
```
maxdeviant, line 421: @gopeter This `path` is not needed here.

This is only for when a submodule contains multiple extensions.

Issue #475, opened 2024-04-11
CSV (column-based highlighting)


JosephTLyons:
### Check for existing issues

- [X] Completed

### Language

CSV

### Tree Sitter parser link

?

### Language server link

?

### Misc notes

We have a CSV extension now, but it highlights items based on type:

<img width="829" alt="SCR-20240411-mvco" src="https://github.com/zed-industries/extensions/assets/19867440/82b74301-076b-4ea1-be13-bc88dbaf73a2">

It would be cool to havea CSV extension that highlights based on column, like [CSV Rainbow](https://marketplace.visualstudio.com/items?itemName=mechatroner.rainbow-csv) from VS Code:

<img width="805" alt="SCR-20240411-mwkq" src="https://github.com/zed-industries/extensions/assets/19867440/d5d721c9-18f9-481d-ab41-f6eae15dae76">

weartist:
I can do this, and if I need to be an extension, I'll need to look at the instructions

weartist:
@JosephTLyons  I'd love to provide a CSV extension to support different colors and table editing, I checked the extension's documentation, and wanted to make sure that at the moment we can only get some of the functions of tree-sitter through extensions, and where can I learn about the new API opening related to extensions

JosephTLyons:
> @JosephTLyons I'd love to provide a CSV extension to support different colors and table editing, I checked the extension's documentation, and wanted to make sure that at the moment we can only get some of the functions of tree-sitter through extensions, and where can I learn about the new API opening related to extensions

Hey @weartist - sorry for the confusion here - extensions currently only support themes and languages - I created these issues so I could remove an issue in the zed repo about this. This column-based CSV extension could possibly be built right now, but the table editor certainly can't. I've added an `awaiting infrastructure` to signal that these types of extensions can't be built yet.

weartist:
> > @JosephTLyons I'd love to provide a CSV extension to support different colors and table editing, I checked the extension's documentation, and wanted to make sure that at the moment we can only get some of the functions of tree-sitter through extensions, and where can I learn about the new API opening related to extensions
> 
> Hey @weartist - sorry for the confusion here - extensions currently only support themes and languages - I created these issues so I could remove an issue in the zed repo about this. This column-based CSV extension could possibly be built right now, but the table editor certainly can't. I've added an `needs infrastructure` to signal that these types of extensions can't be built yet.

[...]

PR #537, opened 2024-04-13
Add rainbow-csv extension

weartist:
Add a csv extension to solve: #475 

it look like:
image

At the same time, the tsv extension has been added, They may be merged into the same extension in the future,and I will refine the logic of automatic indentation alignment in the future

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -286,9 +286,6 @@
 	path = extensions/zed-rainbow-csv
 	url = https://github.com/weartist/zed-rainbow-csv
 
-[submodule "extensions/zed-rainbow-tsv"]
-	path = extensions/zed-rainbow-tsv
-	url = https://github.com/weartist/zed-rainbow-tsv
 
 [submodule "extensions/scala"]
 	path = extensions/scala
--- a/.gitmodules
+++ b/.gitmodules
@@ -282,9 +282,9 @@
 	path = extensions/r
 	url = https://github.com/ocsmit/zed-r.git
 
-[submodule "extensions/zed_rainbow_csv"]
-	path = extensions/zed_rainbow_csv
-	url = https://github.com/weartist/zed_rainbow_csv
+[submodule "extensions/zed-rainbow-csv"]
+	path = extensions/zed-rainbow-csv
+	url = https://github.com/weartist/zed-rainbow-csv
 
 [submodule "extensions/scala"]
 	path = extensions/scala
--- a/.gitmodules
+++ b/.gitmodules
@@ -286,6 +286,10 @@
 	path = extensions/zed-rainbow-csv
 	url = https://github.com/weartist/zed-rainbow-csv
 
+[submodule "extensions/zed-rainbow-tsv"]
+	path = extensions/zed-rainbow-tsv
+	url = https://github.com/weartist/zed-rainbow-tsv
+
 [submodule "extensions/scala"]
 	path = extensions/scala
 	url = https://github.com/scalameta/metals-zed.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -286,7 +286,6 @@
 	path = extensions/zed-rainbow-csv
 	url = https://github.com/weartist/zed-rainbow-csv
 
-
 [submodule "extensions/scala"]
 	path = extensions/scala
 	url = https://github.com/scalameta/metals-zed.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -346,10 +346,6 @@
 	path = extensions/r
 	url = https://github.com/ocsmit/zed-r.git
 
-[submodule "extensions/zed-rainbow-csv"]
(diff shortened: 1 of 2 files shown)


huacnlee:
Hi @weartist you can add this support in csv extension.

https://github.com/huacnlee/zed-csv

I was created that extension last week, but your version is better for display.

We can to merge your version in to that, and you are the member or zed-csv to maintain it.

weartist:
> Hi @weartist you can add this support in csv extension.
> 
> https://github.com/huacnlee/zed-csv
> 
> I was created that extension last week, but your version is better for display.
> 
> We can to merge your version in to that, and you are the member or zed-csv to maintain it.

I looked at your extension, but because we need to separate the different colors by columns, we need a special tree-sitter match, maybe we can coexist

huacnlee:
I mean, if you want, you can:

1. Just use `csv` extension name (If you choose this, just rename your extension name and remove old info.). 
2. Or merge your code into the same project if you like.

I can do both.

Then there is only have 1 CSV extension for Zed.
You're doing is better than I did.

weartist:
> I mean, if you want, you can:
> 
> 1. Just use `csv` extension name (If you choose this, just rename your extension name and remove old info.).
> 2. Or merge your code into the same project if you like.
> 
> I can do both.
> 
> Then there is only have 1 CSV extension for Zed. You're doing is better than I did. Because I will be going to use your version instead of old one when you release. So maybe the old one is useless now.



> I mean, if you want, you can:
> 
> 1. Just use `csv` extension name (If you choose this, just rename your extension name and remove old info.).
> 2. Or merge your code into the same project if you like.
> 
> I can do both.
> 
> Then there is only have 1 CSV extension for Zed. You're doing is better than I did. Because I will be going to use your version instead of old one when you release. So maybe the old one is useless now.

Yes, maybe that's a better way to keep one, and I'm also considering if I could combine csv and tsv into one extension

huacnlee:
Here is an example of 1 Zed extension including 2 languages support.

https://github.com/navi-language/zed-navi

I have done a search, looks like TSV is a part of the CSV?

If that right, I think we can include them and use `CSV` as the extension name.

weartist:
> Here is an example of 1 Zed extension including 2 languages support.
> 
> https://github.com/navi-language/zed-navi
> 
> I have done a search, looks like TSV is a part of the CSV?
> 
> If that right, I think we can include them and use `CSV` as the extension name.

 yes, I also looked at some of the code and found out putting both of them in one extension is currently possible, I made some modifications and it is currently compatible with both formats

weartist:
@maxdeviant hi, I wonder if the extension qualifies, and I'll continue to think about the logic of the indentation if I can

maxdeviant:
> @maxdeviant hi, I wonder if the extension qualifies, and I'll continue to think about the logic of the indentation if I can

Yea, I'd be happy to merge this one.

Could you remove the Rust crate in the extension, since it doesn't appear to be used?

https://github.com/weartist/zed-rainbow-csv/blob/d4075700728d68d2292c7ee7c1b5d88330f57e09/src/lib.rs

So no `Cargo.toml` or `src/` directory.

JosephTLyons:
This is looking 🔥!

weartist:
> > @maxdeviant hi, I wonder if the extension qualifies, and I'll continue to think about the logic of the indentation if I can
> 
> Yea, I'd be happy to merge this one.
> 
> Could you remove the Rust crate in the extension, since it doesn't appear to be used?
> 
> https://github.com/weartist/zed-rainbow-csv/blob/d4075700728d68d2292c7ee7c1b5d88330f57e09/src/lib.rs
> 
> So no `Cargo.toml` or `src/` directory.

I misunderstood this, I thought we needed to keep it there, thanks for the reminder

JosephTLyons:
The crate is optional, only if you are adding language server support would you need it. Also, really excited to use this extension in a blog post Im doing! 😂

maxdeviant:
@weartist I think we're almost there.

We need to remove the `Cargo.toml` from the extension repository, as otherwise the extension builder thinks there's a Rust crate it needs to build: https://github.com/weartist/zed-rainbow-csv/blob/0cc3fa8360a8549f1744c67349fe5b6aaab3dd43/Cargo.toml

weartist:
> @weartist I think we're almost there.
> 
> We need to remove the `Cargo.toml` from the extension repository, as otherwise the extension builder thinks there's a Rust crate it needs to build: https://github.com/weartist/zed-rainbow-csv/blob/0cc3fa8360a8549f1744c67349fe5b6aaab3dd43/Cargo.toml


Handled, sorry for my previous misunderstanding

maxdeviant:
@weartist Thanks, this looks cool!

JosephTLyons:
Very cool! The example in my Markdown file! Going to tweet this extension out tomorrow, if there isn't anything big lined up.

SCR-20240425-tznu

PR #579, opened 2024-04-18
init cargo workspace

eth0net:
While experimenting with making extensions I found exploring the code for other extensions in the repo wasn't supported as they weren't handled gracefully by rust-analyzer.

Implementing a Cargo workspace helped solve this issue however, so I thought I'd open this up to explore others' views on it.


maxdeviant:
It's not really intended for extension work to happen within the submodules in this repo, at least not from a holistic perspective. 

Generally you would want to pull down the upstream repository for working on an extension, or open up a new instance of Zed at the root of the submodule.

Having a Cargo workspace doesn't really make sense here.

PR #594, opened 2024-04-20
Add LSP support for GDScript

grndctrl:
Thank you @moomerman for the LSP implementation.

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -139,7 +139,7 @@ version = "1.0.0"
 
 [gdscript]
 submodule = "extensions/gdscript"
-version = "0.0.2"
+version = "0.1.0"
 
 [gemini]
 submodule = "extensions/gemini"


grndctrl:
Strange, it appears to be failing on code that hasn't changed. Or am I reading this wrong?

maxdeviant:
> Strange, it appears to be failing on code that hasn't changed. Or am I reading this wrong?
> 
> EDIT: @maxdeviant Since it fails on the version bump, is this something I should omit?

It looks this is an existing problem with one of the included queries that wasn’t caught before:

```
[2024-04-20T14:36:09Z INFO  zed_extension] loaded grammar gdscript
Error: Query error at 5:7. Impossible pattern:
(call (name) @name) @reference.call
      ^
```

Back when the GDScript extension was initially added we just bundled the grammars and so long as they compiled to Wasm we called it a day.

Now the publishing process is more sophisticated and attempts to load the grammar and queries after the grammar is built, and that has uncovered a problem with the queries.

grndctrl:
Fixed the tags, all is good now

maxdeviant:
> Fixed the tags, all is good now

Awesome!

PR #628, opened 2024-04-25
Update CFEngine to 1.0.1

olehermanse:


maxdeviant:
@olehermanse Could you update the version in `extensions.toml`?

That's the file that CI looks at to know whether it needs to publish the extension.

PR #643, opened 2024-04-29
Add Gruber Darker theme

th0jensen:
A port of the emacs theme of the same name by rexim
Screenshot 2024-04-29 at 1 30 21 PM


maxdeviant:
@th0jensen It looks like the extension repo has both `extension.json` and `extension.toml`.

Could you remove the `extension.json`?

th0jensen:
Oopsies, sorry for that. Wasn't too sure on the convention so I left both in. Should be good to go now though

maxdeviant:
@th0jensen Thank you!

PR #644, opened 2024-04-29
Add Elixir extension

maxdeviant:
This PR adds the Elixir extension.

Elixir support was extracted from Zed in https://github.com/zed-industries/zed/pull/10948.

Since the Elixir extension provides three different language servers,
you'll need to use the `language_servers` setting to select the one you
want to use:

#### Elixir LS

```json
{
  "languages": {
    "Elixir": {
      "language_servers": [ "elixir-ls", "!next-ls", "!lexical", "..."]
    }
  }
}
```

#### Next LS

```json
{
  "languages": {
    "Elixir": {
      "language_servers": [ "next-ls", "!elixir-ls", "!lexical", "..."]
    }
  }
}
```

#### Lexical

```json
{
  "languages": {
    "Elixir": {
      "language_servers": [ "lexical", "!elixir-ls", "!next-ls", "..."]
    }
  }
}
```

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -119,6 +119,11 @@ version = "0.9.1"
 submodule = "extensions/elisp"
 version = "0.0.4"
 
+[elixir]
+submodule = "extensions/zed"
+path = "extensions/elixir"
+version = "0.0.1"
+
 [elm]
 submodule = "extensions/zed"
 path = "extensions/elm"


maxdeviant:
> Awesome, thanks!
> 
> I've started running it trying to set "elixir-ls" as that's the one I've been using.
> 
> ```json
>   "languages": {
>     "Elixir": {
>       "language_servers": ["elixir-ls", "!next-ls", "!lexical", "..."]
>     }
>   },
> ```
> 
> First thing I've noticed is that it downloaded Next LS...
> 
> Screenshot 2024-04-29 at 1 48 52 PM
> 
> Also, every time I save ~ANY file (elixir, json, makefile, etc)~ (edit: thought I saw this, but only seeing for elixir) I see "Checking... Next LS" in the status bar.
> 
> Perhaps I have something strange in my settings, but NextLS really seems to want to run.

I noticed the same with Next LS.

I wasn't sure if that was expected behavior or not, since I hadn't used it before. I did think the status messages were a little odd, with those weird combinations of characters in them.

baldwindavid:
Yup, kinda strange. At any rate, I'm only concerned about seemingly _only_ being able to use Next LS. I've tried a bunch of combinations and none of the "language_servers" settings seem to have any effect so I haven't been able to use ElixirLS or Lexical.

maxdeviant:
> Yup, kinda strange. At any rate, I'm only concerned about seemingly _only_ being able to use Next LS. I've tried a bunch of combinations and none of the "language_servers" settings seem to have any effect so I haven't been able to use ElixirLS (what I've always used and had configured above) or Lexical.

Oh, I see. I had missed that the first time around.

So Next LS shouldn't be running _at all_ because it's disabled by the `language_servers` setting.

What version of Zed are you using?

baldwindavid:
```
Zed: v0.134.0 (Zed Dev 91b3c24ed35d58438ae33970f07d1ff01d3acfc7)
OS: macOS 14.4.1
Memory: 32 GiB
Architecture: aarch64
```

...built locally with...

```
script/bundle-mac -l -o -i
```

maxdeviant:
> ```
> Zed: v0.134.0 (Zed Dev 91b3c24ed35d58438ae33970f07d1ff01d3acfc7)
> OS: macOS 14.4.1
> Memory: 32 GiB
> Architecture: aarch64
> ```
> 
> ...built locally with...
> 
> ```
> script/bundle-mac -l -o -i
> ```
> 
> I do get this warning...
> 
> ```
> warning: zed@0.134.0: Info: using '91b3c24ed35d58438ae33970f07d1ff01d3acfc7' hash for ZED_COMMIT_SHA env var
> ```
> 
> But that commit hash seems to be up to date... [zed-industries/zed@91b3c24](https://github.com/zed-industries/zed/commit/91b3c24ed35d58438ae33970f07d1ff01d3acfc7)

Okay, so that should all be good.

And just to confirm, this occurs even after a restart (with the setting already in place)?

baldwindavid:
> this occurs even after a restart (with the setting already in place)

That's correct; it has been restarted with the setting already in place.

I originally added the setting to `settings.json` just prior to running `bundle-mac`. When `Zed Dev` opened up it asked to install the elixir extension. I clicked "yes" and it immediately started downloading Next LS, which I could see in the status bar. After that point, I tried a bunch of settings changes to no avail.

I also pulled down an even more up-to-date version (that latest hash) and ran `bundle-mac` so it's been through a couple restart cycles.

maxdeviant:
> > this occurs even after a restart (with the setting already in place)
> 
> That's correct; it has been restarted with the setting already in place.
> 
> I originally added the setting to `settings.json` just prior to running `bundle-mac`. When `Zed Dev` opened up it asked to install the elixir extension. I clicked "yes" and it immediately started downloading Next LS, which I could see in the status bar. After that point, I tried a bunch of settings changes to no avail.
> 
> I also pulled down an even more up-to-date version (that latest hash) and ran `bundle-mac` so it's been through a couple restart cycles.

I'll have to take a look to see if I can reproduce.

We could also jump on a Zed call together sometime, if that would be of interest to you.

baldwindavid:
Sure, whatever is easiest for you; I don't want to take a bunch of your time.

baldwindavid:
The only other difference I've noticed is that as mentioned earlier in `Zed Dev` I get "Checking Next LS" upon saving any file type. I also get "Checking..." upon saving _any_ file type. This is not a message I see in Zed Preview. Perhaps that is just a new unrelated thing though.

maxdeviant:
@baldwindavid Another thought that occurred to me: do you have any other files (e.g., HEEx files) that could be starting up Next LS?

These are the languages that we have the language servers configured to use out-of-the-box:

https://github.com/zed-industries/zed/blob/d743c19fe21b7d54084b8bedcd5b6313051feea0/extensions/elixir/extension.toml#L9-L19

baldwindavid:
@maxdeviant Okay, think I've tracked down the related issue at least! I had tried to entirely clear my settings before, but not prior to the install. I did that and saw that this issue doesn't occur and then went through and found that the conflicting setting is...

```json
  "language_overrides": {
    "CSS": {
      "format_on_save": {
        "external": {
          "command": "prettier",
          "arguments": ["--stdin-filepath", "{buffer_path}"]
        }
      }
    },
    "JavaScript": {
      "format_on_save": {
        "external": {
          "command": "prettier",
          "arguments": ["--stdin-filepath", "{buffer_path}"]
        }
      }
    }
  }
```

If those settings are not included, all is well, but as soon as they are added both Lexical and NextLS immediately start to download (I can see them within the project at `.lexical` and `.elixir-tools`).

Those `language_override` settings were derived from the JS instructions at https://github.com/zed-industries/zed/blob/main/docs/src/languages/javascript.md#code-formatting

maxdeviant:
> @maxdeviant Okay, think I've tracked down the related issue at least! I had tried to entirely clear my settings before, but not prior to the install. I did that and saw that this issue doesn't occur and then went through and found that the conflicting setting is...
> 
> ```json
>   "language_overrides": {
>     "CSS": {
>       "format_on_save": {
>         "external": {
>           "command": "prettier",
>           "arguments": ["--stdin-filepath", "{buffer_path}"]
>         }
>       }
>     },
>     "JavaScript": {
>       "format_on_save": {
>         "external": {
>           "command": "prettier",
>           "arguments": ["--stdin-filepath", "{buffer_path}"]
>         }
>       }
>     }
>   }
> ```
> 
> If those settings are not included, all is well, but as soon as they are added both Lexical and NextLS immediately start to download (I can see them within the project at `.lexical` and `.elixir-tools`) and NextLS takes over.
> 
> Those `language_override` settings were derived from the JS instructions at https://github.com/zed-industries/zed/blob/main/docs/src/languages/javascript.md#code-formatting

Ah, `language_overrides` is an alias for `languages`, so they're both two ways of expressing the same thing.

So in your case you should be able to combine those settings from `language_overrides` under `languages`.

baldwindavid:
> Ah, language_overrides is an alias for languages, so they're both two ways of expressing the same thing.

🥳 

> Another thought that occurred to me: do you have any other files (e.g., HEEx files) that could be starting up Next LS?

Yes, indeed. That ended up being _another_ issue when opening a heex file.

Think this is all sorted now with the following settings...

```json
"languages": {
    "Elixir": {
      "language_servers": ["elixir-ls", "!lexical", "!next-ls", "..."]
    },
    "HEEX": {
      "language_servers": ["elixir-ls", "!lexical", "!next-ls", "..."]
    },
    "CSS": {
      "format_on_save": {
        "external": {
          "command": "prettier",
          "arguments": ["--stdin-filepath", "{buffer_path}"]
        }
      }
    },
    "JavaScript": {
      "format_on_save": {
        "external": {
          "command": "prettier",
          "arguments": ["--stdin-filepath", "{buffer_path}"]
        }
      }
    }
  }
```

Thanks!

baldwindavid:
It does make me wonder if `languages` and `language_overrides` (and similar settings) should/could either merge or warn about conflicts. 🤷 , maybe just an issue as the dust settles

maxdeviant:
> Also, don't think it would be the worst thing to default to only `elixir-ls` and let devs decide if they want to change it. It might not _always_ be the most popular, but believe it still currently is. Installing and enabling all of them and, I guess, defaulting to `next-ls` I suspect would be surprising to most. Or maybe I'm misunderstanding what went down here.

Yea, the plan is to make it so that only one of the language servers is used by default.

There are some design challenges with how to provide that setting from the extension, which is why it hasn't happened yet.

But I'll see what we can do in the short-term (probably before this goes out to Preview tomorrow).

baldwindavid:
👍 Cool and, worst case, some guidance in the release docs should help.

maxdeviant:
@baldwindavid Thanks for your help testing this, and glad we were able to get your issues figured out!

I opened https://github.com/zed-industries/zed/pull/11234 to make it so, by default, Zed will only use Elixir LS, until a setting indicates otherwise.

baldwindavid:
That seems like a nice way to handle it for the time being. Thanks for your help!

moomerman:
@maxdeviant my existing LSP config doesn't get picked up any more:

```json
"lsp": {
    "elixir-ls": {
      "settings": {
        "dialyzerEnabled": false
      }
    }
  },
```

Should this config go somewhere else now?

maxdeviant:
> @maxdeviant my existing LSP config doesn't get picked up any more:
> 
> ```json
> "lsp": {
>     "elixir-ls": {
>       "settings": {
>         "dialyzerEnabled": false
>       }
>     }
>   },
> ```
> 
> Should this config go somewhere else now?

This should be fixed by https://github.com/zed-industries/zed/pull/11302.

Just published v0.0.2 of the extension with the fix.

moomerman:
v0.0.2 Works great, thank you!

baldwindavid:
I was attempting to mess with the Elixir extension utilizing the "Install Dev Extension" button as that seems to work for, say, Git Firefly...

Screenshot 2024-05-05 at 3 55 44 PM

However, it doesn't appear to work for the Elixir directory. Selecting the `elixir` directory results in the "Install" link for the Elixir extension to be disabled for a few seconds, but then enabled again with nothing reported. Seems to be the same situation for Elm so assuming there is some other flow in order to update these extensions that are part of the main repo.

maxdeviant:
> I was attempting to mess with the Elixir extension utilizing the "Install Dev Extension" button as that seems to work for, say, Git Firefly...
> 
> Screenshot 2024-05-05 at 3 55 44 PM
> However, it doesn't appear to work for the Elixir directory. Selecting the `elixir` directory results in the "Install" link for the Elixir extension to be disabled for a few seconds, but then enabled again with nothing reported. Seems to be the same situation for Elm so assuming there is some other flow in order to update these extensions that are part of the main repo.

Hmm, it sounds like the extension might be failing to build.

Do you see anything listed in the Zed logs?

baldwindavid:
Seems to not be specific to Elixir so posted https://github.com/zed-industries/zed/issues/11435 in the main repo.

baldwindavid:
The issue was that I had two installations of rust: rustup and homebrew. https://github.com/zed-industries/zed/issues/11435 has been closed and I need to learn me some rust ha

PR #649, opened 2024-04-30
Add dockerignore extension

eth0net:
This PR brings a new extension to support syntax highlighting for `.dockerignore` files.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -90,6 +90,10 @@
 	path = extensions/dockerfile
 	url = https://github.com/d1y/dockerfile.zed.git
 
+[submodule "extensions/dockerignore"]
+	path = extensions/dockerignore
+	url = https://github.com/eth0net/zed-dockerignore.git
+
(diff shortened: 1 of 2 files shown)


d1y:
I think it would be better if the dockerignore syntax highlighting could be put here https://github.com/d1y/dockerfile.zed
:)

eth0net:
> I think it would be better if the dockerignore syntax highlighting could be put here https://github.com/d1y/dockerfile.zed :)

I also have a Docker Compose extension, would be better to combine them all but I originally avoided adding to your dockerfile extension since ideally it would be a new `Docker` extension. I had also forgotten about dockerignore at the time or I'd have just started on the combined one then 🤦🏻‍♀️

maxdeviant:
@eth0net Did you still want to publish this extension now that `git-firefly` supports `.dockerignore`?

eth0net:
Perhaps not, it might be worth condensing it into a whole docker extension at some point but for now let's close this.

PR #650, opened 2024-04-30
Update git-firefly to v0.0.3

d1y:
In #649, The PR update new style with extension format(old use .json) and gitignore path_suffixes contains .dockerignore

/cc @eth0net

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -185,7 +185,7 @@ version = "0.0.2"
 
 [git-firefly]
 submodule = "extensions/git-firefly"
-version = "0.0.2"
+version = "0.0.3"
 
 [github-dark-default]
 submodule = "extensions/github-dark-default"


maxdeviant (changes requested):
@d1y The `grammars/` directory should be added to `.gitignore` when using the `extension.toml` format.

We don't want the grammar repos referenced as submodules.

maxdeviant:
@d1y Could you also remove the `grammars` directory from Git (since it looks like it was committed before `grammars` was added to the `.gitignore`)?

d1y:
> @d1y Could you also remove the `grammars` directory from Git (since it looks like it was committed before `grammars` was added to the `.gitignore`)?

Is that what you mean?

```bash
git rm -r --cached grammars/
git commit --amend --no-edit
git push -f
```

maxdeviant:
> > @d1y Could you also remove the `grammars` directory from Git (since it looks like it was committed before `grammars` was added to the `.gitignore`)?
> 
> Is that what you mean?
> 
> ```shell
> git rm -r --cached grammars/
> git commit --amend --no-edit
> git push -f
> ```

I believe that should work!

But perhaps something more is needed in the case of submodules? Although I think deleting the directory from Git would also remove the submodules in the `grammars/` directory.

maxdeviant:
> Sorry, it took so long. Is it OK now? @maxdeviant

Could you change the `id` in `extension.toml` back to `git-firefly`?

https://github.com/d1y/git_firefly/blob/7f3a8b79641aac2bcd047cec46d357ad54b1c15b/extension.toml#L1

Right now it is `git_firefly`.

maxdeviant (approved):

PR #651, opened 2024-04-30
Add SCSS extension

bajrangCoder:
Added scss language support for Zed(https://github.com/zed-industries/zed/issues/5356)

Note: I have not found any language server for scss yet :(

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -529,3 +529,6 @@
 [submodule "extensions/zedwaita"]
 	path = extensions/zedwaita
 	url = https://github.com/someone13574/zed-adwaita-theme.git
+[submodule "extensions/zed-scss"]
+	path = extensions/zed-scss
+	url = https://github.com/bajrangCoder/zed-scss.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -518,6 +518,10 @@
 	path = extensions/zed
 	url = https://github.com/zed-industries/zed.git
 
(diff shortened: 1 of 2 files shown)


noClaps:
The [VSCode Language Server for CSS](https://github.com/microsoft/vscode-css-languageservice) also has support for SCSS, so you should be able to use that.

bajrangCoder:
> The [VSCode Language Server for CSS](https://github.com/microsoft/vscode-css-languageservice) also has support for SCSS, so you should be able to use that.

Then I thought not to add LSP through extension because LSP for css is already built in inside zed so user can configure that for scss too(I'm not sure that it's possible or not)

maxdeviant:
@bajrangCoder Thank you!

Otterpocket:
@bajrangCoder ⭐️

amaanq:
thanks for using the ts-grammar org's grammar! glad our org is benefitting users downstream 😁

cseelus:
Thank you!

Does it support `.sass` files (indented syntax) too?

bajrangCoder:
> Thank you!
> 
> Does it support `.sass` files (indented syntax) too?

No

cseelus:
Gotcha, found [this issue on the projects repo](https://github.com/bajrangCoder/zed-scss/issues/3) and the information that this will be a separate extionsion.

PR #706, opened 2024-05-10
Update Swift to v0.2.0

ejjonny:


maxdeviant:
@tonycoco Please don't tag Zed employees like this.

It doesn't make things move any faster.

PR #729, opened 2024-05-15
Add VHDL extension

rapgenic:
This PR implements an extension for the VHDL language using

* https://github.com/alemuller/tree-sitter-vhdl for the tree-sitter grammar
* https://github.com/VHDL-LS/rust_hdl as the language server
* the zed csharp extension as a reference

This is my first attempt at creating an extension, so any feedback is appreciated!

I'm also marking this as a draft, because I cannot test MacOS and Windows, plus I'd like to use it for a few days to iron out any problems (I wrote it in a few hours, so there might be some rough mistakes...).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -601,3 +601,7 @@
 [submodule "extensions/zedwaita"]
 	path = extensions/zedwaita
 	url = https://github.com/someone13574/zed-adwaita-theme.git
+
+[submodule "extensions/zed-vhdl"]
+	path = extensions/zed-vhdl
+	url = https://github.com/rapgenic/zed-vhdl
--- a/extensions.toml
+++ b/extensions.toml
@@ -663,6 +663,10 @@ version = "0.0.1"
 submodule = "extensions/vesper"
 version = "0.0.1"
 
+[vhdl]
+submodule = "extensions/vhdl"
+version = "0.0.1"
+
 [vhs]
 submodule = "extensions/vhs"
 version = "0.1.0"


rapgenic:
I am having some problems with symbol outline and multiple variable definitions in a single statement, as per the following example:

```vhdl
variable a, b : Integer;
```

When showing the outline, the `b` variable appears as a child of the `a` variable, see the following picture:

image

The current tree-sitter query for that node is:

```scm
(variable_declaration
    "variable" @context
    (identifier_list (identifier) @name)
) @item
```

I cannot find a way to make them appear as two distinct variables, with the correct alignment...

NickelLiang:
Propose to use a newer VHDL tree-sitter implementation [here](https://github.com/jpt13653903/tree-sitter-vhdl), instead of the one that seems to be unmaintained in the past years.

rapgenic:
Thank you for the suggestion! I will have a look at it, but since it looks like it would be a complete rewrite of all the scm files, I am not sure it is worth the time! (The extension is basically finished, just waiting for zed-industries/zed#12739) 

Is it solving some problem with the other grammar or better in some way?

Anyway I'm not against changing it, I just don't have much time ATM. If someone wanted to do the work I'd gladly accept a PR unless there is some regression

rapgenic:
As a quick note, I have done a small test with the suggested new tree-sitter grammar and ATM I wasn't able to obtain an acceptable result. These are the problems I found:
* Grammar does not load properly in zed as is due to the use of the `printf` function:
    ```
    [2024-06-27T22:15:26+02:00 ERROR language::language_registry] failed to load language VHDL:
    Failed to instantiate wasm module: invalid import 'printf'
    ```
* Even removing all the `printf` calls from the source code, the parser seems to be a quite buggy (90% of source codes that worked with the previous grammar did not work with the suggested one, i.e. the tree view shows an ERROR node)

I did not investigate further, but since I have not found limitations with the older grammar and since the proposed grammar seems incomplete, I am going to keep using the old one for now.

maxdeviant:
@rapgenic Since there hasn't been any activity on this PR in a few months, I'm going to close it, for now.

Feel free to reopen if you pick it up again.

PR #738, opened 2024-05-17
Update Mau theme to 0.0.2

mauscoelho:


mauscoelho:
Any idea why this doesn't work?

valentinegb:
In the newest PR commit, the submodule isn't updated. It looks like it's currently using commit a1357bf which is from 3 months ago.

mauscoelho:
> In the newest PR commit, the submodule isn't updated. It looks like it's currently using commit a1357bf which is from 3 months ago.

Thanks, I cloned with `--recurse-submodules` flag to be able to update the submodule.

PR #739, opened 2024-05-17
Add neocmake extension

k0tran:
Tested under Manjaro Linux x86_64 Zed 0.135.2

For tree sitter used [uyha/tree-sitter-cmake](https://github.com/uyha/tree-sitter-cmake)\
I'm not proficient with tree-sitter grammar so I took it from [helix](https://github.com/helix-editor/helix/tree/master/runtime/queries/cmake). Used a little higher tree-sitter version though (v0.3.0 instead of pre-v0.2.0 commit)

For LSP used [Decodetalkers/neocmakelsp](https://github.com/Decodetalkers/neocmakelsp) (that's why extension called neocmake). Another option would be [regen100/cmake-language-server](https://github.com/regen100/cmake-language-server), but it's written in python and says "Alpha Stage, work in progress." in the README.md

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -306,6 +306,10 @@
 	path = extensions/navi
 	url = https://github.com/navi-language/zed-navi.git
 
+[submodule "extensions/neocmake"]
+	path = extensions/neocmake
+	url = https://github.com/k0tran/zed_neocmake.git
+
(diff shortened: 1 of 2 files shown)


maxdeviant:
Thank you!

PR #744, opened 2024-05-17
Update XML to v0.0.2

sweetppro:


maxdeviant (changes requested):
Could we revert the changes to the other submodules?

It should only be `extensions/xml` that gets modified.

sweetppro:
@maxdeviant 
> Could we revert the changes to the other submodules?
> 
> It should only be `extensions/xml` that gets modified.

how would I go about this,
my git skills are very basic

sweetppro:
@maxdeviant done

PR #753, opened 2024-05-20
Update serendipity to 0.0.2

meocoder31099:
Fixed format string coloring issues
Added new theme variants Serendipity Midnight


meocoder31099:
I think that #24 should also be closed

PR #773, opened 2024-05-22
Update PHP to v0.0.4

RemcoSmitsDev:
Support for https://github.com/zed-industries/zed/pull/12140


maxdeviant:
@RemcoSmitsDev We need to bump the `extensions/zed` submodule as well so it has the changes from https://github.com/zed-industries/zed/pull/12140.

PR #785, opened 2024-05-25
Add F# extension

### LINKED ISSUE #141: F#
F# language support would be great, especially if implemented alongside C# language support given the interoperability between the languages (#293).

**Language Server Options**
In terms of language servers, there are two main options:
- [fsharp-language-server](https://github.com/fsprojects/fsharp-language-server)
- [FsAutoComplete](https://github.com/fsharp/FsAutoComplete)
both of which are built on top of the official [F# Compiler Service](https://www.nuget.org/packages/FSharp.Compiler.Service).

FsAutoComplete is the option used in Ionide which is the most popular option for F# editing outside of Visual Studio and Rider, so this is probably the most sensible choice.

**Note/Advice**
Just a note on something that is often forgotten with F# editor integrations at first, and that is that in F# projects, file order is important. This means a given file can only reference code in files "before" it. This order is described in the `.fsproj` file, and so it would be nice to have Zed display the source files according to this correct order when a folder is open which includes an F# project.

Hope this info helps :) Keep up the amazing work!

nathanjcollins:
This is a start to F# support in Zed. I have copied the `.scm` files from [here](https://github.com/ionide/tree-sitter-fsharp) and have fsautocomplete working.

Automatic installation doesn't seem to be possible as the extension api does not support installing through dotnet tools, but maybe I can implement this manually? I wasn't sure if this was an accepted way of doing it.

image
I've come across one issue which seems to break fsautocomplete because there is a hidden character (FEFF) - which you can manually delete and fsautocomplete works as usual. Unsure if this is a me issue or a Zed issue.

Happy to take suggestions, as I'm pretty new to all this.

Would close #141 if accepted

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -170,6 +170,10 @@
 	path = extensions/fleet-themes
 	url = https://github.com/skarline/zed-fleet-themes.git
 
+[submodule "extensions/fsharp"]
+	path = extensions/fsharp
+	url = https://github.com/nathanjcollins/zed-fsharp.git
+
 [submodule "extensions/gdscript"]
 	path = extensions/gdscript
 	url = https://github.com/grndctrl/zed-gdscript.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -158,6 +158,10 @@
 	path = extensions/fleet-themes
 	url = https://github.com/skarline/zed-fleet-themes.git
 
+[submodule "extensions/fsharp"]
+	path = extensions/fsharp
+	url = https://github.com/nathanjcollins/zed-fsharp.git
+
 [submodule "extensions/gdscript"]
 	path = extensions/gdscript
 	url = https://github.com/grndctrl/zed-gdscript.git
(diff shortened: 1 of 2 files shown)


nathanjcollins:
@cla-bot check

maxdeviant:
> Automatic installation doesn't seem to be possible as the extension api does not support installing through dotnet tools, but maybe I can implement this manually?

If there is an installation method needed outside of downloading from a GitHub release or installing via npm package, then we typically rely on the user installing it out-of-band and then have Zed detect it from the path, like we do here:

https://github.com/zed-industries/zed/blob/a259042f923b7edf5c104b73834b830a550d102c/extensions/csharp/src/csharp.rs#L14-L16

maxdeviant:
> > Automatic installation doesn't seem to be possible as the extension api does not support installing through dotnet tools, but maybe I can implement this manually?
> 
> If there is an installation method needed outside of downloading from a GitHub release or installing via npm package, then we typically rely on the user installing it out-of-band and then have Zed detect it from the path, like we do here:
> 
> https://github.com/zed-industries/zed/blob/a259042f923b7edf5c104b73834b830a550d102c/extensions/csharp/src/csharp.rs#L14-L16

I see now that is what you've already done!

PR #790, opened 2024-05-26
Update serendipity to v0.0.4

meocoder31099:
- add new theme variant Midnight Minimal
- add new no italic theme variant
- Increase contrast
- Add color support for indent guides (Zed v0.138.0-pre and above)

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -578,7 +578,7 @@ version = "0.0.3"
 
 [serendipity]
 submodule = "extensions/serendipity"
-version = "0.0.4
+version = "0.0.4"
 
 [siri]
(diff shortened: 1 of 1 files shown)


maxdeviant:
> > It looks like now there are two different theme files that both define a theme family with the name `Serendipity`.
> > Could you either combine these all into a single file or change one of the family names to be different?
> 
> Is it mandatory to use different names? Basically, both files are just different variations of `Serendipity`. I separated them into two files to make management and maintenance easier.

Today I think what you are doing will work, but it's not how the theme system was designed, and it will cause problems for your theme down the line.

maxdeviant (approved):

PR #809, opened 2024-05-29
💾 Add Green Monochrome Monitor CRT Phosphor theme

Takk8IS:
This PR adds the Green Monochrome Monitor CRT Phosphor theme for Zed.

[Green Monochrome Monitor CRT Phosphor Dark](https://github.com/Takk8IS/green-monochrome-monitor-crt-phosphor-theme-for-zed/blob/main/README.md)

PR #812, opened 2024-05-30
💾 Add Green Monochrome Monitor CRT Phosphor theme

Takk8IS:
📺 This PR adds the Green Monochrome Monitor CRT Phosphor for Zed.


maxdeviant (changes requested):
This PR has the same issues as #809.

PR #820, opened 2024-05-31
Add CSS language support

filipjanevski:


maxdeviant:
Hey @filipjanevski, thanks for opening this PR.

CSS support is currently built-in to Zed, so I don't think we need an extension for it right now.

If we do decide to move the support into an extension, we'd want to start with the base support from the current Zed implementation.

PR #823, opened 2024-05-31
Update pact to v0.0.2

salamaashoush:


maxdeviant (changes requested):
It looks like there's a problem with one of the Tree-sitter queries:

```
Error: Query error at 28:4. Invalid node type set
```

salamaashoush:
> It looks like there's a problem with one of the Tree-sitter queries:
> 
> ```
> Error: Query error at 28:4. Invalid node type set
> ```

Yes I fixed it thanks

maxdeviant (approved):

PR #833, opened 2024-06-02
Add frosted-theme extension

daviiiL:
This theme is a frosted glass style theme. Would love to publish to the extensions store.


maxdeviant (approved):
Thank you!

PR #848, opened 2024-06-03
Update Siri to 0.0.5

perragnar:


perragnar:
Hmm, I've missed the memo about changing to extension.toml. I will change the file and try again.

PR #857, opened 2024-06-04
Add Nix extension with LSP

msanft:
There is basic support for the Nix language through the [`nix`](https://github.com/hasit/zed-nix) extension, but it is unmaintained and doesn't have LSP support. 

This adds a more feature-rich extension with better completion and LSP support. I thought about replacing the previous Nix extension, but didn't want to cause any breakage for existing installations. Please let me know what'd be the preferred way to go about this.

See the LSP features, such as recognizing unused arguments:
image

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -394,6 +394,10 @@
 	path = extensions/nix
 	url = https://github.com/hasit/zed-nix.git
 
+[submodule "extensions/nix-language"]
+	path = extensions/nix-language
+	url = https://github.com/msanft/nix-zed
+
 [submodule "extensions/noir"]
 	path = extensions/noir
 	url = https://github.com/shuklaayush/zed-noir.git
--- a/extensions.toml
+++ b/extensions.toml
@@ -465,6 +465,10 @@ version = "0.0.1"
 submodule = "extensions/nix"
 version = "0.0.4"
 
+[nix-language]
+submodule = "extensions/nix-language"
+version = "0.0.1"
+
 [noir]
 submodule = "extensions/noir"
 version = "0.0.1"


msanft:
@cla-bot check

maxdeviant (changes requested):
Thanks for opening this, @msanft!

We would like to avoid having multiple extensions doing similar things if possible.

I think contributing language server support to the existing Nix extension would be a good path to pursue.

We could also move ownership of the Nix extension into the @zed-extensions organization. We'd need to talk to the current maintainer about that.

msanft:
> We would like to avoid having multiple extensions doing similar things if possible.

Totally reasonable.

> I think contributing language server support to the existing Nix extension would be a good path to pursue.

Unfortunately, the current maintainer seems quite unresponsive. There is [an issue](https://github.com/hasit/zed-nix/issues/1) for LSP support, but there's no activity from him.

> We could also move ownership of the Nix extension into the @zed-extensions organization. We'd need to talk to the current maintainer about that.

That might be a good idea, if you can get him to answer.

Otherwise, the extension I built only extends what the previous extension did, except for handling bracket completions differently. I'm unsure whether this minor, but breaking change is enough of a reason to not replace the current extension.

maxdeviant:
I have opened https://github.com/hasit/zed-nix/issues/5 to discuss the maintenance of the Nix extension.

We can wait to see if there is any movement on that. If I don't receive a reply within the next few weeks, alternative solutions will be pursued.

I'm going to close this PR, as we don't want multiple Nix extensions, as I mentioned before.

If you open a PR to https://github.com/hasit/zed-nix with language server support, either the current maintainer can merge it, or I can if the repository gets transferred to @zed-extensions.

hasit:
@maxdeviant I have transferred the repo over to you.

maxdeviant:
@msanft The new repo can be found at https://github.com/zed-extensions/nix, if you'd like to put up a PR with language server support.

PR #881, opened 2024-06-07
Add Tokyo Night Storm extensions

ZeroHawkeye:


maxdeviant:
How does this differ from the Tokyo Night Storm theme in the `tokyo-night` extension?

https://github.com/ssaunderss/zed-tokyo-night/blob/fb2c765772a1096c52d44ece746f661460bed9e8/themes/tokyo-night.json#L615-L616

ZeroHawkeye:
In my golang language, such as the routing file of the gin framework, the individual func handler will not be highlighted, but this theme will be highlighted

maxdeviant (changes requested):
As-is, I think having both "Tokyo Night" and a separate "Tokyo Night Storm" extension is going to be confusing for users.

If you give your extension a different name then I think we can move forward with it.

PR #927, opened 2024-06-12
Update AUTHORING_EXTENSIONS.md to use the command in package.json

trbutler4:
While authoring an extension i noticed this says to use "pnpm sort-extensions" but from the package.json it appears to be "pnpm fmt"


maxdeviant:
These are two different commands.

`pnpm fmt` just formats the actual source files for the packaging system with Prettier.

`pnpm sort-extensions` sorts the `extensions.toml` and `.gitmodules` file.

There is no change needed here.

PR #930, opened 2024-06-13
Update monosami to v0.1.1

borngraced:
fix: light theme ghost element selected contrast

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -377,3 +377,6 @@
 [submodule "extensions/zedspace"]
 	path = extensions/zedspace
 	url = https://github.com/Brunowilliang/zedspace.git
+[submodule "extensions/monosami"]
+	path = extensions/monosami
+	url = https://github.com/borngraced/monosami.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -174,6 +174,10 @@
 	path = extensions/modest-dark
 	url = https://github.com/timcole/modest-dark.git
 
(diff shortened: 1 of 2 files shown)


maxdeviant (changes requested):
@borngraced We now require that extensions use the `extension.toml` manifest format instead of `extension.json`.

Would you mind updating your extension to use `extension.toml`?

borngraced:
done

maxdeviant (changes requested):
@borngraced It looks like the ID in `extension.toml` changed.

We'll want to keep it as `monosami`.

borngraced:
??

maxdeviant:
> ??

I didn't know you had made the change, since the submodule was never updated in this PR.

It looks like the version in `extension.toml` still needs to be updated to `0.1.1`: https://github.com/borngraced/monosami/blob/7527efaf622234c8f4fbd8ec5afe7cc8481f6dc3/extension.toml#L3

borngraced:
> > ??
> 
> I didn't know you had made the change, since the submodule was never updated in this PR.
> 
> It looks like the version in `extension.toml` still needs to be updated to `0.1.1`: https://github.com/borngraced/monosami/blob/7527efaf622234c8f4fbd8ec5afe7cc8481f6dc3/extension.toml#L3

ah okay got it..

maxdeviant (approved):

PR #937, opened 2024-06-17
Add Adwaita Pastel Theme

Benjamin-Davies:
Adds a theme based on the default GNOME apps but with more exciting syntax highlighting.

This theme differs from Zedwaita in that it provides better contrast for some frequently-used UI elements and takes a bolder approach to colors.


Benjamin-Davies:
@cla-bot check

maxdeviant (approved):
Thank you!

PR #964, opened 2024-06-24
Update extensions.toml

nooooaaaaah:
added a light theme


nooooaaaaah:
forgot to change to scrollbar.thumb.background in all places

PR #998, opened 2024-07-04
Add QML language support

lkroll:
This PR adds language support for QML files.


a-andreyev:
thanks! works for me via "install dev extention" with zed version 0.143.7 from arch repos:
image

maxdeviant (approved):
Thanks!

PR #1042, opened 2024-07-12
Add catppuccin-blur theme

jenslys:
A fork of the [Catppuccin](https://github.com/catppuccin/zed) that includes 2 custom variants.
Repo: https://github.com/jenslys/zed-catppuccin-blur

##  Preview

Latte Preview

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -107,7 +107,7 @@
 	url = https://github.com/catppuccin/zed.git
 
 [submodule "extensions/catppuccin-blur"]
-	path = extensions/zed-catppuccin-blur
+	path = extensions/catppuccin-blur
 	url = https://github.com/jenslys/zed-catppuccin-blur.git
 
 [submodule "extensions/cedar"]
--- a/.gitmodules
+++ b/.gitmodules
@@ -106,6 +106,10 @@
 	path = extensions/catppuccin
 	url = https://github.com/catppuccin/zed.git
 
+[submodule "extensions/catppuccin-blur"]
+	path = extensions/zed-catppuccin-blur
+	url = https://github.com/jenslys/zed-catppuccin-blur.git
+
 [submodule "extensions/cedar"]
 	path = extensions/cedar
 	url = https://github.com/chrnorm/zed-cedar.git
--- a/extensions.toml
+++ b/extensions.toml
@@ -112,6 +112,10 @@ version = "0.0.1"
 submodule = "extensions/catppuccin"
 version = "0.2.8"
 
+[catppuccin-blur]
+submodule = "extensions/catppuccin-blur"
+version = "0.0.1"
+
 [cedar]
 submodule = "extensions/cedar"
 version = "0.0.1"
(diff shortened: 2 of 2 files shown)


maxdeviant (changes requested):
Can we use a different name for the extension?

Right now it's just "Catppuccin Themes", which is the same as the official Catppuccin extension, which I think is going to cause some confusion:

https://github.com/jenslys/zed-catppuccin-blur/blob/f58a8b5b602c253525822b76be6018c0f596856d/extension.toml#L2

maxdeviant:
> Can we use a different name for the extension?
> 
> Right now it's just "Catppuccin Themes", which is the same as the official Catppuccin extension, which I think is going to cause some confusion:
> 
> https://github.com/jenslys/zed-catppuccin-blur/blob/f58a8b5b602c253525822b76be6018c0f596856d/extension.toml#L2

Alternatively, we could see if the Catppuccin folks would be amenable adding these two theme variants to the main Catppuccin extension.

jenslys:
> Can we use a different name for the extension?
> 
> Right now it's just "Catppuccin Themes", which is the same as the official Catppuccin extension, which I think is going to cause some confusion:
> 
> [jenslys/zed-catppuccin-blur@`f58a8b5`/extension.toml#L2](https://github.com/jenslys/zed-catppuccin-blur/blob/f58a8b5b602c253525822b76be6018c0f596856d/extension.toml#L2)

How about just calling it "Catppuccin Blur"?

Seeing the amount of catppuccin variant forks on vscode, i doubt they will be open to merging the blur variants

maxdeviant:
> > Can we use a different name for the extension?
> > Right now it's just "Catppuccin Themes", which is the same as the official Catppuccin extension, which I think is going to cause some confusion:
> > [jenslys/zed-catppuccin-blur@`f58a8b5`/extension.toml#L2](https://github.com/jenslys/zed-catppuccin-blur/blob/f58a8b5b602c253525822b76be6018c0f596856d/extension.toml#L2)
> 
> How about just calling it "Catppuccin Blur"?
> 
> Seeing the amount of catppuccin variant forks on vscode, i doubt they will be open to merging the blur variants, since i also have other theme changes, besides just making the BG blurry
> 
> image

Yep, "Catppuccin Blur" sounds good!

jenslys:
> > > Can we use a different name for the extension?
> > > Right now it's just "Catppuccin Themes", which is the same as the official Catppuccin extension, which I think is going to cause some confusion:
> > > [jenslys/zed-catppuccin-blur@`f58a8b5`/extension.toml#L2](https://github.com/jenslys/zed-catppuccin-blur/blob/f58a8b5b602c253525822b76be6018c0f596856d/extension.toml#L2)
> > 
> > 
> > How about just calling it "Catppuccin Blur"?
> > Seeing the amount of catppuccin variant forks on vscode, i doubt they will be open to merging the blur variants, since i also have other theme changes, besides just making the BG blurry
> > image
> 
> Yep, "Catppuccin Blur" sounds good!

Updated

maxdeviant (approved):
Thanks!

PR #1052, opened 2024-07-14
Add becker theme

ErnstBecker:
Adds a new theme extension https://github.com/Becker-Theme/zed

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -50,6 +50,10 @@
 	path = extensions/beancount
 	url = https://github.com/zed-extensions/beancount.git
 
+[submodule "extensions/becker-theme"]
+	path = extensions/becker-theme
+	url = https://github.com/Becker-Theme/zed.git
+
 [submodule "extensions/bend"]
 	path = extensions/bend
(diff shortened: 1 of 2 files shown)


maxdeviant (approved):
@ErnstBecker Would you mind updating your `extension.toml` to remove "Zed" from the ID and name?

Since everything in the extension registry is for Zed, we don't need to specify.

https://github.com/Becker-Theme/zed/blob/37acac931222ddc6e0128d100c7526cbaf60c73f/extension.toml#L1-L2

```toml
id = "becker-theme"
name = "Becker Theme"
```

maxdeviant (changes requested):
@ErnstBecker Would you mind updating your `extension.toml` to remove "Zed" from the ID and name?

Since everything in the extension registry is for Zed, we don't need to specify.

https://github.com/Becker-Theme/zed/blob/37acac931222ddc6e0128d100c7526cbaf60c73f/extension.toml#L1-L2

```toml
id = "becker-theme"
name = "Becker Theme"
```

ErnstBecker:
@maxdeviant Changed the id and name of the theme! Thanks for the help!

PR #1056, opened 2024-07-14
Update GDScript

grndctrl:
Adds compatibility for windows.

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -139,7 +139,7 @@ version = "1.0.0"
 
 [gdscript]
 submodule = "extensions/gdscript"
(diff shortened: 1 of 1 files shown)


grndctrl:
Weird, it shows some older commits that should've been merged earlier. How do I resolve the conflict of that older commit?

maxdeviant:
> Weird, it shows some older commits that should've been merged earlier. How do I resolve the conflict of that older commit?

You would need to reset `main` on your fork with `main` on the upstream and then redo your changes.

Generally you'd want to make a new branch for each PR so that this doesn't occur.

PR #1069, opened 2024-07-16
Update Uno Flat Dark theme to v0.1.1

bryanbuchanan:


maxdeviant (changes requested):
We need to update the version in `extensions.toml` as well, otherwise the update won’t get picked up for publishing.

maxdeviant (approved):

PR #1085, opened 2024-07-18
add embeded-template extension

dangh:


maxdeviant (changes requested):
@dangh Would it make sense to have a dedicated extension for EJS files?

The Ruby extension already comes with support for ERB files: https://github.com/zed-industries/zed/blob/272be98ec854b0cd240fe392101cf02a207c29df/extensions/ruby/languages/erb/config.toml#L1-L3

dangh:
@maxdeviant I have created another PR for EJS #1102

PR #1107, opened 2024-07-21
Update GDScript to v0.3.1

grndctrl:
Added experimental windows compatibility.

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -298,7 +298,7 @@ version = "0.0.3"
 
 [gdscript]
 submodule = "extensions/gdscript"
-version = "0.2.0"
+version = "0.3.0"
 
 [gemini]
 submodule = "extensions/gemini"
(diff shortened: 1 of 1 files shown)


maxdeviant (changes requested):
@grndctrl The way we detect the platform needs to be adjusted.

Right now we're using the `cfg!` macro:

https://github.com/grndctrl/zed-gdscript/blob/6a5d7afc7abf0121dd03a6cba64d3abfce99158b/src/gdscript.rs#L16

However, because the extensions are compiled to Wasm, this means that there is a single Wasm bundle that is used on all the platforms.

In order to write platform-specific code, you'll need to use the [`current_platform`](https://docs.rs/zed_extension_api/latest/zed_extension_api/fn.current_platform.html) function to retrieve the current platform from the host and then match on the resulting enum.

Here's an example of where we do that: https://github.com/zed-industries/zed/blob/0019b65ce1d8d9c6e0499889429ddaaed8708258/extensions/gleam/src/gleam.rs#L45-L59

maxdeviant (approved):

PR #1128, opened 2024-07-24
Update New Darcula theme to v1.0.6

e-simpson:
- Fixed legibility of terminal auto-completions thanks to @edgarkanyes
- Updated to scrollbar.thumb.background thanks to @nishantdd


e-simpson:
I see extensions now need to use a .toml extensions file .. updating ..

maxdeviant (approved):

PR #1129, opened 2024-07-25
Update Monospace theme

Abhinav5383:
Updated the version number from 0.1.1 to 0.1.2

Changes:
- Fixed the attribute color being the same as JSX tag color

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -559,7 +559,7 @@ version = "0.1.0"
 
 [monospace-theme]
 submodule = "extensions/monospace-theme"
-version = "0.1.1"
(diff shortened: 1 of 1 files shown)


Abhinav5383:
Can i get some help,
I don't know what's wrong
The version number in the `extensions.toml` matches the number in the submodule's `extension.toml` but it still says version mismatch

the error
```
Error: Incorrect version for extension monospace-theme (Monospace Theme)
Expected version: 0.1.2
Actual version: 0.1.1
```

I have double checked it, the version in both places is `0.1.2`.

PR #1160, opened 2024-07-28
Add Curry extension

fwcd:
This adds [`zed-curry`](https://github.com/fwcd/zed-curry), an extension providing language support for the [Curry programming language](https://en.wikipedia.org/wiki/Curry_(programming_language)).

Screenshot 2024-07-28 at 23 13 56


maxdeviant (approved):
Thanks!

PR #1175, opened 2024-07-31
add extension git branch controller

Betzalel75:


maxdeviant:
Hey @Betzalel75, this doesn't look like a valid Zed extension.

I would suggest looking at the [extension authoring docs](https://github.com/zed-industries/extensions/blob/843066ec6e31a642f5f3f25c0ff3469caedca6b7/AUTHORING_EXTENSIONS.md) to see what is currently possible to do in extensions and how to go about creating one.

PR #1222, opened 2024-08-08
Add LTeX extension

vitallium:
LTeX LSP implements a language server
according to the Language Server Protocol (LSP)
and provides grammar and spelling errors
in markup documents: LATEX, Markdown, etc..
The documents are checked with LanguageTool.

This is how it works in action:

CleanShot 2024-08-08 at 21 54 29@2x
CleanShot 2024-08-08 at 21 55 40@2x

See also:
- https://github.com/valentjn/ltex-ls

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -450,6 +450,10 @@
 	path = extensions/lox
 	url = https://github.com/arian81/zed-lox.git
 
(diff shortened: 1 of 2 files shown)


maxdeviant (approved):
Thanks!

PR #1242, opened 2024-08-12
Update AUTHORING_EXTENSIONS.md

rgbkrk:
Update the zed extension API version to 0.0.6 in the likely block people will copy (myself included).


maxdeviant (approved):
Thanks!

PR #1253, opened 2024-08-15
Add Fennel extension

mattly:
Does not currently support the LSP, because AFAICT Zed only supports LSPs hosted on Github

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1057,6 +1057,3 @@
 [submodule "extensions/zedwaita"]
 	path = extensions/zedwaita
 	url = https://github.com/someone13574/zed-adwaita-theme.git
-[submodule "extensions/zed-lang-fennel"]
-	path = extensions/zed-lang-fennel
-	url = https://codeberg.org/mattly/zed-lang-fennel.git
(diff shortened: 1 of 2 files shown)


maxdeviant:
> so I suspect this failed because of this line in the workflow file: https://github.com/zed-industries/extensions/actions/runs/10411662361/workflow?pr=1253#L68
> 
> I'm not really familiar with Github's CI, so I'm guessing here
> 
> And it's failing because my repository is not hosted on Github.
> 
> So, I could mirror it here, but the fennel community is taking more of an older-school OSS approach to how they distribute and work on things; while the language has a copy of the repository here, most development takes place on SourceHut, including the good language server.
> 
> I suspect a lot of the decisions lashing Zed to Github for things were probably made in a expedient "it's easy and solves most problems" manner, but if it is going to become a truly "open" editor, it needs to support extensibility uses that involve resources hosted elsewhere.

It's actually failing because there is a typo in the `extensions.toml` file: https://github.com/zed-industries/extensions/pull/1253/files#diff-724e5177a66640ebdc9d051879ac354bbfd595c1b4d039bab887600964bb2023R312

PR #1257, opened 2024-08-16
Update neocmake extension to v0.0.5

k0tran:
- Fix lsp path for linux
- Add aarch64 support for MacOS

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -639,7 +639,7 @@ version = "0.0.2"
 
 [neocmake]
 submodule = "extensions/neocmake"
-version = "0.0.2"
+version = "0.0.4"
 
 [neosolarized]
 submodule = "extensions/neosolarized"
--- a/extensions.toml
+++ b/extensions.toml
@@ -639,7 +639,7 @@ version = "0.0.2"
 
 [neocmake]
 submodule = "extensions/neocmake"
-version = "0.0.4"
+version = "0.0.5"
 
 [neosolarized]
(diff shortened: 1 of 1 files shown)


k0tran:
@cla-bot check

maxdeviant:
@k0tran It looks like in this version the code to remove older language servers was removed:

https://github.com/k0tran/zed_neocmake/compare/7cab1dd03e0f5055f253571778c571004aed0f4d...376321432a945fcd91c5aac008ff1b26fa9dcf29#diff-7a58343a9453f7caca008e4a9375027a58151234c9537ed79f44078fcb824343L74-L81

Was there a particular reason for this?

k0tran:
> Was there a particular reason for this?

When I first saw this code I thought it is for removing everything except lsp binary in the archive. And since neocmake-lsp is just a single binary - I removed it. If it's also removing old lsp versions then I can bring it back for that purpose

maxdeviant:
> > Was there a particular reason for this?
> 
> When I first saw this code I thought it is for removing everything except lsp binary in the archive. And since neocmake-lsp is just a single binary - I removed it. If it's also removing old lsp versions then I can bring it back for that purpose

Yea, that code is there for removing all but the current version of the language server, so that they don't accumulate and take up space.

k0tran:
@maxdeviant Updated extension, thanks for the feedback!

maxdeviant (approved):

PR #1289, opened 2024-08-23
Add pylsp extension

rgbkrk:
Add https://github.com/rgbkrk/python-lsp-zed-extension as another Python LSP option. Motivated by requests and pain points in https://github.com/zed-industries/zed/issues/7296.


maxdeviant (approved):
Nice!

PR #1296, opened 2024-08-24
Add KDL language extension

elkowar:
Adds language support for the KDL configuration language.


extensions.toml
```diff
@@ -536,6 +536,10 @@ version = "0.1.1"
 submodule = "extensions/kanagawa-themes"
 version = "0.0.1"
 
+[kdl-syntax]
```
maxdeviant, line 539: Any reason to not call this just `kdl` instead of `kdl-syntax`?

elkowar: I thought there might at some point be a more fully featured kdl extension with some LSP or something, but I guess for now just calling it kdl works fine. I'll rename it!

elkowar: Done

maxdeviant (approved):
Thanks!

PR #1299, opened 2024-08-26
Update Vitesse theme to v0.0.2

musicq:
Makes hints more readable


maxdeviant (changes requested):
@musicq It looks like the Vitesse extension is still using the old `extension.json` format.

We'll need to update it to use `extension.toml`: https://zed.dev/docs/extensions/developing-extensions#directory-structure-of-a-zed-extension

musicq:
Hi @maxdeviant , thanks for the information. I've updated the extension file format, but not sure how to rerun the CI to do the check.

maxdeviant (approved):

PR #1314, opened 2024-08-27
Add phpactor based lsp for PHP

jelmervdl:
Adds a LSP for PHP using Phpactor.

The phar is downloaded to the extension's work directory if it isn't already there. It assumes you have PHP installed on your machine.

It's pretty barebones, but functional. I wanted to add an update check, but the only way I know how to get the version from phpactor is by running it with `--version` and parsing the output. But I couldn't find a way to do this from the extension.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1097,3 +1097,6 @@
 [submodule "extensions/zedwaita"]
 	path = extensions/zedwaita
 	url = https://github.com/someone13574/zed-adwaita-theme.git
+[submodule "extensions/php-phpactor"]
+	path = extensions/php-phpactor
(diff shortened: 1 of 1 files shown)


jelmervdl:
@cia-bot check

jelmervdl:
@cla-bot check

maxdeviant:
Thanks for the PR!

If you weren't already aware, the PHP extension already has support for Phpactor built-in: https://github.com/zed-industries/zed/blob/4e67d33d8803e674cafa36cf91a093fa8ca8163c/extensions/php/src/language_servers/phpactor.rs

jelmervdl:
Hahaha, I somehow completely missed it in the docs and failed to try beforehand apparently. My apologies!

Issue #521, opened 2023-07-04
crates.io helper


Badless:
### Check for existing issues

- [X] Completed

### Describe the feature

a feature that tells you "hey, this package is outdated, you can update it anytime in cargo.toml file!" just like neovim plugin [crates.nvim](https://github.com/Saecki/crates.nvim)

### If applicable, add mockups / screenshots to help present your vision of the feature

[Screencast from 2023-03-11 05-29-22.webm](https://user-images.githubusercontent.com/43008152/224464963-9810110f-2923-4346-a442-9d4f2723bdff.webm)

williamdes:
The version info should/could be always shown like https://marketplace.visualstudio.com/items?itemName=serayuzgur.crates does

Vollbrecht:
It seams there exist now https://github.com/appelgriebsch/zed-crates-lsp

williamdes:
Indeed !
@JustNixy maybe close this one ?

<img width="914" height="319" alt="Image" src="https://github.com/user-attachments/assets/a5133bc2-af7e-4a8b-9a65-b20598d9ebef" />

boozook:
@williamdes that LS has many problems and doesn't respect local dependencies with both fields "version" and "path" as well as [renamed dependencies](https://github.com/MathiasPius/crates-lsp/issues/2).

boozook:
@JustNixy I suppose the name of this issue should be like "dependencies in cargo package manifest helper" instead of support just one registry.

PR #1316, opened 2024-08-28
Add cargo-tom extension

frederik-uni:
this is a language server that suggests features, versions & crates in the cargo.toml file. it has an offline&online mode. when the offline mode is downloading/reading the data it defaults to the crates.io/v1 api & the requests are cached.


frederik-uni:
#521

maxdeviant (approved):
Thank you!

PR #1319, opened 2024-08-28
Add Sourcepawn language extension

tsuza:
[Sourcepawn](https://github.com/alliedmodders/sourcepawn) is a fork of a language called "Pawn", formerly known as "Small", which is a C-like language generally used for scripting. Sourcepawn was mainly made for [Sourcemod](https://www.sourcemod.net/), which is a modding framework for the Source engine.

Throughout the years, Sourcepawn has completely diverged from Pawn, both syntax-wise and back-end wise, becoming completely different from Pawn, so before anyone says "can't the support be added to Pawn then?", no, it can't.

The extension works fine. It correctly highlights code & the LSP functions correctly. Unfortunately Zed's themes are extremely barebones due to their small number of captures, so some specific highlights do not work. But this is not a big problem.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -894,6 +894,10 @@
 	path = extensions/solidity
 	url = https://github.com/zarifpour/zed-solidity.git
 
+[submodule "extensions/sourcepawn"]
+	path = extensions/sourcepawn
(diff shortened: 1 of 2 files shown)


maxdeviant (approved):
Thanks!

PR #1342, opened 2024-09-02
Add OMNeT++ extension

adamgeorge309:
NED and MSG (extensions .ned and .msg) are DSLs for the OMNeT++ network simulator. They are used by model frameworks for OMNeT++, such as the INET framework, Simu5G, Veins, etc. We are the team that develops [OMNeT++](https://github.com/omnetpp/omnetpp) and the [INET Framework](https://github.com/inet-framework/) and we have created tree-sitter grammars for these languages, as well as an extension for the Zed editor.

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -850,6 +850,11 @@ version = "0.1.0"
 submodule = "extensions/oh-lucy"
 version = "0.0.1"
 
+[omnetpp]
+submodule = "extensions/omnetpp-zed"
+path = "extensions/omnetpp"
+version = "0.0.1"
+
 [one-dark-darkened]
 submodule = "extensions/one-dark-darkened"
 version = "0.0.1"
--- a/extensions.toml
+++ b/extensions.toml
@@ -852,7 +852,6 @@ version = "0.0.1"
 
 [omnetpp]
 submodule = "extensions/omnetpp-zed"
-path = "extensions/omnetpp"
 version = "0.0.1"
 
 [one-dark-darkened]
--- a/extensions.toml
+++ b/extensions.toml
@@ -851,7 +851,7 @@ submodule = "extensions/oh-lucy"
 version = "0.0.1"
 
 [omnetpp]
-submodule = "extensions/omnetpp-zed"
+submodule = "extensions/omnetpp"
 version = "0.0.1"
 
 [one-dark-darkened]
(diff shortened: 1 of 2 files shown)


adamgeorge309:
@cla-bot check

maxdeviant (changes requested):
Could you please add the submodule to this PR?

The `.gitmodules` entry is present, but the submodule itself is not.

extensions.toml
```diff
@@ -794,6 +794,11 @@ version = "0.1.0"
 submodule = "extensions/odin"
 version = "0.1.0"
 
+[omnetpp]
+submodule = "extensions/omnetpp-zed"
+path = "extensions/omnetpp"
```
maxdeviant, line 799: We don't need this `path` here, as the extension repo has the extension at the root.

```suggestion
```

adamgeorge309:
I added the submodule and accepted your change. Thanks!

extensions/omnetpp-zed
maxdeviant, line 1: Can we change this directory to be just `extensions/omnetpp`, so it matches the extension ID?

maxdeviant:
It looks like there may be some conflicts with `main` that need to be resolved.

extensions/omnetpp-zed
adamgeorge309: Yes, we meant to do that, so that we don't have zed in the name.

adamgeorge309: Do I understand correctly that `path = extensions/omnetpp` in .gitmodules specifies the directory name?

maxdeviant: `path` is for a path within the extension repository. It's generally not needed unless you have multiple extensions in a repository, or need them located someone other than the root.

But what we want is for the directory in this repository to be `extensions/omnetpp`. The actual repository name can be `omnetpp-zed`.

What you want to do is add the submodule like this:

```sh
git submodule add https://github.com/omnetpp/omnetpp-zed.git extensions/omnetpp
```

adamgeorge309: I did this, thank you!

adamgeorge309:
I fixed the extension ID in extension.toml to match extensions.toml

maxdeviant:
> @maxdeviant, I fixed the extension ID in extension.toml to match extensions.toml

@adamgeorge309 I see the correct ID in your extension repository, but the submodule commit in this repo needs to be updated to match. I don't have the ability to push any changes to this branch, as your fork is on a GitHub organization.

adamgeorge309:
Thank you

PR #1355, opened 2024-09-03
Update Fountain to v0.0.5

LaPingvino:
Changing how sections and dialogue works so the breadcrumbs are nicer. Now it would be nice if I can add fold support too...


LaPingvino:
Sorry for keeping you busy @maxdeviant :laughing:

maxdeviant:
> Sorry for keeping you busy @maxdeviant 😆

It's no problem 😄

maxdeviant (approved):

PR #1372, opened 2024-09-05
Update Fountain to v0.0.7

LaPingvino:
Tree-sitter grammar had some false matches that I tried to improve before but didn't manage. Now I managed by adding a match for false character names with higher priority and making sure scene descriptions cannot be empty.


LaPingvino:
祝你有美好的一天 @maxdeviant

maxdeviant (approved):

PR #1385, opened 2024-09-09
Add cfml extension

ghedwards:


ghedwards:
@cla-bot check

maxdeviant (changes requested):
Could you please add the Git submodule to this PR?

The entry in `.gitmodules` is present, but the actual submodule reference is missing.

ghedwards:
> Could you please add the Git submodule to this PR?
> 
> The entry in `.gitmodules` is present, but the actual submodule reference is missing.

@maxdeviant done.

maxdeviant (approved):
Thank you!

PR #1404, opened 2024-09-12
Add harper extension

Stef16Robbe:


Stef16Robbe:
@cla-bot check

Stef16Robbe:
Totally missed that CI failed on the extension name, should be fixed now ^^

extensions.toml
```diff
@@ -510,6 +510,10 @@ version = "1.0.0"
 submodule = "extensions/halcyon"
 version = "0.0.5"
 
+[harper-ls]
```
maxdeviant, line 513: Could we change the extension ID to just `harper`?

maxdeviant: > Does that mean I should init and update all submodules on my fork?

If you see an error from other extensions on your branch you typically just need to merge in `main` (no need to do anything with the other submodules).

PR #1412, opened 2024-09-14
add Github themes v0.0.1

claytercek:
this adds 9 new themes, all generated from the official github design system, [primer](https://primer.style/).

- Github Dark Default
- Github Dark Colorblind
- Github Dark Dimmed
- Github Dark High Contrast
- Github Dark Tritanopia
- Github Light Default
- Github Light Colorblind
- Github Light High Contrast
- Github Light Tritanopia

This is different from the existing `github theme` extension because it includes themes for _all_ primer color schemes (similar to the github VSCode themes), and generates them directly from the primer npm package.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -394,6 +394,10 @@
 	path = extensions/github-theme
 	url = https://github.com/PyaeSoneAungRgn/github-zed-theme.git
 
+[submodule "extensions/github-themes"]
+	path = extensions/github-themes
+	url = https://github.com/claytercek/github-zed-themes.git
+
 [submodule "extensions/glazier"]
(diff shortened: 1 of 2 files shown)


maxdeviant (changes requested):
Thanks for the PR!

> This is different from the existing github theme extension because it includes themes for all primer color schemes (similar to the github VSCode themes), and generates them directly from the primer npm package.

We don't want to have two extensions that are so similarly named in the registry.

It would be nice if we could upstream the changes to the other theme.

claytercek:
Understood! I'll make a PR on the existing theme.

PR #1416, opened 2024-09-16
update vesper to 0.1.0

bdsqqq:
adds `Vesper Orange (P̶e̶p̶p̶e̶r̶m̶i̶n̶t̶)` theme variation. Same theme, but green in syntax highlighting is replaced with subtle gray. Greens used for Success and Additions in UI is still used.


bdsqqq:
Closing, there's something wrong with my local fork. The diff from this PR includes thins it should not.

PR #1444, opened 2024-09-21
Add java-extension submodule

Kwesodo:


maxdeviant (changes requested):
We already have two Java-related extensions:

- Java: https://github.com/zed-extensions/java
- Eclipse JDTLS language server support: https://github.com/ABckh/zed-java-eclipse-jdtls

How does this differ from those other extensions?

PR #1469, opened 2024-09-30
Add kanagawa-theme extension

Kareeme246:


Kareeme246:
@cla-bot check

maxdeviant:
Thanks for the PR!

We already have a [Kanagawa theme](https://github.com/ethangilmore/zed-kanagawa) extension in the extension registry.

We don't want to have multiple extensions that provide the same thing, so I would encourage you to contribute any improvements to the existing extension.

PR #1473, opened 2024-09-30
Add vscode-classic-theme

CharlesSBL:
This time everything is alright, so it has to pass finally all the checks.


maxdeviant (approved):
Thanks!

PR #1475, opened 2024-10-01
Update typos extension to v0.0.2

BaptisteRoseau:
The version 0.0.2 of the typos extension adds support for the following languages:

    - AsciiDoc
    - Astro
    - Bash
    - Biome
    - C#
    - C++
    - Deno
    - Docker
    - Emmet
    - GLSL
    - Groovy
    - Luau
    - Makefile
    - PHP
    - Prisma
    - Proto
    - Python
    - R
    - Rego
    - ReStructuredText
    - Scala
    - Svelte
    - TailwindCSS
    - Terraform
    - Uiua
    - Vue
    - Yarn

If there is a way to have a wildcard to enable the extension on all non-binary files instead of having to list them all please tell me.

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -1231,7 +1231,7 @@ version = "0.1.1"
 
 [typos]
 submodule = "extensions/typos"
-version = "0.0.1"
+version = "0.0.2"
 
 [typst]
 submodule = "extensions/typst"


BaptisteRoseau:
It seems like the `the \`wasm32-wasip1\` target may not be installed` in the CI job, is this intentional ?

maxdeviant:
> If there is a way to have a wildcard to enable the extension on all non-binary files instead of having to list them all please tell me.

We don't have this today, but it is something I am interested in supporting after I ran into the same issue: https://github.com/maxdeviant/zed-code-stats/blob/fc204f13289c00a013c4c71bdb7fa9383592aed9/extension.toml#L11-L60

maxdeviant (approved):

PR #1479, opened 2024-10-01
Add protobuf-ls 0.1.0

prskr:
This PR adds protobuf-ls in its first version.
The extension is still WIP.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -874,6 +874,10 @@
 	path = extensions/powershell
 	url = https://github.com/wingyplus/zed-powershell.git
 
+[submodule "extensions/protobuf-ls"]
+	path = extensions/protobuf-ls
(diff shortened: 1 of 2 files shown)


prskr:
@cla-bot check

maxdeviant:
@prskr What do you think about including the language server support with Proto itself?

I just extracted it out into its own extension, if you'd like to open a PR there to add support for the language server: https://github.com/zed-industries/zed/tree/cddd7875a42cb017b2751e5b6639ba322f6629df/extensions/proto

prskr:
Sure thing! I actually already thought about it but I'm not really proficient with Rust and your codebase, so I figured, a simple extension would be an easier start but will give it a try in the upcoming days, thanks!

PR #1506, opened 2024-10-10
Update XML to v0.0.7

sweetppro:
remove hard coded path extensions in favor of first_line_pattern regex


notpeter:
```
first_line_pattern = '^<.*xml'
```

Is there a reason you're not just using `<\? *xml` - Are there other possibilities for valid XML headers besides which you are trying to match?
```
<?xml 
<? xml
<?        xml
```

sweetppro:
Xliff files for example can use something like:
`<xliff xmlns="urn:oasis:names:tc:xliff:document:1.2" version="1.2">`

notpeter:
FYI I added a little bit of official docs for this too.
So if your regex isn't working for folks they can just add their own extensions.
https://zed.dev/docs/languages/xml

sweetppro:
@notpeter when will the update be live in the gallery?

image

PR #1313, opened 2024-08-27
Update Scala to v0.0.4

filipwiech:
Update Scala extension to the newest available version. Includes:
* https://github.com/scalameta/metals-zed/pull/15
* https://github.com/scalameta/metals-zed/pull/18
* https://github.com/scalameta/metals-zed/pull/19
* https://github.com/scalameta/metals-zed/pull/20

PR #1507, opened 2024-10-10
Update Scala to v0.1.0

WeetHet:
- Added highlighting support for auto completions and symbol search
- Added runnable tags for the main three ways scala apps can be run
- Set `tab_size` to 2

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -1099,7 +1099,7 @@ version = "1.0.1"
 
 [scala]
 submodule = "extensions/scala"
-version = "0.0.4"
+version = "0.1.0"
 
 [scheme]
 submodule = "extensions/zed"


filipwiech:
Thank you for the improvements! :+1: I think we need to update the version in the ` extensions.toml` file in this repository as well, since the previous PR #1313 for an example. :slightly_smiling_face:

WeetHet:
> Thank you for the improvements! 👍 I think we need to update the version in the `extensions.toml` file in this repository as well, see the previous PR #1313 for an example. 🙂

Yeah, I've completely forgotten about that. Updating the version in three places is a recipe for disaster, you would absolutely forget to do it in one of them

maxdeviant (approved):

PR #1513, opened 2024-10-11
Add Mosel theme

pierrenel:
I've been using [Mosel](https://github.com/Domeee/mosel.nvim) in neovim for a long time - figured it would be a good idea to port it to Zed :)

preview


maxdeviant (approved):
Thank you!

PR #1520, opened 2024-10-13
Add SageMath language extension

rot256:
This adds basic language support for [SageMath](https://www.sagemath.org/).

SageMath is *very* similar to Python: the language adds some Sage-specific syntactic sugar, changes some operators (e.g. ^ is exponentiation). The language is "compiled" down to python before execution. Nonetheless these changes prevent e.g. Python language servers from working correctly, therefore the need for a separate extension for ".sage" files.

Currently the extension works by using the Python grammar, but changes the highlights.scm to add Sage-specific keywords / builtins / constants.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -998,6 +998,10 @@
 	path = extensions/s-dark-theme
 	url = https://github.com/sinamombeiny/S-DarkTheme.zed.git
 
+[submodule "extensions/sagemath"]
+	path = extensions/sagemath
+	url = https://github.com/rot256/zed-sagemath
+
 [submodule "extensions/scala"]
 	path = extensions/scala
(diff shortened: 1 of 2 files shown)


notpeter:
Just FYI your grammar includes some NeoVim specific tree-sitter things (e.g. `lua-match` highlights like these):
https://github.com/rot256/zed-sagemath/blob/b795f975ca5ea4a735acdbc2b80b7dd4d46cec63/languages/sage/highlights.scm#L10-L18

I believe most of them are ignored and will be harmless, but wanted to mention it.

For reference here are the [Zed Python queries](https://github.com/zed-industries/zed/tree/main/crates/languages/src/python) which should work better.

notpeter:
Thanks! 
Screenshot 2024-10-13 at 22 05 50

PR #1522, opened 2024-10-14
xml: Bunmp to v0.0.7 (fix)

notpeter:
Fix for https://github.com/zed-industries/extensions/pull/1506 not bumping the version number in `extensions.toml` too.

We should probably add a CI test for this.

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -1433,7 +1433,7 @@ version = "1.1.3"
 
 [xml]
 submodule = "extensions/xml"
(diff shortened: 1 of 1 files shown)


maxdeviant:
> We should probably add a CI test for this.

I'm not sure this is fixable in CI.

The `extensions.toml` is the source of truth for whether to check an extension for whether it needs to be published. We don't even pull down the submodules unless we detect that there is a potential version bump needed.

PR #1534, opened 2024-10-16
Add D2 extension

gabeidx:
This PR adds an extension that provides support for the D2 diagram scripting language. 

It introduces a basic syntax highligthing feature for `.d2` files, as well as `d2` block in markdown files.


notpeter:
Thanks! And congrats on your first extension!

PR #1536, opened 2024-10-16
Add Graphviz extension

gabeidx:
This PR adds an extension that provides support for Graphviz: <https://graphviz.org>

It introduces a basic syntax highligthing feature for `.dot` and `.gv` files.

https://github.com/gabeidx/zed-graphviz


notpeter:
Thanks!
Screenshot 2024-10-16 at 15 37 35

PR #1544, opened 2024-10-18
Update basedpyright to v0.1.0

m1guer:


osiewicz:
We should also bump extension version in extensions.toml. See https://github.com/zed-industries/extensions/pull/1116 for reference.

m1guer:
> We should also bump extension version in extensions.toml. See #1116 for reference.

https://github.com/m1guer/basedpyright-zed/compare/6d41a3c0214ab829d5f812da85f190f834b35638...208a9e9204d0e1071dfb093e809fd4581eee9bef#diff-21e42dd2276e594017780607a13f07b058d020447a73c11f032c77823a8ed493

m1guer:
> Yeah, that's the extension manifest, but we also need to edit this file in extensions repo: https://github.com/zed-industries/extensions/blob/main/extensions.toml#L97
> 
> Should be as straightforward as changing the version number there.

oh really i fogot about this one

osiewicz:
Thanks!

PR #1560, opened 2024-10-19
Update QML extension to 0.0.2

lkroll:
This update includes LSP support utilizing qmlls. Thanks to @flukejones!

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -1075,7 +1075,7 @@ version = "0.0.3"
 
 [qml]
 submodule = "extensions/qml"
-version = "0.0.1"
+version = "0.0.2"
 
 [quiet-light-theme]
 submodule = "extensions/quiet-light-theme"


notpeter:
A possible future improvement, instead of using the hardcoded path of `/usr/bin/`:
https://github.com/lkroll/zed-qml/blob/1ecf7059ca926dd25b604aa62fe99fe0d59c7ec2/src/qmljs.rs#L25

You can use `worktree.which` to check for the binary in the users PATH
https://github.com/zed-extensions/java/blob/main/src/lib.rs#L29-L36

Thanks!

lkroll:
I just installed the extension on another machine and I think there is something wrong when checking for qmlls, as the extension is now trying to run qmlls6 even though qmlls is there. Still new to this.

`failed to spawn command. path: "/usr/bin/qmlls6"`

Will probably debug a bit further and push an update including your "fix" as well.

notpeter:
Yes. As a workaround you should be able to symlink your qmlls to `/usr/bin/qmlls6` and get it working.

PR #1565, opened 2024-10-21
vue: Bump to v0.1.2

notpeter:
Vue moved to a new repository. This is a no-op version bump which switches to that submodule.

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -1452,7 +1452,7 @@ version = "0.0.2"
 [vue]
 submodule = "extensions/zed"
 path = "extensions/vue"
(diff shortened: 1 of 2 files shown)


extensions.toml
```diff
@@ -1452,7 +1452,7 @@ version = "0.0.2"
 [vue]
 submodule = "extensions/zed"
 path = "extensions/vue"
```
maxdeviant, line 1454: We need to remove this path and make it point to the `extensions/vue` submodule.

maxdeviant (approved):

PR #1623, opened 2024-11-01
Add Hare language extension

### LINKED ISSUE #1592: Hare
### Check for existing issues

- [X] Completed

### Language

Hare

### Tree Sitter parser link

https://github.com/tree-sitter-grammars/tree-sitter-hare

### Language server link

n/a

### Misc notes

there are a couple LSPs for hare but theyre both quite old, unmaintained, and buggy. i dont think theres much reason to bother with them

xdBronch:
closes #1592
a made a few small changes to the highlighting, some fixes and some just kinda personal preference
the ts seems to work somewhat strangely with `export` (hares `pub`) which makes it appear next to function declarations rather than as part of them so i couldnt figure out how to get it to show up nicely in the outline, just left them out :man_shrugging:

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -526,6 +526,10 @@
 	path = extensions/haml
 	url = https://github.com/davidcornu/zed-haml.git
 
+[submodule "extensions/hare"]
+	path = extensions/hare
+	url = https://github.com/xdBronch/hare-zed.git
+
 [submodule "extensions/harper"]
 	path = extensions/harper
 	url = https://github.com/Stef16Robbe/harper_zed.git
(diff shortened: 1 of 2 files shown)


notpeter:
Hi! 👋 
If your language does not include a language server and is just providing tree-sitter grammars you don't need `src/lib.rs` and `Cargo.toml`.  You can always add them back later, but as-is they are unnecessary and should be removed.

xdBronch:
ah thanks i didnt realize that, makes sense

notpeter:
Thanks! Looks good!
image

PR #1631, opened 2024-11-04
Add UUID extension

mlavrinenko:
It's a common task to generate some UUIDs while you write new fixtures/migrations. This extension adds the `/uuid` slash command that generates a random UUIDv4.

This would be much more helpful if slash command usage were allowed in the editor (https://github.com/zed-industries/zed/issues/20095).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1286,6 +1286,10 @@
 	path = extensions/unoflat
 	url = https://github.com/bryanbuchanan/unoflat
 
+[submodule "extensions/uuid"]
+	path = extensions/uuid
+	url = https://github.com/mlavrinenko/zed-uuid.git
+
 [submodule "extensions/v"]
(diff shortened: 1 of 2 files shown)


notpeter:
In the future, if you commit to a branch (and not main) on your fork instead and check the box "allow maintainers to contribute to your branch" we can attempt to help you fix up your PRs.

mlavrinenko:
Thanks for the tip, didn't know.

And if it's preferred I can close the PR and make another one with uuid branch and option turned on.

maxdeviant:
This isn't really the intended use for slash commands. Their purpose is for providing context to the model.

As you noted, they only work in the Assistant Panel.

I implemented this functionality natively into the editor, for now: https://github.com/zed-industries/zed/pull/21656

PR #1657, opened 2024-11-11
Add Sorbet extension

notchairmk:
Adds Sorbet language server as a Ruby extension for Zed.

Source is here: https://github.com/notchairmk/zed-sorbet

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1190,6 +1190,10 @@
 	path = extensions/solidity
 	url = https://github.com/zarifpour/zed-solidity.git
 
+[submodule "extensions/sorbet"]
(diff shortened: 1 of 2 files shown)


maxdeviant:
@notchairmk What do you think about making this language server available in the [Ruby extension](https://github.com/zed-extensions/ruby)?

notchairmk:
@maxdeviant my first pass was adding it (https://github.com/zed-extensions/ruby/pull/9) to the Ruby extension, but the suggestion there was breaking it out into a separate extension. I'm happy either way though

maxdeviant (approved):
Thank you!

PR #1662, opened 2024-11-12
Add gren language extension

johanalkstal:
Adds basic syntax highlighting for the Gren programming language.


johanalkstal:
@cla-bot check

johanalkstal:
```
Error: IDs in `extensions.toml` and `extension.toml` do not match:

extensions.toml: gren
 extension.toml: gren-lang
```
Fixed this in the extension repository. Should work if the workflow is re-run.

johanalkstal:
Yeah I remembered I had to update the submodule and pushed that at the same time as you described how to do it 😅 
I forced pushed the change though cause it didn't seem necessary to make that change in a separate commit 🙂

notpeter:
Thanks! And congrats on your first extension!

PR #1672, opened 2024-11-14
Added Maya Theme 0.4.3 to extensions

sbowman:
I've created a theme based on a dark color theme I've been slowly creating across many other editors and IDEs over the years, and would like to include it with the available Zed extensions.


maxdeviant (approved):
Thanks!

PR #1676, opened 2024-11-15
Bump One Black Theme to v0.0.2

serhiiboreiko:
- The color of Git-ignored files is now more distinct.
- Fixed the placeholder text color to match the theme.
- Cyan in the terminal now matches the cyan in the editor.
- `border.variant` now matches other borders.
- Enhanced the visibility of the disabled icon color.


notpeter:
Thanks!

PR #1735, opened 2024-12-01
Add ziggy

lvignoli:
[Ziggy](https://ziggy-lang.io) support for Zed.

The LSP cannot be installed because of an upstream issue: releases for ziggy are distributed as `tar.xz`, not `tar.gz`.
As I understand, Zed extension SDK does not support this compression format.

Linked issues:
- https://github.com/zed-industries/zed/issues/21407
- https://github.com/kristoff-it/ziggy/issues/48

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1537,3 +1537,6 @@
 [submodule "extensions/zen-abyssal"]
 	path = extensions/zen-abyssal
 	url = https://github.com/KevInCompile/ZenAbyssal.git
+[submodule "extensions/ziggy"]
+	path = extensions/ziggy
+	url = https://github.com/lvignoli/zed-ziggy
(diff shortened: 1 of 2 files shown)


notpeter:
If we know that the download/extraction will fail, can we comment out that code and just require users to manually download and install ziggy into their PATH and output an error if `which` doesn't find the ziggy binary?

notpeter (changes requested):
If we know that the download/extraction will fail, can we:
1. comment out that code
2. require users to manually download and install `ziggy` into their PATH 
3. output an error when `which` fails to find the ziggy binary?

lvignoli:
@notpeter indeed, that's much sounder. Is this the proper way to do so?

notpeter (approved):

notpeter:
Thanks!

PR #1792, opened 2024-12-12
Bump Yara to v0.0.2

egibs:
Quick PR to bump the Yara extension to `0.0.2` which has better highlighting. The next update won't be such a fast-follow. 😅


notpeter:
Thanks!

PR #1803, opened 2024-12-15
Add phine theme 0.0.1

phisch:
It's my personal theme, but I think some people would enjoy it. A bit work in progress, but certainly good enough for daily use.

image


notpeter:
Thanks! And congrats on your first Zed theme extension.

PR #1807, opened 2024-12-15
Bump LaTeX to 0.1.4

lnay:
## Changes
- Improved highlighting for markers in commands like `\label` and `\ref`
- Better context aware self closing brackets


notpeter:
Thanks!

PR #1808, opened 2024-12-16
Update Gentle Dark Theme to v1.0.0

brianjgamble:
Releasing version 1.0 🥳 
- Update theme schema to version 0.2.0
- Adjust the colors on the tab bar/tabs so the active tab is more visible
- Add missing color values to reflect the theme color palette


notpeter:
Thanks!

PR #1810, opened 2024-12-16
Add zed-rhai extension

elkowar:
This PR adds a new language extension to support the [rhai](rhai.rs) scripting language.


notpeter:
Thanks! I renamed the submodule directory but otherwise looks good!

PR #1829, opened 2024-12-25
Add Path of Exile filter file extension

egibs:
This PR adds a new language extension for Path of Exile `.filter` files. 

Most of the time these files are auto-generated by https://www.filterblade.xyz/, but manual tweaking is useful and no such automation currently exists for Path of Exile 2.

Example:
image


notpeter:
Thanks!

PR #1830, opened 2024-12-25
Update Yara submodule to pick up URL fix

egibs:
This PR fixes a typo I made with the repository URL for the Yara extension. 

Previously, it was `https://github.com/egibs/yara-zed` when it should have been `https://github.com/egibs/yara.zed`.


notpeter:
Thanks

PR #1834, opened 2024-12-26
Update New Darcula to v1.0.7

e-simpson:
Fixed some highlight text on some imports/typedefs

https://github.com/e-simpson/new-darcula-z/issues/9


notpeter:
Thanks!

PR #1853, opened 2025-01-03
Update COBOL to v0.0.3

willswire:
There was a bug in `v0.0.2` where the downloaded LSP server binary was not marked as executable. This update fixes this issue.


notpeter:
Thanks!

Issue #1828, opened 2024-08-05
`racket-langserver` LSP support for Racket


bhargavkulk:
### Check for existing issues

- [X] Completed

### Describe the feature

Hi. Thanks for all the phenomenal work on this editor. The Racket programming language has an LSP called [`racket-langserver`](https://github.com/jeapostrophe/racket-langserver). It would be very nice and useful if the official Racket extension in Zed could support this LSP.

Reilithion:
I tried implementing this by copying the racket extension from the zed main repo, adding a Cargo.toml and a racket.rs that implements `language_server_command`, and adding it as a dev extension. It seems to add the extension just fine (I can see it in Extensions, with the new version number and my name in the authors), but there is no change in behavior in Zed when editing Racket source files. Checking with `ps -ef`, it seems like Zed isn't even attempting to run the language server. Is there something specific I need to do to get it to make use of the language server?

spdegabrielle:
> Is there something specific I need to do to get it to make use of the language server?

Only what is in the readme https://github.com/jeapostrophe/racket-langserver?tab=readme-ov-file#racket-langserver

bhargavkulk:
I had the same issue when I tried changing the extension to support the lsp server. Tangentially, the syntax highlighting for racket code is pretty poor. The treesitter syntax supports complex forms like match, match-define etc, but the the syntax definitions on Zed's side are very bare bones. Racket is not just brackets, tokens, and whitespace!

Reilithion:
> Only what is in the readme https://github.com/jeapostrophe/racket-langserver?tab=readme-ov-file#racket-langserver

I more meant on the Zed side of things, rather than the Racket language server side. I tried using the command to restart the language server, but it didn't seem to do anything. I'm wondering if the built-in extension for Racket support is taking precedence over my dev extension for handling .rkt files...

Reilithion:
Ah! After some fiddling, I got it to work. I needed this in my extension.toml file:

```
[language_servers.racket]
name = "Racket LSP"
languages = ["Racket"]
```

And the "Racket" in the square brackets had to be _capitalized._ I'm still not entirely sure where the name it has to match is defined. Maybe it's the Name field in languages/racket/config.toml? Anyway, as soon as I figure out a slightly less hard-coded way to locate the racket binary I'll throw this in a repo and make a pull request.

Reilithion:
I've submitted a pull request: #1857

probably-neb:
[...]

PR #1857, opened 2025-01-03
Add external Racket extension to add LSP support for Racket

Reilithion:
Addresses #1828. With this, it is now possible to use `racket-langserver` in Zed, providing the benefits of the Racket language server.


maxdeviant (changes requested):
Thank you for the PR!

I think we'd like to see the language server support added to the existing Racket extension: https://github.com/zed-extensions/racket

PR #1871, opened 2025-01-07
Add activitywatch extension

sachk:
This extension works pretty much the same internally as the wakatime extension that is already published in this repo


maxdeviant (approved):
Thanks!

PR #1904, opened 2025-01-13
Add ESLint LSP extension

GuyGooL5:
ESLint LSP support

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -430,6 +430,10 @@
 	path = extensions/env
 	url = https://github.com/zarifpour/zed-env
 
+[submodule "extensions/eslint"]
+	path = extensions/eslint
+	url = https://github.com/GuyGooL5/zed-eslint-extension.git
+
 [submodule "extensions/everforest"]
 	path = extensions/everforest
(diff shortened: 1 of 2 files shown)


GuyGooL5:
@cla-bot check

maxdeviant:
Thanks for the PR!

How is this different from Zed's existing ESLint support?

GuyGooL5:
Hi @maxdeviant, I wasn't even aware Zed already supports eslint as it didn't work for me, I'm running a project under an nx managed monorepo, is it possible that there's a bug there, or am I missing configuration?

maxdeviant:
> Hi @maxdeviant, I wasn't even aware Zed already supports eslint as it didn't work for me, I'm running a project under an nx managed monorepo, is it possible that there's a bug there, or am I missing configuration?

Yep, Zed comes with ESLint support built-in!

If you're having issues with it, would you mind opening an issue in the Zed repo? A sample repro with a reproduction of the issue would be great in helping us resolve it.

I'm going to close this PR out, as we already have ESLint built-in.

PR #1917, opened 2025-01-17
Add WebAssembly Text Format extension

g-plane:
This PR adds WebAssembly (wat) language support via:
- language server: [g-plane/wasm-language-tools](https://github.com/g-plane/wasm-language-tools) `wat_server`
- tree-sitter: https://github.com/wasm-lsp/tree-sitter-wasm

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1634,6 +1634,10 @@
 	path = extensions/warp-one-dark
 	url = https://github.com/distributed-intelligence/warp-one-dark
 
+[submodule "extensions/wasm"]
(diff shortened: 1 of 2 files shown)


notpeter:
Missing a `repository` in your extension.toml.

maxdeviant (changes requested):
Could we change the extension ID from `wasm` to `wat`, since it seems specific to the WebAssembly Text Format (and we already have a separate WIT extension)?

maxdeviant (approved):
Thank you!

PR #1937, opened 2025-01-22
Update Codebook to 0.1.1 to work with older Linux builds

blopker:
Fixed https://github.com/blopker/codebook/issues/1 by switching to using musl instead of libc for building Rust on Linux.

Thank you!


blopker:
Thanks!

PR #1961, opened 2025-01-28
Add Snowfall theme

freethinkel:
banner_dark


maxdeviant:
@freethinkel The Rust crate in your extension is causing the extension CLI to think that your extension has a Rust component.

You'll need to move the Rust code to a subdirectory to prevent it from being picked up.

freethinkel:
@maxdeviant I've fixed it, can you please check again?

maxdeviant (approved):
Thank you!

PR #1978, opened 2025-02-01
Add tree-sitter-query extension

vitallium:
Hi, this pull request adds a new extension for proper highlighting and editing of tree-sitter query files (*.scm). The repo is here - https://github.com/vitallium/zed-tree-sitter-query

Here are before&after screenshots:

CleanShot 2025-02-01 at 11 19 24@2x

CleanShot 2025-02-01 at 11 20 15@2x

The only thing is: Scheme language also uses the `scm` extension so there might be a conflict when you open a Scheme file or a Tree-sitter query file. Thanks!

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1538,6 +1538,10 @@
 	path = extensions/tomorrow-theme
 	url = https://github.com/biaqat/tomorrow-theme-zed.git
 
+[submodule "extensions/tree-sitter-query"]
+	path = extensions/tree-sitter-query
+	url = https://github.com/vitallium/zed-tree-sitter-query.git
+
 [submodule "extensions/tsar"]
 	path = extensions/tsar
 	url = https://github.com/x032205/tsar-zed-theme.git
--- a/extensions.toml
+++ b/extensions.toml
@@ -1630,6 +1630,10 @@ version = "0.1.2"
 submodule = "extensions/tomorrow-theme"
 version = "0.9.0"
 
+[tree-sitter-query]
+submodule = "extensions/tree-sitter-query"
+version = "0.0.1"
+
 [tsar]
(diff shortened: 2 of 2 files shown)


notpeter:
Thanks for this @vitallium.

Assuming this new tree-sitter works well and is stable, perhaps after testing for a week or two, I think we should update the suggested extension to point here instead of `scheme`.  People using Zed, editing a `.scm` file are 10x more likely to be editing a tree-sitter-query than a scheme file (in my experience).

https://github.com/zed-industries/zed/blob/0afc40abe13c6852dd0461285e0b4a50f7ed2a7a/crates/extensions_ui/src/extension_suggest.rs#L62

I've also gone ahead and moved the [scheme extension](https://github.com/zed-industries/zed/tree/main/extensions/scheme) out of the zed tree to [zed-extensions/scheme](https://github.com/zed-extensions/scheme/) as well.

vitallium:
> Thanks for this @vitallium.
> 
> Assuming this new tree-sitter works well and is stable, perhaps after testing for a week or two, I think we should update the suggested extension to point here instead of `scheme`. People using Zed, editing a `.scm` file are 10x more likely to be editing a tree-sitter-query than a scheme file (in my experience).
> 
> https://github.com/zed-industries/zed/blob/0afc40abe13c6852dd0461285e0b4a50f7ed2a7a/crates/extensions_ui/src/extension_suggest.rs#L62
> 
> I've also gone ahead and moved the [scheme extension](https://github.com/zed-industries/zed/tree/main/extensions/scheme) out of the zed tree to [zed-extensions/scheme](https://github.com/zed-extensions/scheme/) as well.

Yeah, that's a great idea! Thanks for the suggestion. Will add note about the `extension_suggest.rs` file.

PR #1987, opened 2025-02-03
Update LaTeX to v0.1.9

lnay:
Add preview and two-way sync for linux+GNOME.

I also add a `textobjects.scm` as per [the docs](https://zed.dev/docs/extensions/languages?highlight=text#text-objects)
defining captures for 'class' but vim-mode movements like `[[` and `]]` seem unaffected. I'm adding here to experiment and see if it's a dev-extension limitation.


notpeter:
Thanks!

PR #2002, opened 2025-02-06
Add VSCode great icons theme

RandaZraik:
Add VSCode Great Icons theme which is sourced from [VS Code Great Icons](https://github.com/EmmanuelBeziat/vscode-great-icons).


maxdeviant (approved):
Thank you!

PR #2013, opened 2025-02-08
Add JetBrains New UI Icons Theme extension

ankddev:
# Description
Add JetBrains New UI Icons Theme extension
# Screenshots
## Dark
image
## Light
image

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -776,6 +776,10 @@ version = "6.0.0"
 submodule = "extensions/java-eclipse-jdtls"
 version = "0.2.4"
 
+[jetbrains-newui-icons]
+submodule = "extensions/jetbrains-newui-icons"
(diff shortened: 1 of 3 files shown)


extensions.toml
```diff
@@ -776,6 +776,10 @@ version = "6.0.0"
 submodule = "extensions/java-eclipse-jdtls"
 version = "0.2.4"
 
+[jetbrains-newui-icons]
```
maxdeviant, line 779: Could we change the extension ID to `jetbrains-new-ui-icons`?

The change would need to be made here, in the submodule directory name, and in the `extension.toml`.

ankddev: Done

maxdeviant (approved):
Thanks!

PR #2031, opened 2025-02-11
Update docker-compose to v0.2.0

eth0net:
Release notes:

https://github.com/eth0net/zed-docker-compose/releases/tag/v0.2.0


eth0net:
So this extension works when loaded into Zed locally, but something in the CI package process isn't liking the dependency on yaml 🤔

PR #2034, opened 2025-02-11
Add iCal extension

TitouanReal:


maxdeviant:
How does this differ from the [vCard](https://github.com/TitouanReal/zed-vcard)` extension? I see they're both using the vCard Tree-sitter grammar.

PR #2078, opened 2025-02-19
Add colored-zed-icons-theme

TheRedXD:
This is an icon theme, which is based on the official zed monochrome icons, except this theme colors them!

The repository is here: https://github.com/TheRedXD/zed-icons-colored-theme


maxdeviant (approved):
Thank you!

PR #2093, opened 2025-02-21
Add material-icon-theme-v2

navneetlal:
Adds material-icon-theme-v2 will all set of icons


maxdeviant (changes requested):
Thank you!

We'd like to avoid having two Material Icon theme extensions, but we would welcome a PR to the existing extension to fill out the missing icons: https://github.com/zed-extensions/material-icon-theme

PR #2102, opened 2025-02-23
Update one-dark-flat to v0.1.1

0PandaDEV:
Release notes:

https://github.com/0PandaDEV/zed-one-dark-flat/releases/tag/v0.1.1

PR #2103, opened 2025-02-23
Update one-dark-flat to v0.1.1

0PandaDEV:
Release notes:

https://github.com/0PandaDEV/zed-one-dark-flat/releases/tag/v0.1.1


maxdeviant:
@0PandaDEV I updated the submodule in https://github.com/zed-industries/extensions/pull/2102 to include the version bump, so I think we're all set.

PR #2135, opened 2025-02-26
Update min-theme to v0.1.8

phibr0:
Release notes:

https://github.com/phibr0/zed-min-theme/releases/tag/v0.1.8

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -992,7 +992,7 @@ version = "0.3.0"
 
 [min-theme]
 submodule = "extensions/min-theme"
-version = "0.0.1"
+version = "0.1.8"
 
 [modest-dark]
 submodule = "extensions/modest-dark"


phibr0:
I don't understand why CI is failing :/

notpeter:
It's because [themes/base.json](https://github.com/phibr0/zed-min-theme/blob/main/themes/base.json) is not a theme.

notpeter:
It's because your themes aren't nested properly, the individual themes are in the top level of the json file; these objects should be nested under a top level themes array (like how `themes/min-theme.json` used to be). I assume you meant to combine them into `[themes/base.json](https://github.com/phibr0/zed-min-theme/compare/97f6689df56da4b17b1ab60731f0195814ff9a1b...ed2404516f271c7a94d9e87fa541340d78685150#diff-4b44fe4fe1226925aa191eb04985fa1d69bd93dbd417535b3f20a7fa90ae39e7) which has the right structure but an empty themes array.

phibr0:
All three files are combined into `min-theme.json` in the root. Can I somehow configure where the theme file is?

PR #2152, opened 2025-02-28
Add gitlab-ci-ls extension

tzabbi:
Hey there I want to publish my extension which is based on the gitlab-ci-ls project. See: https://github.com/alesbrelih/gitlab-ci-ls


maxdeviant (approved):
Thanks!

PR #2154, opened 2025-03-01
Add railscast theme

pbsds:
image


pbsds:
@cla-bot check

notpeter:
> Error: Extension IDs should not end with "-zed", as they are all Zed extensions: "railscast-zed".

Please update your extension id to just `railscast` in your extension.toml.
Then also fix it in `extensions.toml` here and pull forward the git submodule to the updated commit.

maxdeviant (approved):
Thanks!

PR #2174, opened 2025-03-03
Add HTML-Jinja extension

JaagupAverin:
I use this over the existing jinja extension since it provides a bit cleaner highlighting for HTML.

~Its a quick hack so I'm fine with not merging this and waiting for a nicer solution down the line. Just putting it out there if someone might find it useful for now.~

Update: Actually ended up forking the tree-sitter-html repository to flesh this out more and handle some edge cases better. Updated README. Works great already.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -714,6 +714,10 @@
 	path = extensions/horizon
 	url = https://github.com/ayn2op/zed-horizon.git
 
+[submodule "extensions/html-jinja"]
+	path = extensions/html-jinja
+	url = https://github.com/JaagupAverin/html-jinja.git
+
 [submodule "extensions/http"]
 	path = extensions/http
 	url = https://github.com/tie304/zed-http.git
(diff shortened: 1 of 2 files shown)


notpeter:
You need to add the submodule directory and commit that as well:
```sh
git add extensions/html-jinja
git commit -m "submodule too"
git push
```

JaagupAverin:
My bad didn't add the submodule correctly.

JaagupAverin:
Quick update: I'm actually gonna clean this up a bit, handle some more edge cases by forking the html tree-sitter. Will bump version soon enough.

JaagupAverin:
I rewrote the captures to allow jinja to be highlightable separately from the main language, since it effectively is a language inside a language. https://github.com/JaagupAverin/html-jinja?tab=readme-ov-file#captures
Not sure if there would've been a more standard way of doing this.

JaagupAverin:
I'm apparently horrible at submodules. Sorry I'll update the submodule in a moment.

notpeter:
Looks good! Congrats on your first Zed Extension!

PR #2224, opened 2025-03-08
Add one-dark-pro-max theme

kussumma:


maxdeviant (changes requested):
@kussumma Could you please update your theme to use the non-deprecated style property?

> Error: Theme "One Dark Pro Max" is using a deprecated style property: scrollbar_thumb.background. Use `scrollbar.thumb.background` instead.

kussumma:
Hai @maxdeviant I have update the deprecated style, thanks for your help

maxdeviant (approved):
Thank you!

PR #2246, opened 2025-03-12
Add Angular extension

sheikalthaf:


jpike88:
Test failing because the angular grammar lacks the `function_declaration` node type.

tree-sitter-angular should really be considered superset of TypeScript, so it's really an incorrect testing approach.

Two approaches I can think of:
- Zed combines the TypeScript and Angular grammars for now in its test for proper syntax coverage (faster, probably more 'correct' and flexible)
- The tree-sitter-angular library may need to explicitly include TypeScript grammar and be self-contained (slower, may not be desirable by the maintainers due to it effectively itself to TypeScript versions)

jpike88:
@maxdeviant this one too... note my comment that the test is a false negative, this extension works fine

jpike88:
@maxdeviant is there anything preventing you from letting this through? would like my devs to stop having to compile their extensions just for basic angular support

maxdeviant:
> @maxdeviant is there anything preventing you from letting this through? would like my devs to stop having to compile their extensions just for basic angular support

CI will need to pass before this can be merged.

jpike88:
The issue here is that the CI is trying to validate the angular extension's grammar as if it expects the entire TypeScript grammar to be baked into it. `function_declaration` is a TypeScript grammar.

The question is:

- should the angular extension have to explicitly bake in TypeScript grammar?
- or is there some kind of inheritance/dependency you need from us to make this work?

nathansbradshaw:
If I remember correctly, The Angular LSP is meant to run on top of the typescript LSP. It's an extension to the typescript LSP.  So the assumption that is made by the LSP is that the typescript LSP already exists. See the constraints section of the angular LSP design doc found here https://github.com/angular/vscode-ng-language-service/wiki/Design

maxdeviant:
> If I remember correctly, The Angular LSP is meant to run on top of the typescript LSP. It's an extension to the typescript LSP. So the assumption that is made by the LSP is that the typescript LSP already exists. See the constraints section of the angular LSP design doc found here https://github.com/angular/vscode-ng-language-service/wiki/Design

That is fine for the language server, but at the grammar level in Zed if you're using queries that correspond to a different grammar you need to provide that grammar in the extension so that the queries are valid.

nathansbradshaw:
So to get this into Zed's extensions we'd need to pull in the Typescript Grammer?

jpike88:
I guess best approach to just yank the latest typescript grammar files directly out of the zed repo each time you build, copy them in.

Aside from (unlikely) misalignment problems, will there be performance implications or side effects/bugs from doing that?

jpike88:
For sake of correctness, maybe best to copy the grammar directly from https://github.com/tree-sitter/tree-sitter-typescript, I noticed zed grammars more simplified for some reason... not sure why there is a divergence there

maxdeviant:
> So to get this into Zed's extensions we'd need to pull in the Typescript Grammer?

@nathansbradshaw After digging into this a bit more, it seems that the queries that are failing to compile are ones that Zed does not actually use.

The following queries should be removed from the Angular extension:

- `folds.scm` (this is the one producing the errors)
- `locals.scm` (which is already empty)

Separately, we were confused by the presence of `component.ts` as a path suffix for the Angular language: https://github.com/nathansbradshaw/zed-angular/blob/88f68544e42d18f5111826630325e06c9e835278/languages/angular/config.toml#L2

It seems like the Angular grammar is HTML-based and for Angular templates, so it doesn't seem valid for it to apply to a `.ts` file. I'm a little surprised this is even working at all, unless it isn't actually working and the `component.ts` files are being matched to the TypeScript language instead of Angular.

sheikalthaf:
@maxdeviant Let me check remove the files mentioned and adjust the path suffix.

sheikalthaf:
@maxdeviant I have update the PR as per your suggestions.

maxdeviant (approved):
Thank you!

PR #2348, opened 2025-03-27
Add ultralytics-snippets extension

ayberkgezer:


maxdeviant (changes requested):
Can we remove the "zed" from the extension ID and call it just `ultralytics-snippets`?

maxdeviant (approved):
Thank you!

PR #2352, opened 2025-03-28
+ shell-slash-command extension

subtleGradient:
This Zed extension adds a /sh slash command that allows you to run shell commands directly from Zed and see their output.


subtleGradient:
@cla-bot check

notpeter:
Apologies for the delayed response.  At this time I don't think this is something we're looking to ship as-is, especially with the introduction of the agent panel.  We might revisit this in the future, but for the moment I think it's a bit of a footgun.

Nifty idea though!

PR #2362, opened 2025-03-29
Add Svelte Snippets extension

bobbymannino:
this adds a few `.svelte` only snippets

I am looking into having `.ts` snippets for svelte in the same extension but not sure at the moment how to have multiple files in the snippets field in extension.toml 

if I use multiple it does not show any snippets


bobbymannino:
@cla-bot check

maxdeviant (approved):
Thank you!

PR #2366, opened 2025-03-30
Add Postgres Language Server

jeffreyguenther:
In this PR, I introduce an extension to add the [Postgres Language Server](https://pgtools.dev) built by the lovely folks at Supabase.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1422,6 +1422,10 @@
 	path = extensions/postgres-context-server
 	url = https://github.com/zed-extensions/postgres-context-server.git
 
+[submodule "extensions/postgres-language-server"]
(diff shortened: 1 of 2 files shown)


maxdeviant (changes requested):
Thank you!

@jeffreyguenther Would you mind changing the extension ID to `postgres-language-server`?

It seems that is the actual name they use, and "LSP" refers to the protocol, not to the language server itself.

jeffreyguenther:
Good call. 

Standby for an updated PR. I'll make the name change throughout the extension and for the repo.

jeffreyguenther:
@maxdeviant ready for a second look.

maxdeviant (approved):
Thank you!

PR #2441, opened 2025-04-10
Bump gruber-darker to v0.0.8

th0jensen:
I couldn't find a way to add the "theme" tag to the extension, but I would like that added in this revision of the extension :)


notpeter:
The tags are added as extensions are updated (we didn't backfill previously built extension versions) so after this merges it will have the `theme` tag.

th0jensen:
That's perfect, thanks! Praise be the great CI/CD you guys made

PR #2448, opened 2025-04-11
Update yara to v0.0.4

egibs:
This PR bumps the Yara extension to `0.0.4` which adds support for the Yara Language Server if installed.

I can install the new version locally as a dev extension without having YLS installed, so this shouldn't cause any issues.

More info: https://github.com/egibs/yara.zed/pull/1


notpeter:
Thanks!

PR #2471, opened 2025-04-16
Add wgsl-wesl

lucascompython:
Since this extension adds support for both WGSL and WESL, I'm not sure if I should create a directory called `wesl` in the languages folder. It works just fine on both WGSL as WESL as it is right now.

extension: https://github.com/lucascompython/wgsl-wesl-zed
tree-sitter: https://github.com/wgsl-tooling-wg/tree-sitter-wesl
lsp: https://github.com/wgsl-analyzer/wgsl-analyzer


notpeter:
This should work fine.  The language name is `WGSL/WESL` which should be ok. If you ever need to split them into distinct languages you can do so in a future update.

Thanks!

PR #2474, opened 2025-04-16
Add Fiber-snippets extension

ayberkgezer:


notpeter:
The repo of your extension is not available:
- https://github.com/ayberkgezer/fiber-zed-snippets

notpeter:
Your submodule had `Fiber-snippets` instead of `fiber-snippets` so I fixed that (everything should be lowercase). Now it's failing because the id in your extension.toml is uppercase.

```
Error: IDs in `extensions.toml` and `extension.toml` do not match:

extensions.toml: fiber-snippets
 extension.toml: Fiber-snippets
```

Commit a change which fixes that and then pull forward the submodule here and we should be good to go. (I pushed changes to your fork, so don't forget to `git pull`).

notpeter:
Thanks!

PR #2514, opened 2025-04-24
Add TeXpresso extension

lnay:
This is an extension to be used alongside the LaTeX one, providing a (very new) language server which only deals with live preview of a LaTeX document using [TeXpresso](https://github.com/let-def/texpresso), while `texlab` (from other extension) will provide all the other language server capabilities (go-to-definition, completions, snippets...). TeXpresso is a very experimental project but pulls off something not seen anywhere else in the LaTeX world with the preview updating smoothly with every keystroke. This particular project (LSP and zed integration) was prompted by a [user request in the LaTeX extension](https://github.com/rzukic/zed-latex/issues/85) and since then I have personally found it incredibly compelling.

*Demo showing live preview and inverse-search (clicking on parts of preview to move cursor in editor:*

https://github.com/user-attachments/assets/e3fd9773-c396-419b-8a12-c3925cdd28a3

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1922,6 +1922,10 @@
 	path = extensions/terrible-theme
 	url = https://github.com/nooooaaaaah/terrible-zed.git
 
+[submodule "extensions/texpresso"]
+	path = extensions/texpresso
+	url = https://github.com/lnay/zed-texpresso.git
+
 [submodule "extensions/the-best-theme"]
 	path = extensions/the-best-theme
 	url = https://github.com/Nidal-Bakir/zed-the-best-theme.git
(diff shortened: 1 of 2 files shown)


leanhdung1994:
TeXpresso currently could not handle the [problem of multi-pass](https://github.com/let-def/texpresso/issues/73#issuecomment-2311482608), so cross-references and citations will be displayed as `?`.  A simple [solution](https://github.com/let-def/texpresso/issues/73#issuecomment-2312285695) has been proposed. Could you please implement this feature? In this way, there is a button that has the following two effects:
1. The command `latexmk -xelatex"` is triggered such that the resulted intermediate files will be kept. The command can be configured/changed in the setting of the extension.
2. The internal state of TeXpresso is refreshed. The above intermediated files will be used by TeXpresso. The cross-references and citations will be resolved correctly.

Thank you very much for your consideration.

lnay:
I've made a note of this here: https://github.com/lnay/texpresso-lsp/issues/1

PR #2534, opened 2025-04-28
Add nightfox_m theme

mpopadic:
This is a custom Nightfox theme.
There is a new option called `midnight_m` I am personally using.


notpeter:
Thanks and congrats on your first Zed theme.

PR #2538, opened 2025-04-29
Replace nightfox_m theme with night-m 

mpopadic:
To comply with naming convention I am renaming the extension from `nightfox_m` to `night-m`.
The linked repo will remain the same. 
I have also updated the description and names in the extension itself.


notpeter:
Changing the names of your themes under `themes/*.json` is fine.
Changing the name and description in extension.toml is fine.
The only thing that needs to remain the same is `id` of the extension `nightfox-m`.

notpeter:
Thanks!

PR #2576, opened 2025-05-05
Update: tombi 0.1.1

yassun7010:
Update v0.1.1


notpeter:
This looks like a no-op change that just bumps the extension version forward, but makes no changes to the Zed extension itself.

Was this intentional?

PR #2586, opened 2025-05-06
Update Ruby to v0.7.0

vitallium:
Hi! This is an exciting update to the Ruby extension that introduces the new Process API. This version aims to improve the initial DX when starting the LSP, as well as enhance the overall DX of the Ruby extension. You can find the short changelog here: https://github.com/zed-extensions/ruby/releases/tag/v0.7.0 

Thanks!


notpeter:
Thanks!

PR #2641, opened 2025-05-13
Add Vue snippets extension

rubjo:

PR #2646, opened 2025-05-13
Add Vue snippets (with correct extension ID...)

rubjo:
Previously:
- https://github.com/zed-industries/extensions/pull/2645
- https://github.com/zed-industries/extensions/pull/2641


notpeter:
Thanks!

PR #2658, opened 2025-05-14
Add mcp-server-grafana

sd2k:
This commit adds an extension for the Grafana MCP server. It's useful
when adding observability to existing code, as the MCP server can
do things like search for existing Prometheus metrics or create a
dashboard using the metrics it finds in the code.


sd2k:
@notpeter Yep, done :+1:

PR #2659, opened 2025-05-14
Add Monokai Nebula theme

JacobCallahan:
This commit introduces the Monokai Nebula theme, which provides both dark and light variants.

Dark
rust_hover

Light
light_theme


notpeter:
> Error: Theme "Monokai Nebula" is using a deprecated style property: scrollbar_thumb.background. Use `scrollbar.thumb.background` instead.

clabot is having a bad day, I manually verified @JacobCallahan has signed.

notpeter:
Yeah, we don't have a straightforward story for running them locally and I'd like to improve the schema validation so theme authors get notified not just for deprecated theme items, but warned when they are missing new/updated theme elements. But not today.

Today, you can checkout the zed repo and then run:
```sh
cargo run -p extension_cli -- --source-dir ~/zsource/monokai-nebula-zed --output-dir /tmp/foo --scratch-dir /tmp/foo2
```
Which when I used https://github.com/JacobCallahan/monokai-nebula-zed/tree/b938a70052a4c2a191aeee503be71ff0c56e8faf  it generated the same:
```
Error: Theme "Monokai Nebula" is using a deprecated style property: scrollbar_thumb.background. Use `scrollbar.thumb.background` instead.
```
We build `extension_cli` for use in CI here, but we don't otherwise distribute it. (we could to aid in authors running tests in their own repos).

JacobCallahan:
@notpeter thanks for the pointers. In the meantime, I have a working version of the extension cli run in my actions from this point on: 
https://github.com/JacobCallahan/monokai-nebula-zed/blob/master/.github/workflows/pull_request.yml#L23

PR #2722, opened 2025-05-22
Update mcp-server-exa-search to v0.0.3

Feel-ix-343:


Feel-ix-343:
this includes the web_search_exa tool which is currently missing

PR #2746, opened 2025-05-25
Add Anthracite theme

boycsuk:
PR for a theme I just made, hope you guys like it.


boycsuk:
@cla-bot check

boycsuk:
I see the error about `pnpm sort-extensions`. I tried that command but it gave me 2 errors. 

First one about not having installed @iarna/toml. Did so and it still gave me another error. Is pnpm sort-extensions a must to pass the test?

Issue #2104, opened 2025-02-23
Tracking Issue: Extensions using `extension.json`


maxdeviant:
This is a tracking issue for extensions still using the `extension.json` manifest format instead of `extension.toml`.

## Extensions using `extension.json`

- [ ] `assembly`
  - Repository URL: [https://github.com/DevBlocky/zed-asm](https://github.com/DevBlocky/zed-asm)
  - Issue:
    - [Create Issue](https://github.com/DevBlocky/zed-asm/issues/new?title=Migrate+to+%60extension.toml%60&body=This+extension+has+been+identified+as+still+using+the+legacy+%60extension.json%60+manifest+format.%0A%0AExtensions+should+use+the+new+%60extension.toml%60+manifest+format.+See+the+%5BZed+extension+documentation%5D%28https%3A%2F%2Fzed.dev%2Fdocs%2Fextensions%2Fdeveloping-extensions%23directory-structure-of-a-zed-extension%29+for+more+information.)
- [ ] `asteroid`
  - Repository URL: [https://github.com/webhooked/asteroid-zed](https://github.com/webhooked/asteroid-zed)
  - Issue:
    - [Create Issue](https://github.com/webhooked/asteroid-zed/issues/new?title=Migrate+to+%60extension.toml%60&body=This+extension+has+been+identified+as+still+using+the+legacy+%60extension.json%60+manifest+format.%0A%0AExtensions+should+use+the+new+%60extension.toml%60+manifest+format.+See+the+%5BZed+extension+documentation%5D%28https%3A%2F%2Fzed.dev%2Fdocs%2Fextensions%2Fdeveloping-extensions%23directory-structure-of-a-zed-extension%29+for+more+information.)
- [ ] `blackfox`
  - Repository URL: [https://github.com/matejcerny/BlackFoxZed](https://github.com/matejcerny/BlackFoxZed)
  - Issue:
    - [Create Issue](https://github.com/matejcerny/BlackFoxZed/issues/new?title=Migrate+to+%60extension.toml%60&body=This+extension+has+been+identified+as+still+using+the+legacy+%60extension.json%60+manifest+format.%0A%0AExtensions+should+use+the+new+%60extension.toml%60+manifest+format.+See+the+%5BZed+extension+documentation%5D%28https%3A%2F%2Fzed.dev%2Fdocs%2Fextensions%2Fdeveloping-extensions%23directory-structure-of-a-zed-extension%29+for+more+information.)
- [ ] `blanche`
[...]

PR #2748, opened 2025-05-27
Update assembly-syntax to v0.0.2

DevBlocky:
Uses the new `extension.toml` format for the extension

Fixes my extension for #2104 

Does the folder name need to be the same as the extension ID? If so, I can update that

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -97,7 +97,7 @@ version = "0.0.1"
 
 [assembly]
 submodule = "extensions/assembly"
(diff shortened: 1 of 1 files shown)


notpeter:
Please update your extension id to `assembly` instead of `assembly-syntax` and pull forward the submodule.

DevBlocky:
@notpeter All set

notpeter:
Thanks!

PR #2751, opened 2025-05-28
Add rune extension

zhuhaow:
Add support for rune language https://rune-rs.github.io

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1818,6 +1818,10 @@
 	path = extensions/ruby
 	url = https://github.com/zed-extensions/ruby.git
 
+[submodule "extensions/rune"]
+	path = extensions/rune
+	url = https://github.com/zhuhaow/zed-rune
+
 [submodule "extensions/rust-rover-dark-theme"]
 	path = extensions/rust-rover-dark-theme
 	url = https://github.com/blkmlk/RustRover-Zed-Theme.git
--- a/extensions.toml
+++ b/extensions.toml
@@ -1860,6 +1860,10 @@ submodule = "extensions/zed"
 path = "extensions/ruff"
 version = "0.1.0"
 
+[rune]
+submodule = "extensions/rune"
+version = "0.0.1"
+
 [rust-rover-dark-theme]
 submodule = "extensions/rust-rover-dark-theme"
 version = "1.0.3"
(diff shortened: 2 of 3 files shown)


zhuhaow:
I tried to remove the rust grammar from `extension.toml` but the `rn` files then won't be correctly highlighted.

If I put it back then it works again.

Is there anything I did wrong? Or maybe you missed to rebuild the extension? Or it behaves differently with local dev extension?

notpeter:
It seems to work for me without bundling the tree-sitter: 🤔 
Screenshot 2025-06-06 at 10 40 12

It looks like it doesn't work when built by the cli zed-extension used for publishing extensions, so that's not actually an option (yet). 

You should also remove the Cargo.toml / Cargo.lock / src/lib.rs since they are currently a no-op.  Zed language extensions without LSP support (tree-sitter only) do not require any rust code (what you have is a no-op).

I'm also don't really love re-using the rust tree-sitter for another language for which it was not designed. Even for the most trivial script the tree-sitter shows a bunch of errors (dev: show syntax tree view).  I noticed you started work on a dedicated tree-sitter here, how's that going:
- https://github.com/zhuhaow/tree-sitter-rune

zhuhaow:
image

[tree-sitter-rune](https://github.com/zhuhaow/tree-sitter-rune) should be ready, but there is no query yet.

d1y:
I think you need to add the rune syntax manually
```diff
diff --git a/extension.toml b/extension.toml
index d80879c..39e69d4 100644
--- a/extension.toml
+++ b/extension.toml
@@ -6,6 +6,6 @@ authors = ["Zhuhao Wang <zhuhaow@gmail.com>"]
 description = "Rune language support for Zed"
 repository = "https://github.com/zhuhaow/zed-rune"
 
-[grammars.rust]
-repository = "https://github.com/tree-sitter/tree-sitter-rust"
-rev = "18b0515fca567f5a10aee9978c6d2640e878671a"
+[grammars.rune]
+repository = "https://github.com/zhuhaow/tree-sitter-rune"
+commit = "91dd6fa73fb14a02c728601ae0a853238625b627"
\ No newline at end of file

```

notpeter:
I'm happy to review this in a subsequent PR. You'll need to remove the src/Cargo.* as I specified above.  You should be able to write some basic highlights for your tree-sitter-rune (`dev: open syntax tree view` in Zed might be of help here).

Thanks!

PR #2756, opened 2025-05-29
Update marble theme to v0.2.0

TeenCoder159:
I have updated my extensions theme (added a light version of my existing theme)


notpeter:
Thanks!
P.S. You probably want to add .DS_Store to your .gitignore

PR #2760, opened 2025-05-30
Add Steep language extension

wouterken:
A zed extension that enables the [Steep](https://github.com/soutaro/steep) language server for Ruby

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1978,6 +1978,10 @@
 	path = extensions/statamic-antlers
 	url = https://github.com/mynetx/zed-statamic-antlers.git
 
+[submodule "extensions/steep"]
+	path = extensions/steep
+	url = https://github.com/wouterken/zed-steep.git
+
 [submodule "extensions/stimulus"]
 	path = extensions/stimulus
 	url = https://github.com/vitallium/zed-stimulus
--- a/extensions.toml
+++ b/extensions.toml
@@ -2025,6 +2025,10 @@ version = "0.3.0"
 submodule = "extensions/statamic-antlers"
 version = "0.1.0"
 
+[steep]
+submodule = "extensions/steep"
+version = "0.1.0"
+
 [stimulus]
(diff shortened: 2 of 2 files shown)


wouterken:
@cla-bot check

notpeter:
Hey, while I'm happy to ship a dedicated `steep` extension I wanted to ask before it ships whether you would be willing to consider integrating it with the existing [Zed Ruby Extension](https://github.com/zed-extensions/ruby).  There is already support for four other Ruby LSPs (see [src/language_servers](https://github.com/zed-extensions/ruby/tree/main/src/language_servers)) and it might be easier Steep was handled in the same way rather than as a distinct extension.

I know nothing about steep and I'm happy to merge this as-is, but just thought I'd ask.

cc: @vitallium In case [soutaro/steep](https://github.com/soutaro/steep) wasn't already on your radar.

wouterken:
> whether you would be willing to consider integrating it with the existing [Zed Ruby Extension](https://github.com/zed-extensions/ruby)

Absolutely, that would be a much cleaner experience. If @vitallium is open to it, I'm more than happy to close this and submit an equivalent PR there instead.

vitallium:
> > whether you would be willing to consider integrating it with the existing [Zed Ruby Extension](https://github.com/zed-extensions/ruby)
> 
> Absolutely, that would be a much cleaner experience. If @vitallium is open to it, I'm more than happy to close this and submit an equivalent PR there instead.

@notpeter @wouterken Hi! I think that would be great! The Ruby extension uses Process API so detecting `steep`, whether from the bundle, path, or extension-provided version, should be relatively easy.I wonder now, is it makes sense to add Sorbet as well. /cc @notchairmk

wouterken:
Closed to pick up the conversation here instead https://github.com/zed-extensions/ruby/pull/102 in hope of getting Steep supported inside the existing Zed Ruby Extension.

PR #2762, opened 2025-05-30
Update monokai-nebula to 0.2.1

JacobCallahan:
Scrollbar thumbs had a new active_background element added, so let's do the same!


notpeter:
Thanks!

PR #2763, opened 2025-05-30
Add Termux

paveloom:
Adds initial support for the [`termux-language-server`](https://github.com/termux/termux-language-server). The code of the extension is [here](https://github.com/paveloom-r/zed-termux).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2090,6 +2090,10 @@
 	path = extensions/tera
 	url = https://github.com/uncenter/zed-tera
 
+[submodule "extensions/termux"]
+	path = extensions/termux
+	url = https://gitlab.com/paveloom-g/rust/zed-termux
+
 [submodule "extensions/terraform"]
 	path = extensions/terraform
 	url = https://github.com/zed-extensions/terraform.git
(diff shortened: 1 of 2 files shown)


paveloom:
Thanks for the hint! I did try using the built-in grammar when I was developing the extension, and it didn't work. But now it did.

notpeter:
Ugh. It appears to work properly via "Install Dev Extension" but not when built with the zed-extension cli tool.

paveloom:
Unless the `zed-extension` CLI tool can be updated to support the built-in grammars, should I bring back the `grammars` section to `extension.toml`? `zed-extension` succeeds in building the extension with those lines present, and, after moving the contents of the created archive to the `~/.local/share/zed/extensions/installed/termux` directory, the extension appears to be working properly.

notpeter:
Yeah, go ahead and ship the grammar. 

If we update `zed-extension` to support referencing built-in languages (rust, bash, c, go, json, etc) we can always change it in a future update to the extension.

Sorry for the distraction.

notpeter:
Thanks and congrats on your first Zed extension!

PR #2787, opened 2025-06-05
Update Angular to v0.0.3

jpike88:
better syntax highlighting and correct an previously invalid extension.toml


notpeter:
In the future, in addition to updating the subdmoule you also need to update the version number in `extensions.toml`. (I've done it for you).

Thanks!

PR #2818, opened 2025-06-09
Add azure-context-server

oWretch:


maxdeviant (dismissed):
Could we change the extension ID to `mcp-server-azure` to make it a bit more specific?

oWretch:
Used `azure-context-server` as the id instead, following the `postgres-context-server` example

notpeter:
Thanks and sorry this sat open for so long.

PR #2809, opened 2025-06-08
Add Spai Zero theme extension

BasaiCorp:
## Description
Adding Spai Zero theme - a sleek dark theme with cyan accents and modern styling.

## Features
- Dark background with high contrast
- Cyan accent colors for better visibility
- Carefully crafted syntax highlighting
- Modern and clean design

## Screenshots
1. Screenshot1: [screenshots/screenshot1.png](https://github.com/BasaiCorp/zed-spai-zero-theme/blob/59a189b3cc0263c529d661d6843a64cf8e693cd7/screenshots/screenshot1.png)
2. Screenshot2: [screenshots/screenshot2.png](https://github.com/BasaiCorp/zed-spai-zero-theme/blob/59a189b3cc0263c529d661d6843a64cf8e693cd7/screenshots/screenshot2.png)
3. Screenshot3: [screenshots/screenshot3.png](https://github.com/BasaiCorp/zed-spai-zero-theme/blob/59a189b3cc0263c529d661d6843a64cf8e693cd7/screenshots/screenshot3.png)
4. Screenshot4: [screenshots/screenshot4.png](https://github.com/BasaiCorp/zed-spai-zero-theme/blob/59a189b3cc0263c529d661d6843a64cf8e693cd7/screenshots/screenshot4.png)

## Testing
- [x] Extension installs locally without errors
- [x] Theme displays correctly in Zed
- [x] All syntax highlighting works as expected

Thank you for your time!

PR #2825, opened 2025-06-09
Add mcp-server-llm-rule

hunner:
## Overview
This pull request adds the **mcp-server-llm-rules** extension to the Zed extensions repository. This extension enables direct integration of the [llm-rules](https://www.npmjs.com/package/llm-rules) engine into the Zed editor via the Model Context Protocol (MCP), allowing Zed to dynamically read and enforce rules from the same location as Cursor.

## Changes
- **Extension Addition:**  
  - Added new extension under `extensions/mcp-server-llm-rules`.
  - Includes all necessary Rust and Node setup for running the MCP context server that supplies rules to Zed from the `llm-rules` npm package.

- **Configuration Update:**  
  - Added an entry to the top-level configuration to register `mcp-server-llm-rules` as an available extension to Zed, including metadata, description, and repository reference.

- **Dev Environment:**  
  - Provided a ready-to-use development environment via `flake.nix` and `.envrc`.

- **Documentation:**  
  - Added a `README.md` detailing usage, the rule discovery process, frontmatter schema for rules, and how the context server works to present dynamic tools to the LLM in Zed.

## How It Works
- Discovers rule files (typically in `.cursor/rules/*.mdc`) and exposes each as an MCP tool to Zed.
- Requires YAML frontmatter in rule files describing the rule, file patterns, and enforcement behavior.
- The LLM automatically applies matching rules to file edits, guided by frontmatter metadata, with no user activation needed.
- Rules and their metadata are presented as context tools to Zed clients, leveraging project and file context for precision.

## Testing
- Verified discovery and correct exposure of rule tools from sample `.mdc` files.
- Confirmed extension runs and integrates properly with Zed’s MCP server system.
- Documentation checked for clarity and workflow guidance.

## Additional Notes
- The extension currently targets version **0.1.0**.
- See `README.md` for details on rule formats and usage patterns.
- No manual user activation is required—the LLM applies rules automatically based on file context and metadata.

## Checklist
- [x] Extension files and server code added under `extensions/mcp-server-llm-rules`
- [x] Build/test with Zed’s context server interface
- [x] Dev environment configuration provided
- [x] Documentation included and up-to-date

Thank you for reviewing this PR! Please let me know if you have questions or feedback.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1254,6 +1254,10 @@
 	path = extensions/mcp-server-linear
 	url = https://github.com/LoamStudios/zed-mcp-server-linear.git
 
+[submodule "extensions/mcp-server-llm-rules"]
+	path = extensions/mcp-server-llm-rules
+	url = https://github.com/hunner/mcp-server-llm-rules.git
+
 [submodule "extensions/mcp-server-miaoduo"]
 	path = extensions/mcp-server-miaoduo
 	url = https://github.com/benmooo/zed-mcp-server-miaoduo.git
--- a/extensions.toml
+++ b/extensions.toml
@@ -1280,6 +1280,10 @@ version = "0.1.1"
 submodule = "extensions/mcp-server-linear"
 version = "0.0.1"
 
+[mcp-server-llm-rules]
+submodule = "extensions/mcp-server-llm-rules"
+version = "0.1.0"
+
 [mcp-server-miaoduo]
 submodule = "extensions/mcp-server-miaoduo"
 version = "0.0.1"


notpeter:
I'm not sure this is the approach we want to support in Zed.

It sounds like your MCP creates a tool, per rules file and then requires the user to explicitly enable/disable each of the rules files they wish to have apply -- given that these are dynamic (per project) it means that the tools will also be dynamic and project specific.

Can you show me a screenshot of how this look when used in Zed? 

Also, if you're providing the rules as a tool, doesn't the model have to explicitly choose to call the tool to get the results? Isn't that the opposite of how rules are normally applied?

Separately, this description looks like AI generated slop. Please make sure that future replies are written by a human -- do not just paste drivel.  Show respect for me as a human being by valuing my time and taking a moment to write something yourself.

Apologies if I'm misunderstanding or mistaken.

hunner:
> I'm not sure this is the approach we want to support in Zed.

I agree with you! I was hoping I would be able to take advantage of my coworkers' Cursor rules in Zed, but after a week of using this, `llm-rules` turned turned out to be pretty clunky.

First, it isn't able to read rules from multiple locations. I'd love to have personal rules in my homedir, company rules in another location, and per-repo rules for each project in my Zed workspace(s). But as it is the `llm-rules` is limited to just `${the root of where Zed was launched}/.cursor/rules/*.mdc` and can't read any other directories or even subdirectories under `.cursor/rules/`.

Second, Zed appears to shares mcp server states across all instances so the root of where the FIRST zed is launched is the tools/rules for all other zeds. So if I `zeditor .` in one project root, those are the only rules I get across all instances.

I would love a rules interface that is programmatic and can read rules beyond the single `${projectroot}/.rules` file or Zed's rules UI from sqlite, but this extension isn't it.

> It sounds like your MCP creates a tool, per rules file and then requires the user to explicitly enable/disable each of the rules files they wish to have apply [...] Also, if you're providing the rules as a tool, doesn't the model have to explicitly choose to call the tool to get the results? Isn't that the opposite of how rules are normally applied?

Users actually don't get a choice of any of the rules/tools to enable... they are just all available on the MCP and the LLM uses them based on their `description`, `glob`, and `alwaysApply` fields at its on discretion. This part hasn't been any worse than trying to get an LLM to use other tools. Just have to write good tool descriptions and give the LLM the right prompt. Unfortunately Cursor doesn't populate the frontmatter for any rules that my coworkers are producing so I have to maintain those myself.

I only started using rules a week ago, so I'm still figuring out how rules are supposed to work. I assume `glob` and `alwaysApply` take precedent outside of the LLM and conditionally add the rules to the context. So yeah, it sounds like this isn't how rules should be applied.

> Can you show me a screenshot of how this look when used in Zed?

Sure! Here's one project with a mix of rules and how they show up to Zed in the mcp server settings panel:

image

> this description looks like AI generated slop

Apologies! It was based on https://github.com/zed-industries/extensions/pull/2809, but yes it is and I shouldn't have used it.

---

I look forward to any improvements Zed has with respect to rules! I'll have a look around the extension API to see if there is anything more specific to the Zed rules that I can implement to parse rules from a given location, but it seems that an MCP rules server is not the solution.

notpeter:
Thanks for confirming my hunches about the utility of this in Zed.

One thing I thought I'd mention. Although this is very much model specific, we found that model performance drastically degraded when they were presented with too many tools.  As someone who hasn't done much beyond trivial MCP tools, seeing 51 tools from GitHub and another 20 from your MCP gives me pause.

Perhaps this Claude 4/Opus 4 are better, but Zed initially included more specialized tools, but found the model (Claude 3.7 mostly) got distracted and picked the wrong tool and we had better results with a more minimal set.  Officially Anthropic says you can have hundreds of tools, but Zed currently ships with 16 built-in tools because it performed better with fewer available.

hunner:
Interesting! I'll keep an eye out for LLM performance with many tools enabled. I've switched from claude 3.7 to gpt 4.1 for most of my code interactions (it's much more straightforward and less chatty) and it's seemed okay. I'll give claude 4 some more time and see how it fares.

> Officially Anthropic says you can have hundreds of tools

That seems like an awfully low number! The number of typical commands someone could use in a day could easily exceed that. Maybe the model training hasn't taken tools into account much because they are so recent, but as time goes on the context can be populated with lots of rules and tools and the models will fare just fine.

PR #2833, opened 2025-06-11
Update Unison to v0.0.6

zetashift:
Brings in the version that updates the treesitter grammar and slightly improves the highlighting queries


notpeter:
Thanks!

PR #2851, opened 2025-06-17
Update browser-tools-context-server to v0.0.2

mirageN1349:
Fixed new agent mode settings, added support agent MCP


notpeter:
Thanks!

PR #2897, opened 2025-06-22
Add Deputy extension

filiptibell:
Deputy (Dependency Utilities) is a language server for your dependencies.
It provides autocomplete, hover info, and diagnostics for dependencies in package managers such as Cargo, NPM

https://github.com/filiptibell/deputy


filiptibell:
Seems like CI failed because the repository had Git LFS enabled. I've disabled it and force pushed + updated submodule to that latest commit without LFS.

PR #2925, opened 2025-06-27
Update  zen-abyssal a v1.3.2

KevInCompile:


notpeter:
When doing these PRs, you also need to update extensions.toml too (I did it for you this time): https://github.com/zed-industries/extensions/blob/f74bdaedc8821964a92f151df32ab91dcaf3e264/extensions.toml#L2643

Thanks

PR #2953, opened 2025-07-02
Update QML to v0.0.4

lkroll:
This PR will update the QML extension to version 0.0.4 which allows to set custom CLI arguments for qmlls.


notpeter:
Thanks!

PR #2955, opened 2025-07-03
Update Julia to v0.1.7

piechologist:
- Fix language injection syntax in injections.scm
- Add support for struct field docstrings
- Update Julia builtin functions and types
- Add auto-close support for triple-quoted-strings and for triple-backticks
- Fix indents for `begin` block
- Add `@testset` and `include` to Zed's outline views
- README.md: add instructions to switch `julia` executable
- Implement comprehensive indentation configuration for Julia (requires Zed 0.194.0+)

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -1068,7 +1068,7 @@ version = "0.4.0"
 
 [julia]
 submodule = "extensions/julia"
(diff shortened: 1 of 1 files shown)


notpeter:
Thanks!

PR #2969, opened 2025-07-05
Update Stylelint to v1.0.0

florian-sanders:
The extension no longer depends on a fork of the VSCode official extension to retrieve and build the Language Server from source.
It now fetches the built Language Server from the open-vsx package instead.

Changelog: https://github.com/florian-sanders/zed-stylelint/releases/tag/1.0.0


notpeter:
Thanks!

PR #2982, opened 2025-07-09
Update Slint extension submodule to official repository

xunicatt:
Change `extensions/slint` submodule origin from [flukejones/zed-slint](https://gitlab.com/flukejones/zed-slint) to [slint-ui/slint](https://github.com/slint-ui/slint).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2076,7 +2076,7 @@
 
 [submodule "extensions/slint"]
 	path = extensions/slint
-	url = https://gitlab.com/flukejones/zed-slint.git
+	url = https://github.com/slint-ui/slint.git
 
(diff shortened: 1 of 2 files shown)


maxdeviant:
@xunicatt Would it be possible to bump the version number of the Slint extension so that we can publish a new version that has metadata pointing to the right spot?

xunicatt:
Sure! I’ll first create a PR in the `slint` repo to update the `extension` version. Once that’s merged, I’ll follow up with a version bump here in `extensions`.

xunicatt:
Done

maxdeviant:
> Done

Awesome, thanks again!

PR #2986, opened 2025-07-09
Update Fountain to v0.2.0

LaPingvino:
Diving deep into some issues I suspected. Hoping this works properly because NixOS makes testing hard. Also trying to add an LSP...


notpeter:
Thanks!

PR #3001, opened 2025-07-13
Add codely-theme extension

GOI17:
Codely Dark theme port from VS Code


notpeter:
The submodule commit you've added: https://github.com/GOI17/codely-theme-zed/tree/cf4145f6998f0f150d97de21216640ddaef2d701 is not a member of any branch (e.g. master) and doesn't include your changes.

I assume you mean to use https://github.com/GOI17/codely-theme-zed/tree/63886aca0a6f4854279d15d86f02d2c0ef195096 (current master HEAD) instead so I've updated the submodule here to match.

Diff: https://github.com/GOI17/codely-theme-zed/compare/cf4145f6998f0f150d97de21216640ddaef2d701...63886aca0a6f4854279d15d86f02d2c0ef195096

If this is not what you intended, please adjust accordingly.

Also: 
> Error: Theme "Codely Dark" is using a deprecated style property: scrollbar_thumb.background. Use `scrollbar.thumb.background` instead.

notpeter:
Thanks!

PR #3042, opened 2025-07-22
Update cobol to v0.0.4

willswire:
Update COBOL extension

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -382,7 +382,7 @@ version = "0.1.0"
 
 [cobol]
 submodule = "extensions/cobol"
-version = "0.0.3"
+version = "0.0.4"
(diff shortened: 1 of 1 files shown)


notpeter:
```
[2025-07-22T13:07:24Z INFO  zed_extension] loaded grammar cobol
Error: Query error at 16:2. Invalid node type END-PERFORM
```
It errors on all of:
```
; "END-PERFORM" @keyword
; "END-IF" @keyword
; "EVALUATE" @keyword
; "END-EVALUATE" @keyword
```

You can see your upstream tree-sitter doesn't actually support these correctly and is erroring.
Screenshot 2025-07-29 at 17 25 29

PR #3044, opened 2025-07-23
Add Shader-ls extension

cjhowe-us:
This adds the Shader Language Server extension.

LSP: https://github.com/antaalt/shader-sense/tree/main/shader-language-server


notpeter:
Thanks!

PR #3049, opened 2025-07-23
Add rust-snippets extension

bobbymannino:
At the moment there is only 2 snippets but tomorrow i will add some more


notpeter:
Thanks!

PR #3056, opened 2025-07-25
Update lean4

blackbird314:
Update extension zed_extension_api version to 0.6.0.


notpeter:
@blackbird314 In addition to the submodule change you also need to update the version in extensions.toml.

PR #3081, opened 2025-07-29
Update helm

qvalentin:
The helm extension has gotten some improvements which are not included here yet.


notpeter:
Upstream needs to bump their version number before we can release an update.
Then `extensions.toml` has to be updated too.

PR #3085, opened 2025-07-29
Add autumnal-marscape theme

lilyyy411:
Adds the Autumnal Marscape theme, an original red, pink, and orange theme.


notpeter:

Issue #916, opened 2024-06-10
Add support for Go templates


nicolasparada:
### Check for existing issues

- [X] Completed

### Language

go template

### Tree Sitter parser link

https://github.com/ngalaiko/tree-sitter-go-template

### Misc notes

Add support for Go templates. See [text/template](https://pkg.go.dev/text/template) and [html/template](https://pkg.go.dev/html/template).

Right now one can get sort of syntax highlight of HTML, but by doing so it messes and inlines the template interpolation parts.

I also tried using prettier with the [go-template plugin](https://github.com/NiklasPor/prettier-plugin-go-template) and it works to some extend, but its tricky to configure for a non-js project.

nicolasparada:
For the curious, this is how you can configure zed to use prettier with the go-template plugin:

```shell
npm i -g prettier prettier-plugin-go-template
```

```json
{
  "file_types": {
    "HTML": ["tmpl"]
  },
  "languages": {
    "HTML": {
      "format_on_save": {
        "external": {
          "command": "prettier",
          "arguments": [
            "--stdin-filepath",
            "{buffer_path}",
            "--plugin",
            "/opt/homebrew/lib/node_modules/prettier-plugin-go-template/lib/index.js"
          ]
        }
      }
    }
  }
}
```

You might want to change `/opt/homebrew/lib/node_modules/prettier-plugin-go-template/lib/index.js` to whatever path the plugin was installed.

MonstraG:
I just gonna mention `.gohtml` here as a search keyword

HajagosNorbert:
I'll post my workaround here, since this is the first result on the subject. Sorry for being so blunt and ignorant, but I don't understand the dance that zed does with prettier, but someone may be in my situation. [nicolas's answer](https://github.com/zed-industries/extensions/issues/916#issuecomment-2157293096) didn't work for me, the prettier command couldn't be found, even though it was in the PATH.

I wanted to get .html files to be formatted as .gohtml, since I was using GoHugo. What solved it for me was to put this in the config (could be in a project's `.zed/settings.json` file as well)
```json
{
  "languages": {
    "HTML": {
      "prettier": {
        "allowed": true,
        "parser": "go-template",
        "plugins": ["prettier-plugin-go-template"]
      }
    }
  }
}
```
[...]

PR #3112, opened 2025-08-03
Add Gotmpl extension

hjr265:
Extension: https://github.com/hjr265/zed-gotmpl
Tree-sitter grammar: https://github.com/ngalaiko/tree-sitter-go-template
Language: https://pkg.go.dev/text/template and https://pkg.go.dev/html/template

hello gohtml

Related to #916


hjr265:
@cla-bot check

PR #3119, opened 2025-08-05
Update Tombi to v0.1.7

ya7010:


notpeter:
Looking at [the diff](https://github.com/tombi-toml/tombi/commit/540aa52f6f561b6bd8f4e917fb18541530ab8e54) for https://github.com/tombi-toml/tombi/pull/776, I think hard coding the search tombi inside a `.venv` virtualenv or at `node_modules/.bin/tombi` before running `worktree.which("tombi")` seems sub-optimal.

In general the preferred order of preference is:
- lsp.binary via settings
- worktree.which (check path)
- zed managed version

For example: [extensions/ruff/src/ruff.rs](https://github.com/zed-industries/zed/blob/main/extensions/ruff/src/ruff.rs#L20-L39)

The biggest reason to use this is that worktree.which takes the user's environment in account, and that includes support for directory hook integrations like direnv, asdf, virtualenvwrapper, etc.

Anyways, it's not strictly required but I wanted to mention it.
Super excited to see tombi progress, thanks for all your hard work.

ya7010:
Thank you for reporting this!

I was not aware of this issue. I will take some time to investigate it myself.

This was a modification made to align with VSCode's logic and prevent confusion for users when switching editors.

PR #3142, opened 2025-08-06
Add Maple theme

subframe7536:
While I'm porting my [VSCode theme](https://github.com/subframe7536/vscode-theme-maple) to Zed, I found someone already have already port it (extension id: `maple-theme`), but lack some font styles in syntax and use the old color palette, also the project seems inactive. So I want to submit a new theme extension here.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1264,7 +1264,7 @@
 
 [submodule "extensions/maple-theme"]
 	path = extensions/maple-theme
-	url = https://github.com/subframe7536/zed-theme-maple
+	url = https://github.com/subframe7536/zed-theme-maple.git
 
 [submodule "extensions/marble"]
 	path = extensions/marble
--- a/.gitmodules
+++ b/.gitmodules
@@ -2394,6 +2394,10 @@
 	path = extensions/the-dark-side
 	url = https://github.com/Imgkl/the-dark-side.git
 
+[submodule "extensions/theme-maple"]
+	path = extensions/theme-maple
+	url = https://github.com/subframe7536/zed-theme-maple
+
 [submodule "extensions/theme-rainbow"]
 	path = extensions/theme-rainbow
 	url = https://github.com/athxx/zed-theme-rainbow.git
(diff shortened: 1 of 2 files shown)


subframe7536:
@cla-bot check

notpeter:
Could you open a PR to improve [bushuai/zed-maple](https://github.com/bushuai/zed-maple) rather than adding another version of the same theme?  We try to avoid duplicates and would prefer collaboration wherever possible.  Thanks.

subframe7536:
Thanks for the suggestion! I understand the preference to avoid duplicates, but I still believe creating a separate extension is the better path forward in this case.

The existing maple-theme extension appears inactive and doesn't reflect the updated color palette or font styling I’ve implemented. I’d prefer to maintain and iterate on a standalone version that aligns more closely with my original VSCode theme and vision. If possible, it would be ideal for the older extension to be deprecated or removed to reduce confusion and ensure users get the most up-to-date experience.

Happy to discuss further if needed!

maxdeviant:
> Thanks for the suggestion! I understand the preference to avoid duplicates, but I still believe creating a separate extension is the better path forward in this case.
> 
> The existing maple-theme extension appears inactive and doesn't reflect the updated color palette or font styling I’ve implemented. I’d prefer to maintain and iterate on a standalone version that aligns more closely with my original VSCode theme and vision. If possible, it would be ideal for the older extension to be deprecated or removed to reduce confusion and ensure users get the most up-to-date experience.
> 
> Happy to discuss further if needed!

Given that you are the original author of the Maple theme, we would be open to allowing you to take over the `maple-theme` extension.

subframe7536:
All tasks are completed. Thank you for the support and feedback throughout! 🙌 Looking forward to seeing this merged.

subframe7536:
Fix the submodule issue and everything looks good on my end. I’d really appreciate it if you could take a moment to review and consider merging when you have time.

maxdeviant (approved):
Thanks!

PR #3170, opened 2025-08-11
Add todotxt extension

pursvir:
This time I removed dots from extension IDs which caused its instant removal (#2983).


maxdeviant (changes requested):
We're missing the actual submodule reference in this PR:

https://zed.dev/docs/extensions/developing-extensions#publishing-your-extension

PR #3171, opened 2025-08-11
feat: Add search match highlighting

kurokirasama:
- Update theme version
- Introduce search match background colors
- Set current search match background color

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -22,10 +22,6 @@ version = "0.0.2"
 submodule = "extensions/adech"
 version = "2.0.1"
 
-[adwaita]
-submodule = "extensions/adwaita"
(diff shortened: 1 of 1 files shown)


kurokirasama:
i don't now why I'm getting the error with this amber extension, that i didn't eve touched. I checked anyway and the current version in the repo is the Expected version: 0.1.2...

maxdeviant (changes requested):
It looks like there are a number of unrelated changes to other extensions.

Please follow the instructions for publishing a new version of an extension: https://zed.dev/docs/extensions/developing-extensions#updating-an-extension

kurokirasama:
I did but I don't know why it says I made those changes... I'll try again.

PR #3210, opened 2025-08-19
Update mcp-server-byterover to v0.0.3

ngduyanhece:
updates the mcp-server-byterover extension to version 0.0.3 with schema fixes

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1364,7 +1364,7 @@
 
 [submodule "extensions/mcp-server-byterover"]
 	path = extensions/mcp-server-byterover
-	url = https://github.com/campfirein/byterover-zed-extension.git
+	url = https://github.com/ngduyanhece/byterover-zed-extension.git
 
 [submodule "extensions/mcp-server-container-use"]
 	path = extensions/mcp-server-container-use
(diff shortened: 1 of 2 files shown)


.gitmodules
```diff
@@ -1364,7 +1364,7 @@
 
 [submodule "extensions/mcp-server-byterover"]
 	path = extensions/mcp-server-byterover
-	url = https://github.com/campfirein/byterover-zed-extension.git
+	url = https://github.com/ngduyanhece/byterover-zed-extension.git
```
maxdeviant, line 1367: Did you mean to change this back to the `ngduyanhece` repo after it was changed to `campfirein` in #3204?

wzlng: > > 2. Updating the `schema_version` to 2.
> 
> This was the problem. `schema_version = 2` does not exist.
> 
> > Switching the submodule to campfirein, which is an organization, while Zed’s instructions recommend forking it into a personal repo.
> 
> These instructions are for when forking the [zed-industries/extensions](https://github.com/zed-industries/extensions) repo, so that Zed staff can push to PRs opened against it (which we can't do when the extensions fork is in a GitHub organization.
> 
> There's no issue with an extension repository itself being in a GitHub organization.

Got it. We'll update this PR.

PR #3241, opened 2025-08-24
poc: update extensions with renovate

qvalentin:
This is a POC for how renovate could be used to automate the update of the extension.
It works by using a custom datasource that fetches the extension.toml from each extension repo to check if it matches extensions.toml in this repo and then bumps extensions.toml.

Additional one could run a script that then updates the git submodule of the updated extension (using renovates postUpgradeTasks (this requires self-hosting renovate i guess) or an CI job).

The poc only works for github repos currently is very hacky but I would like to know if you would be interested in this. I think the automation potential is huge since updating the submodules manually is quite error prone.

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -1,123 +1,153 @@
 [0x96f]
 submodule = "extensions/0x96f"
 version = "1.3.5"
+repository = "filipjanevski/zed-theme"
 
 [actionscript]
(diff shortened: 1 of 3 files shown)


notpeter:
This isn't the direction we're looking to head at the moment.
Thank you though.

PR #3322, opened 2025-09-05
Add an extension to include zuban as a python language server.

CMLL:
Adds support for Zuban as a language server for Python. https://zubanls.com/


CMLL:
@cla-bot check

PR #3341, opened 2025-09-10
New Odoo extension

fda-odoo:
Add extension for Odoo projects

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1778,6 +1778,10 @@
 	path = extensions/odin
 	url = https://github.com/zed-extensions/odin
 
+[submodule "extensions/odoo"]
+	path = extensions/odoo
+	url = https://github.com/odoo/odoo-zed.git
+
 [submodule "extensions/oh-lucy"]
 	path = extensions/oh-lucy
 	url = https://github.com/abdurrehman0206/Oh-Lucy.git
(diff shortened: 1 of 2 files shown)


fda-odoo:
@cla-bot check

notpeter:
Thanks!

notpeter:
Note, I don't think [these lines](https://github.com/odoo/odoo-zed/blob/8d234942614acf73fb12340ee5bbf17c21fd004d/src/odoo.rs#L95-L99) will work as you expect, because the `target_env` will be evaluated at extension packaging time (compiling to wasm) which I believe will resolve as `""` for wasm.  [target_env docs](https://doc.rust-lang.org/reference/conditional-compilation.html#target_env)

fda-odoo:
> Note, I don't think [these lines](https://github.com/odoo/odoo-zed/blob/8d234942614acf73fb12340ee5bbf17c21fd004d/src/odoo.rs#L95-L99) will work as you expect, because the `target_env` will be evaluated at extension packaging time (compiling to wasm) which I believe will resolve as `""` for wasm. [target_env docs](https://doc.rust-lang.org/reference/conditional-compilation.html#target_env)

You're right, I'll update that and commit it in the next version!
Thank you very much

PR #3343, opened 2025-09-10
elixir: Update to v0.2.3

smitbarmase:
Includes: 
- https://github.com/zed-extensions/elixir/pull/36

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -675,7 +675,7 @@ version = "0.0.4"
 
 [elixir]
 submodule = "extensions/elixir"
-version = "0.2.2"
+version = "0.2.3"
 
 [elle]
 submodule = "extensions/elle"


Benjamin-van-Heerden:
This works, thank you

But now things break when I am inside of specific heex expressions (and I am pretty sure this wasn't the case before)
e.g.
Here it is working (with the :if syntax)
image


but when inside an explicit if block
image

The same thing happens for heex case and for statements and propogates downward, i.e. if there is any top level heex syntax then the lsp works in none of the children

I opened an issue for this as well https://github.com/zed-industries/zed/issues/38300

Upon further inspection it breaks **anywhere downward** from where the if/for/case statement is
So if I have an <%= if @thing %> somewhere, then the entire rest of the file from there on won't have lsp for tailwind.

I also checked. Things are working in VSCode as expected, so this should hopefully be something simple

Issue #2819, opened 2025-06-09
Ctags/cscope support for large monorepos


xvbcfj:
### Check for existing issues

- [x] Completed

### Language

Multiple

### Misc notes

Hello,
 Ctags/Cscope support would be very good for when lsp cannot be used (large repos).

lnay:
Potentially relevant project: https://github.com/netmute/ctags-lsp

PR #3361, opened 2025-09-14
Add universal ctags lsp support

Mazurel:
This extension adds `ctags-lsp` language server to the Zed editor. User can override any language and set it as a language server, to use it. It can be useful on big, especially legacy, projects :)

It was also requested in #2819

- Extension: https://github.com/Mazurel/zed-ctags
- Ctags LSP: https://github.com/netmute/ctags-lsp
- Ctags: https://github.com/universal-ctags/ctags

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -498,6 +498,10 @@
 	path = extensions/csv
 	url = https://github.com/huacnlee/zed-csv.git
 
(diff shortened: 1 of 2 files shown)


maxdeviant (approved):
Thanks!

PR #3380, opened 2025-09-17
chore: add zed-emmylua-ls

und3fined:
LuaLS is best but emmylua better...


und3fined:
Now i mask email as Primary, pls check again

PR #3382, opened 2025-09-17
remove basedpyright

m1guer:
Looks like the BasedPyright extension is no longer necessary, so I will remove it.


m1guer:
ur welcome max! love u guys work

PR #3393, opened 2025-09-20
Publish comment language extensions

thedadams:
This extension highlights TODOs, FIXMEs, and other comments.


maxdeviant:
Awesome! I know a lot of folks have been asking for this.

Would you mind changing the name of the extension from "Zed Comment" to just "Comment"?

https://github.com/thedadams/zed-comment/blob/7909c2e551e0f37f940fd7ba6bedebc1c590cd56/extension.toml#L2

thedadams:
Thanks @maxdeviant. I updated the extension as you suggested.

maxdeviant (approved):
Thank you!

htahboub:
@thedadams it's not clear exactly how to use this. Just installing the extension doesn't seem to be doing anything. The expected behavior that would make sense to me is installing would automatically highlight TODO, NOTE, etc in comments in all languages. Is some additional setup necessary, and if so, why?

armaandh:
It's also a bit unclear to me how to get this working. (In my case, with TS/JS files)

thedadams:
Sorry for the confusion! See this from the README: https://github.com/thedadams/zed-comment?tab=readme-ov-file#compatibility

It turns out that getting this to work with the built-in languages is a little more effortful than I anticipated. I have a (broken) PR for the built-in languages and a couple other PRs for other languages.

PR #3422, opened 2025-09-26
Add Vynora theme extension

QuarkPixel:
This PR adds the **Vynora** theme to the extensions repository.

Vynora is a theme based on the Monokai Pro color scheme, featuring more saturated tones with a retro touch.


maxdeviant (approved):
Thanks!

PR #3460, opened 2025-10-01
Add license validator

JosephTLyons:
This PR adds validation to check if an extension repository contains either an MIT or Apache 2 license. If one is not present, it throws an error.

The pattern checking for MIT and Apache 2.0 licensing is a bit simple, it just verifies the first line or sentence of each block from each license. I could've matched on the whole license, but that felt a bit extreme; we can adjust if we find this isn't ideal.

### DIFF
--- a/src/lib/fs.js
+++ b/src/lib/fs.js
@@ -36,3 +36,34 @@ export async function fileExists(path) {
     throw err;
   }
 }
+
+/**
+ * Read all files in an extension directory with their content
+ * @param {string} extensionPath
+ * @returns {Promise<Array<{name: string, content: string}>>}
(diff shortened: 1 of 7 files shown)


Be-ing:
I see the text of the GPLv3 was added in this PR but the code does not check for it. Was that intentional or an oversight?

Be-ing:
There are existing extensions that are GPL licensed, for example the [Slint extension](https://github.com/slint-ui/slint/blob/5ba8206edb6d982d8f473a7bfc1bd9fea2f43730/editors/zed/Cargo.toml#L9). Why would that not be allowed anymore?

PR #3508, opened 2025-10-04
Update ton to 0.3.1

Bionic2113:
Update extensions api for windows support
Update highlights and tree sitter for tolk lang


maxdeviant (changes requested):
@Bionic2113 You need to change the extension ID back to `ton`:

https://github.com/Bionic2113/ton-zed/blob/0b67bfb5b505f3c112379c4d191380553dad1cf5/extension.toml#L1

PR #3509, opened 2025-10-04
Add Latte template engine extension

josbeir:
This adds initial support for the [Latte](https://latte.nette.org/) template engine (PHP)

* Tree sitter: [tree-sitter-latte](https://github.com/josbeir/tree-sitter-latte)
* Extension: https://github.com/josbeir/zed-latte
* Syntax highlighting, brackets, outline, indents.
* Emmet LSP enabled


maxdeviant (approved):
Thank you!

PR #3523, opened 2025-10-08
Adding JWT extension

mjovanc:


mjovanc:
@cla-bot check

maxdeviant:
Thank you for the PR, but this isn't really the intended usage of slash commands, and we don't want to encourage extensions that use it like this.

Slash commands are intended to provide context to the agent when using text threads.

PR #3530, opened 2025-10-09
Update EdgeJS extension to v1.0.1

Hexacker:
Add EdgeJS language server that brings a lot of useful features.


Hexacker:
> @Hexacker Please update the submodule as well.

Fixed

PR #3532, opened 2025-10-09
verilog: Bump to v0.0.11

someone13574:
- Updates grammar to newer versions
  - Fixes highlights for updated grammar
- Fix highlighting of comments
- Added the `.vh` file association 
- Updates language servers to newer releases
- Updates extension api version to 0.6.0


someone13574:
I didn't see that. Should I update to `0.7.0` then?

MrSubidubi:
Would be very nice if you could do that!

PR #3533, opened 2025-10-10
Add roto

ConradIrwin:
Co-authored by: @tertsdiepraam

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2238,7 +2238,6 @@
 	path = extensions/rose-pine-theme
 	url = https://github.com/rose-pine/zed
 
-
 [submodule "extensions/rosevin"]
 	path = extensions/rosevin
 	url = https://github.com/anson-ryea/rosevin-zed.git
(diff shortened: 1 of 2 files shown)


tertsdiepraam:
It seemed to fail on a missing license (if I'm reading it correctly). I've added one to the repo so it should succeed if we rerun the CI.

tertsdiepraam:
That almost worked 😄 I think I was missing the "MIT License" header that your regex check contains. To be sure it will succeed now, I copied the license from another extension (with the name changed of course).

ConradIrwin:
@tertsdiepraam I think we're not supposed to create a submodule for the grammar, but let the extension builder check that out for us.

tertsdiepraam:
I'll remove it then!

tertsdiepraam:
Updated! (don't mind the git log, `jj` didn't seem to like git submodules 😄)

tertsdiepraam:
Thanks @ConradIrwin!

PR #3559, opened 2025-10-14
Add Flow LSP extension

jthomaschewski:
Adds support for the [Flow](https://flow.org/) language.

It adds static typechecking to JavaScript, similar to Typescript.  
Used mostly by Facebook, e.g. React is written in Flow.

Flow comes with LSP support built-in. Also no additional tree-sitter grammar is required, it seems to work fine with the already existing JavaScript grammar.

Features:
- Activates for JavaScript-files if `.flowconfig` file is present (else it returns a meaningful error)
- Uses local `flow` binary from `nodes_modules` if available

Extension: https://github.com/jthomaschewski/zed-flow

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -818,7 +818,6 @@ version = "2.0.1"
 
 [flow]
 submodule = "extensions/flow"
-path = "flow"
 version = "0.0.1"
 
 [flow-theme]


jthomaschewski:
@cla-bot check

jthomaschewski:
I fixed the submodule url (use https:// instead of ssh).
The pipeline should succeed now hopefully.

It would be great if you could approve the workflow again. Thanks!

extensions.toml
```diff
@@ -816,6 +816,11 @@ submodule = "extensions/flexoki"
 path = "zed"
 version = "2.0.1"
 
+[flow]
+submodule = "extensions/flow"
+path = "flow"
```
maxdeviant:
```suggestion
```

maxdeviant (approved):
Thanks!

PR #3563, opened 2025-10-15
Update norminette to v0.2.2

DavidLee18:
There is no direct update to the extension itself, but `inject_42h`, a CLI this extension uses, is updated, so I thought it would be reasonable to also update the extension itself.

The changelog for `inject_42h` is [here](https://github.com/DavidLee18/inject_42h/releases/tag/v0.3.0).


MrSubidubi (approved):
Thanks!

PR #3564, opened 2025-10-15
Codebook v0.2.1

blopker:
Add Zig and LaTeX support! 🚀


blopker:
Hm, I'm not sure what the error in CI is about. Did something in the CI process change to check for LICENSE files? The repo has had a license file since the beginning though: https://github.com/blopker/codebook-zed/blob/main/LICENSE

blopker:
K, looked at the CI source and fixed the license file to pass the test. GitHub did accurately identify the old license as MIT though. I wonder if using https://docs.github.com/en/rest/licenses/licenses?apiVersion=2022-11-28#get-the-license-for-a-repository would be a more robust test?

blopker:
Outdated: https://github.com/zed-industries/extensions/pull/3595

PR #3565, opened 2025-10-15
Add emmylua extension

CppCXY:
This PR introduces the zed-emmylua plugin, which uses the [emmylua-analyzer-rust](https://github.com/EmmyLuaLs/emmylua-analyzer-rust) as language server.

I think our own language server should be maintained by ourselves, rather than merged into zed-extensions/lua. Its maintenance is too limited and does not meet our needs.

If we have time in the future, we can continue to integrate DAP and introduce a debugger, so this is not only a language server project.


CppCXY:
@cla-bot check

notpeter:
You'll need to run `pnpm run sort-extensions` to sort the `.gitmodules` and `extensions.toml`.

CppCXY:
done!, I have sort the extensions and gitsubmodules

MrSubidubi:
Thanks for this!

Could you please briefly elaborate on
> I think our own language server should be maintained by ourselves, rather than merged into zed-extensions/lua. Its maintenance is too limited and does not meet our needs.
? We usually try to only have one extension per language, hence my question.

CppCXY:
> Thanks for this!
> 
> Could you please briefly elaborate on
> 
> > I think our own language server should be maintained by ourselves, rather than merged into zed-extensions/lua. Its maintenance is too limited and does not meet our needs.
> > ? We usually try to only have one extension per language, hence my question.

I hope to maintain my own plugin and clearly distinguish it from luals as two separate plugins. We use emmylua as the language server and may use emmylua debugger as the debugger in the future. The zed-lua plugin is different—they usually prefer to use luals as the language server and lua-debug as the debugger.

maxdeviant:
Looks like there is an error in one of the Tree-sitter queries:

```
[2025-10-26T04:09:54Z INFO  zed_extension] loaded grammar lua
Error: Query error at 2:2. Invalid node type snippet
```

CppCXY:
I have removed some unrelated files, and this issue should be fixed now.

PR #3575, opened 2025-10-16
Add crates autocomplete extension

countradooku:
Add crates autocomplete extensions to Zed based on crates-lsp


MrSubidubi:
Thanks for this!

We do already seem to have support for this language server via https://github.com/appelgriebsch/zed-crates-lsp, so I'll close this out for the moment. Feel free to reach out here should you disagree. Thanks!

PR #3584, opened 2025-10-17
verilog: Bump to v0.0.13

someone13574:
Adds `slang-server` as a third language server, but it should be disabled by default in https://github.com/zed-industries/zed/pull/40442 due to an issue with it not initializing correctly on Windows.

The binaries are built by CI in the extension repo due to there not being upstream builds (as is done for the veridian ls).

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -2857,7 +2857,7 @@ version = "0.0.9"
 
 [verilog]
 submodule = "extensions/verilog"
-version = "0.0.12"
+version = "0.0.13"
 
 [vesper]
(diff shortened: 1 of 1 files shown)


someone13574:
> Looks good, thank you for this.
> 
> In combination with the changes in the main repo, I would go ahead and merge this prior to the next release on wednesday and cherry pick the changes onto stable so we have minimal disruption for users of your extension (we should really change that so this isn't necessary, but yeah, different topic). Is that fine for your or do you have objections?

Sounds good to me! Thank you!

PR #3591, opened 2025-10-17
Add Maybe Material theme

iamawatermelo:
Adds the Maybe Material theme, containing six colours and a load of variants

amethyst-standard

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1442,6 +1442,10 @@
 	path = extensions/maya
 	url = https://github.com/sbowman/maya.zed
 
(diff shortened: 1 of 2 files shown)


MrSubidubi (dismissed):
Thanks for this!

One minor question, could we perhaps change the extension ID here to `maybe-material` or do you have something against that?

iamawatermelo:
No, not at all—I've gone ahead and updated the ID.

maxdeviant (approved):
Thank you!

PR #3593, opened 2025-10-17
add YAML extension

TheBearodactyl:


TheBearodactyl:
nvm looks like my problem isn't a lack of yaml support

PR #3600, opened 2025-10-18
Update Tombi to v0.2.0

ya7010:
Changed Zed's LSP ID from `toml` to `tombi`.

And fix `zed::Command` generation logic.

https://github.com/tombi-toml/tombi/releases/tag/zed-v0.2.0


ya7010:
Changing the `lsp.$id` in `settings.json` is a breaking change, but I overlooked this when creating the Zed extension.

If there's an appropriate place to make an announcement, I'll make the announcement there.

PR #3604, opened 2025-10-19
Add VSCode Icons Original extension

robertohuertasm:
This PR adds a new icon extension built by the VSCode Icons team.

Extension: https://github.com/vscode-icons/vcode-icons-zed

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2878,6 +2878,10 @@
 	path = extensions/vscode-great-icons
 	url = https://github.com/RandaZraik/zed-vscode-great-icons.git
 
+[submodule "extensions/vscode-icons"]
+	path = extensions/vscode-icons
+	url = https://github.com/vscode-icons/vcode-icons-zed.git
+
 [submodule "extensions/vscode-light-plus"]
 	path = extensions/vscode-light-plus
 	url = https://github.com/MadLittleMods/zed-theme-vscode-light-plus.git
@@ -3061,6 +3065,3 @@
 [submodule "extensions/zwirn"]
 	path = extensions/zwirn
 	url = https://codeberg.org/polymorphicengine/zwirn-zed-extension.git
-[submodule "extensions/vscode-icons"]
-	path = extensions/vscode-icons
-	url = https://github.com/vscode-icons/vcode-icons-zed.git
--- a/extensions.toml
+++ b/extensions.toml
@@ -2923,7 +2923,6 @@ version = "0.2.6"
 submodule = "extensions/vscode-icons"
 version = "12.15.0"
 
-
 [vscode-light-plus]
 submodule = "extensions/vscode-light-plus"
 version = "0.0.2"


MrSubidubi:
Thanks for this!

I do see we have https://github.com/chawyehsu/zed-vscode-icons-theme for this already, so would it perhaps be better to replace that extension and have it instead point at the official repository in your organization? Then, existing users would seamlessly be migrated to the official version and we would not have the duplicate. CC @chawyehsu

robertohuertasm:
@MrSubidubi, when we submitted this, there were at least 3 extensions using our icons. We thought about maybe pinging some of these devs, but in the end we thought it was easier to start from scratch and publish the extension as is. I guess we hoped for other extensions authors to decide what to do with their repos.



image

MrSubidubi:
@robertohuertasm Quickly checking in on thees, it seems that two of these are a port of your theme, the others mirror distinct themes from the ecosystem. 

If you do not mind, I'd quickly chat with some folks internally on this and then see how to proceed from here.

robertohuertasm:
@MrSubidubi sure, whatever is best for everyone.

chawyehsu:
@robertohuertasm I admire your work on vscode-icons and it's great to know you make official support for Zed.

@MrSubidubi Thanks for pinging me to let me know about this. I made https://github.com/chawyehsu/zed-vscode-icons-theme months ago because the discontinuance of previous port https://github.com/pranavmangal/zed-vscode-icons (the most amout of installs on the icons theme list). I created a transformer to make the vscode-manifest -> zed-manifest process eaiser so I can track and update the extension without taking much time. I also updated the extension to support named directory icons.

I understand that having multiple extensions with the same functionality in the list is not a good idea, especially now that it’s officially supported. Therefore, I’ll accept any outcome, including the removal of my extension or ownership transfer. However, I hope the transition can be seamless so that users can migrate to the official version with minimal effort.

robertohuertasm:
@chawyehsu thank you for the nice words and for all the effort you put to bring `vscode-icons` to Zed! ❤️ 🙏🏻 🙇🏻 

We are now generating the Zed extension manifest as part of the build process itself. So the idea is to leverage that to publish both VS Code and Zed extensions at the same time in an automated way. 

We also expanded a little bit the way we work with VS Code `languages`, which are not supported in Zed, and added known extensions and filenames to the languages so we can easily map the icons to the proper extensions, and even created four flavours of the theme to support `Official Icons`, `Angular` and `NestJs` variants in the icons. 

image

In VS Code we can execute a command to dynamically generate the theme, which allows us and the users to add/change icons dynamically. Not sure if we could do something like that in Zed because for what I see the extension APIs are far way limited, but I guess that if there's a way to contribute custom commands we could be able to accomplish the same. For now, the easiest way was to generate those theme variants.

In any case, I'm happy to make whatever is better for the Zed ecosystem. When I saw there were many different `vscode-icons` extensions I thought it would be easier to just to create a new one and let the other authors decide what to do than starting lots of threads to try to arrive to some agreement among us. The main priority was to have the icon set published as soon as possible.

chawyehsu:
> and even created four flavours of the theme to support Official Icons, Angular and NestJs variants in the icons.

This is great and I believe it solves https://github.com/chawyehsu/zed-vscode-icons-theme/issues/3 in a way I didn't think of.

maxdeviant:
@robertohuertasm Since this is the official VSCode Icons extension, let's use `vscode-icons` for the extension ID.

This will replace the existing extension published under that ID. Since you already have the version at `12.15.0` and the current one is at `0.2.0`, that won't be a problem.

robertohuertasm:
@maxdeviant I think I managed to do the changes. Let me know if I need to do anything else. Thanks for your help!

Just to be clear, I used the `extensions/vscode-icons` submodule, I didn't touch the one related to @chawyehsu 's extension. I hope that's what you were expecting.

robertohuertasm:
@maxdeviant one thing: when I go to the [Zed Extensions website](https://zed.dev/extensions?filter=icon-themes&query=vscode) and I click over the extension I get redirected to another one.

It seems that the url [https://zed.dev/extensions/vscode-icons](https://zed.dev/extensions/vscode-icons) redirects to [https://zed.dev/extensions/chawyehsu-vscode-icons](https://zed.dev/extensions/chawyehsu-vscode-icons).

EDIT: I've tried installing the extension from the Zed IDE and it works there, so it seems it's only affecting the url redirection.

robertohuertasm:
@chawyehsu maybe you asked for that redirection at some point?

robertohuertasm:
@maxdeviant @chawyehsu couldn't find any issue or PR talking about that redirection... so I'm guessing it may be related to both extensions sharing the same name and that affecting how urls work?

maxdeviant:
> @maxdeviant @chawyehsu couldn't find any issue or PR talking about that redirection... so I'm guessing it may be related to both extensions sharing the same name and that affecting how urls work?

We don't have a system in place for this right now.

One option would be to update the extension name/description to indicate that it is deprecated and direct users to the official VS Code Icons extension.

robertohuertasm:
@maxdeviant what I mean is that you're currently redirecting users to the wrong extension. So anyone clicking over this extension will be redirected to @chawyehsu's. So you may have some sort of redirection going on for sure. At least, that's what I'm experiencing in all the machines I've tried this link: https://zed.dev/extensions/vscode-icons

maxdeviant:
> @maxdeviant what I mean is that you're currently redirecting users to the wrong extension. So anyone clicking over this extension will be redirected to @chawyehsu's. So you may have some sort of redirection going on for sure. At least, that's what I'm experiencing in all the machines I've tried this link: https://zed.dev/extensions/vscode-icons

Ah, I see. Sorry, I misunderstood what you meant.

I removed that redirect, so https://zed.dev/extensions/vscode-icons should now show the right extension.

robertohuertasm:
@maxdeviant Thank you! ❤️

PR #3633, opened 2025-10-22
Add theme VS Code Dark Modern

TheWirv:


thewirv:
@cla-bot check

maxdeviant:
@TheWirv How is this different from https://github.com/kcamcam/vscode_dark_modern.zed?

thewirv:
> @TheWirv How is this different from https://github.com/kcamcam/vscode_dark_modern.zed?

I mean, I cannot from the top of my head tell you exactly which colors and which properties differ, but it is different. If you pull the two and make a diff, you would see. But I thought that wasn't a problem. There are quite a few seemingly "duplicate" extensions after all.
Ultimately, mine is more accurate IMHO, considering Zed's theming "limitations".

thewirv:
Why is the CI failing? I have added the MIT license...

maxdeviant:
> > @TheWirv How is this different from https://github.com/kcamcam/vscode_dark_modern.zed?
> 
> I mean, I cannot from the top of my head tell you exactly which colors and which properties differ, but it is different. If you pull the two and make a diff, you would see. But I thought that wasn't a problem. There are quite a few seemingly "duplicate" extensions after all. Ultimately, mine is more accurate IMHO, considering Zed's theming "limitations".

We try to avoid having duplicate extensions, especially when the only way to distinguish them by a user is a by looking for an extra `-` in the name.

thewirv:
> > > @TheWirv How is this different from https://github.com/kcamcam/vscode_dark_modern.zed?
> > 
> > 
> > I mean, I cannot from the top of my head tell you exactly which colors and which properties differ, but it is different. If you pull the two and make a diff, you would see. But I thought that wasn't a problem. There are quite a few seemingly "duplicate" extensions after all. Ultimately, mine is more accurate IMHO, considering Zed's theming "limitations".
> 
> We try to avoid having duplicate extensions, especially when the only way to distinguish them by a user is a by looking for an extra `-` in the name.

I understand, and what would you recommend doing now?

thewirv:
@maxdeviant I totally understand your point, but my theme is truly the most faithful adaptation, especially now that Zed has introduced the `keyword.control` option, and this theme is using it. However, I don't agree with you on the fact the "the only way to distinguish them by a user is a by looking for an extra `-` in the name", because most users will look for a theme on the Extensions tab in Zed. And over there the human-readable title and the description are featured, and they both make it very clear from my theme that it is the best choice for users who want the "VS Code Dark Modern" experience in terms of visuals. In the end, there are already three basically "VS Code Dark" themes, and I don't think a fourth one would make that situation a lot much worse.

image

thewirv:
This does not need the "awaiting response" label anymore imho

thewirv:
> @maxdeviant I totally understand your point, but my theme is truly the most faithful adaptation, especially now that Zed has introduced the `keyword.control` option, and this theme is using it. However, I don't agree with you on the fact the "the only way to distinguish them by a user is a by looking for an extra `-` in the name", because most users will look for a theme on the Extensions tab in Zed. And over there the human-readable title and the description are featured, and they both make it very clear from my theme that it is the best choice for users who want the "VS Code Dark Modern" experience in terms of visuals. In the end, there are already three basically "VS Code Dark" themes, and I don't think a fourth one would make that situation a lot much worse.
> 
> image

I just want to reiterate my previous point, and also point out that since writing that (and since the release that added support for the `keyword.control` option) none of the other options have been updated, so my theme is still the most uptodate and accurate option in terms of "VS Code Dark" theme.

thewirv:
I made some changes to the most popular VS Code Dark theme instead. https://github.com/kcamcam/vscode_dark_modern.zed/pull/26

PR #3665, opened 2025-10-25
Add Lince theme extension

xaviduds:


xaviduds:
@cla-bot check

xaviduds:
@cla-bot check

xaviduds:
Explanation: This is a theme I've always created in other editors because I've never encountered something that was as Black/White as I liked. Other B&W themes still maintain a lot of border colors and are not "flat" and homogeneous

maxdeviant (approved):
Thanks!

PR #3668, opened 2025-10-26
Add Popping and Locking Theme extension

randoneering:
This is my first attempt at adding the popping and locking theme to zed.

screen1
screen2


randoneering:
@cla-bot check

randoneering:
@maxdeviant  handled. Thanks for that.

PR #3687, opened 2025-10-27
Add min-theme-plus theme extension

NiFate:
This PR adds a new Min Theme Plus theme, which is based on the original Min Theme but mixed with the color scheme from One Dark Pro.


NiFate:
@cla-bot check

PR #3711, opened 2025-10-29
Update VS Code Icons repository URL

robertohuertasm:
I realized the URL of the repo was wrong. It's working because Github keeps the redirection. In any case, I think it is better to correct it.


robertohuertasm:
@maxdeviant I guess I'll have to update the repo url when we do another release?

MrSubidubi (approved):
Fine as is. Thanks!

PR #3745, opened 2025-11-01
Update path to java extension and release v6.6.0

playdohface:
Closes https://github.com/zed-extensions/java/issues/112
This makes the java extension repo point at the zed-extensions/java.git repo instead of the one that is has been moved from, and bumps the version to the latest release


MrSubidubi:
Ah, you'll need to update the version in the extension repo as well. See https://github.com/zed-extensions/clojure/pull/11 for an example of that.

playdohface:
Done :)

PR #3716, opened 2025-10-29
Add OpenCode extension

rekram1-node:

PR #3758, opened 2025-11-03
Update extension CLI sha to get Agent Servers

rtfeldman:
The previous CLI did not have access to the new Agent Servers stuff, which means that https://github.com/zed-industries/extensions/pull/3716 got published with all that metadata getting stripped out - leading to the behavior that the local dev extension works, but the official one does not. 😅 

To fix that, after landing this PR we'll need to ask the Opencode folks to bump that extension to 0.1.1 just to trigger a fresh publish with the updated CLI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ concurrency:
   cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
 
 env:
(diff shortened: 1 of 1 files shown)


maxdeviant:
@rtfeldman We'll need to build a new version of the extension CLI, which can be done using this action:

https://github.com/zed-industries/zed/actions/workflows/publish_extension_cli.yml

And corresponding script:

https://github.com/zed-industries/zed/blob/main/script/bump-extension-cli

PR #3769, opened 2025-11-04
Add new theme One Dark Pro for Zed

hadez8877:
One Dark Pro for Zed brings the most popular Visual Studio Code theme to the Zed editor, offering a carefully crafted dark theme that enhances code readability and reduces eye strain during long coding sessions.

### Features

* **Multiple Theme Variants**: 
  * One Dark Pro (Classic)
  * One Dark Pro Darker
  * One Dark Pro Flat
  * One Dark Pro Mix
  * One Dark Pro Night Flat

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -2072,7 +2072,7 @@ version = "0.0.10"
 
 [one-dark-pro-for-zed-theme]
 submodule = "extensions/one-dark-pro-for-zed-theme"
-version = "0.0.11"
+version = "0.0.12"
 
 [one-dark-pro-max]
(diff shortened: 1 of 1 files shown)


maxdeviant (changes requested):
You'll need to change your theme ID, as we don't want `zed` or `for-zed` anywhere in the theme ID, as all extensions are Zed extensions.

hadez8877:
> You'll need to change your theme ID, as we don't want `zed` or `for-zed` anywhere in the theme ID, as all extensions are Zed extensions.

But what do I name it, if I think there's already an extension called one-dark-pro?

PR #3782, opened 2025-11-06
Bump fsharp to 0.0.9

nathanjcollins:
Bumps fsharp to [0.0.9](https://github.com/nathanjcollins/zed-fsharp/compare/0.0.7...0.0.9)


nathanjcollins:
> Thanks for this!
> 
> One minor nit: Could you check whether `std::env::current_dir()` could replace your `get_extension_home` method? That would be our preferred API for that purpose and should also be more reliable than the workaround you have for that.

@MrSubidubi hey, thanks for the suggestion! I've just pushed this change now.

PR #3800, opened 2025-11-08
Add Doxygen language support

Clement-Lap:
Add doxygen extension

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -866,6 +866,10 @@
 	path = extensions/dogi
 	url = https://github.com/DogukanUrker/DogiZed.git
 
+[submodule "extensions/doxygen"]
+	path = extensions/doxygen
+	url = https://github.com/Clement-Lap/doxygen-zed
+
 [submodule "extensions/dprint"]
 	path = extensions/dprint
 	url = https://github.com/panikkastudio/dprint-zed.git
--- a/extensions.toml
+++ b/extensions.toml
@@ -877,6 +877,10 @@ version = "0.1.0"
 submodule = "extensions/dogi"
 version = "1.0.4"
 
+[doxygen]
+submodule = "extensions/doxygen"
+version = "0.1.0"
+
 [dprint]
 submodule = "extensions/dprint"
 version = "0.0.1"


Clement-Lap:
Hello !
Yes, it is to inject Doxygen into C/C++ comments

Clement-Lap:
Sorry if I am making a lot of mistakes, it's my first time helping an open source project, and doing pull requests

MrSubidubi:
No worries, all good and this definitely looks good!

However, it would actually be better for this to properly work if you were to add the injections for C/C++ within Zed itself. Hence, could you remove both c and cpp from your extension and open a PR in Zed itself for the injections? This is a limitation of our current injection API (one I hope to fix sometime soonish)? Then I am more than happy to merge this, but I'd like the injections to be available upstream so that you don't have to maintain the c/cpp languages in your extensions (they would just be duplicates). Does that work for you?

Clement-Lap:
Yes seems fine to me ! Tell me if i'm wrong, but do i need to add my `doxygen` repo into `zed/crates/languages/src` ?

MrSubidubi:
Not quite - your doxygen repositors stays in an extension, but the injection queries you had to get for C and C++ go into the main repo. That way, we don't duplicate language definitions and your extension will still immediately work once your extension is installed.

MrSubidubi:
So
- add the injections to C and C++ within Zed
- remove C and C++ from your extension

and we should be good to go. Does that make sense?

MrSubidubi:
Olay, regarding https://github.com/zed-industries/zed/pull/43581/ - you'd need to add the injections upstream again without removing the existing comment injections (they can exist in parallel) and then this here would need a rebase - can you look into that/are you still interested in this? 

Also, sorry for letting this drag a bit

neohe-imer:
I will take a look at this and report back. @esthertrapadoux 

zed-industries/extensions#3800
zed-industries/extensions#4763
From #3800 review notes, it sounds like the missing piece is adding C/C++ Doxygen injections upstream in Zed without breaking existing comment injections. I’m going to focus on that

MrSubidubi:
@neohe-imer The comment in that is rather stale, at the moment we are waiting for a reply from @Clement-Lap here to decide how to move forward with this extension. I do not think any other action is needed at this time.

neohe-imer:
@MrSubidubi Got it, thanks for the update.

MrSubidubi:
Will close this as stale. Still, thank you for getting the ball rolling here!

PR #3815, opened 2025-11-10
Add RelytONE MCP server extension

YuweiXiao:
This PR adds the RelytONE MCP server extension.

Extension repository: https://github.com/YuweiXiao/zed-mcp-server-relytone.git
Extension ID: mcp-server-relytone
Version: 0.1.0

RelytONE (https://data.cloud/relytone) offers all-in-one serverless Postgres with built-in extensions for vectors, full text, analytics, GIS, and time-series workloads. It exposes an MCP endpoint through the official @relyt/mcp-server-relytone package, so this extension launches that server via npx, supplies the user’s bearer token, and lets Zed connect directly to RelytONE.

Please let me know if anything needs adjusting!


MrSubidubi (changes requested):
Thanks for this! Could you please instead use the `zed_extension_api` provided APIs for installing extensions (or are there problems with that)? 

See https://github.com/zed-extensions/mcp-server-brave-search/blob/main/src/mcp_server_brave_search.rs for an example.

PR #3834, opened 2025-11-11
Update yamura to v0.0.2

Yxmura:


Yxmura:
oh boy I'm stupid why did I find this so difficult :sob: anwyays this should work now <3

maxdeviant (approved):

PR #3846, opened 2025-11-12
Update opencode to v1.0.61

thdxr:
Updating OpenCode extension to v1.0.61

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -2105,7 +2105,7 @@ version = "0.1.1"
 [opencode]
 submodule = "extensions/opencode"
 path = "packages/extensions/zed"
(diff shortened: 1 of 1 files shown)


maxdeviant:
@thdxr I updated the OpenCode to the latest version in https://github.com/zed-industries/extensions/pull/3848.

Would it be possible to update your release workflow to use a GitHub account that is able to sign the CLA? If not, I can see whether we'd be able to accept contributions that are authored by the GitHub bot account.

rekram1-node:
@maxdeviant thank you, I think i can update the action, the token i wrote it with is tied to dax but im the one managing it 

Will fix

PR #3874, opened 2025-11-14
Add ported Everforest theme

albertsko:
This PR aims to add an exact port of sainnhe's [Everforest](https://github.com/sainnhe/everforest) vim color scheme for Zed. The theme comes in regular, material, and blur variants.

We already have an Everforest theme in `zed-industries/extensions/extensions/everforest`, unfortunately it doesn't reflect the [original color palette](https://github.com/sainnhe/everforest/blob/d84d9ec896564730855c3f6705dc9a734c74a344/palette.md) accurately, while this theme does.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -822,6 +822,10 @@
 	path = extensions/everforest
 	url = https://github.com/ThomasAlban/everforest-zed.git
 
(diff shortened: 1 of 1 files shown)


albertsko:
@cla-bot check

PR #3877, opened 2025-11-14
Add Thorn theme

celioreyes:
Adding a port of Neovim's [Thorn](https://github.com/jpwol/thorn.nvim) theme


celioreyes:
@MrSubidubi updated! Let me know if that works!

PR #3889, opened 2025-11-16
Updates arkts to v0.2.0

liuyanghejerry:
This pull request updates arkts extension to 0.2.0, which integrates language server, providing more symbol semantics.

With this update, "go to definition", "find references" and "module definition" is all enabled.

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -121,7 +121,7 @@ version = "0.0.1"
 
 [arkts]
 submodule = "extensions/arkts"
-version = "0.1.0"
+version = "0.2.0"
 
 [asciidoc]
 submodule = "extensions/asciidoc"


MrSubidubi:
Thanks! 

Minor note: 

Instead of

```rust
            Ok(None) => {
                download_language_server()?;
                zed::set_language_server_installation_status(
                    language_server_id,
                    &zed::LanguageServerInstallationStatus::Downloading,
                );
            }
```

I think

```rust
            Ok(None) => {
                zed::set_language_server_installation_status(
                    language_server_id,
                    &zed::LanguageServerInstallationStatus::Downloading,
                );
                download_language_server()?;
            }
```

would make more sense, but this should not block the changes here and is fine for a later patch. Thanks!

PR #3891, opened 2025-11-16
Add git diff extension

eloi-menaud:


eloi-menaud:
@cla-bot check

MrSubidubi:
Thanks, but we should have language support for Git Diff built-in: 

https://github.com/zed-industries/zed/blob/e833d1af8d8ba7b3445b07df7356108886cb40aa/crates/languages/src/diff/config.toml#L2

Can we not instead extend the highlighting there? Or does this extension add anything to that? Even then, could you look into contributing this into core instead?

eloi-menaud:
Sorry, I think it hasn't been managed because it doesn't work on my side (v0.212.6 NixOS), but it must be on my side then. 

Moreover, `@diff.plus` and `@diff.minus` tags are used but regarding the [doc](https://zed.dev/docs/extensions/languages#syntax-highlighting) they do not exists, are they implemented but absent from the documentation, or do they not point to anything?

MrSubidubi:
> Sorry, I think it hasn't been managed because it doesn't work on my side (v0.212.6 NixOS), but it must be on my side then.

No need to apologize, I'll also take a look tomorrow morning and see whether this might be on us. Could you elaborate what the issue was here?

> Moreover, @diff.plus and @diff.minus tags are used but regarding the [doc](https://zed.dev/docs/extensions/languages#syntax-highlighting) they do not exists, are they implemented but absent from the documentation, or do they not point to anything?

A contributor fixed this a few days ago and this will roll out next week to preview, the week after to Stable IIRC (might be a week earlier and already on preview). Short summary is that we should 
- document syntax highlighting keys better (probably more than half of what is available is not documented, if not more)
- enable better fallback highlighting for such cases

Both is on my TODO-list, but does not have the highest priority sadly. But I definitely keep it in mind and might actually look at the latter rather soon.

Gonna close this given the circumstances, thank you still for getting this PR up! And please, also elaborate on what issues you ran into. Thanks!

PR #3914, opened 2025-11-18
Add ghostty-zed extension

Else00:
Added a simple extension that enable the syntax highlighting for the Ghostty config file from https://github.com/bezhermoso/tree-sitter-ghostty

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -994,6 +994,10 @@
 	path = extensions/gentle-dark
 	url = https://github.com/gentlelionstudios/gentle-dark-zed.git
 
+[submodule "extensions/ghostty"]
+	path = extensions/ghostty
+	url = https://github.com/Else00/ghostty-zed-extension.git
+
 [submodule "extensions/git-firefly"]
 	path = extensions/git-firefly
 	url = https://github.com/d1y/git_firefly.git
@@ -3385,6 +3389,3 @@
 [submodule "extensions/zwirn"]
 	path = extensions/zwirn
 	url = https://codeberg.org/polymorphicengine/zwirn-zed-extension.git
-[submodule "extensions/ghostty"]
-	path = extensions/ghostty
-	url = https://github.com/Else00/ghostty-zed-extension.git


Else00:
@cla-bot check

Else00:
@cla-bot check

Else00:
@cla-bot check

Else00:
@cla-bot check

Else00:
Hi, and thanks for merging the Ghostty extension, @MrSubidubi.

After it landed, I did some more testing on both macOS and Linux and I am seeing some inconsistent behavior around how Zed applies the Ghostty language based on `path_suffixes`. I am not completely sure I have mapped every pattern perfectly, because the behavior is not fully consistent, but there is probably enough here to see whether this is expected or a bug.

Environment:

* Zed 0.213.3
* macOS 26.1
* Ubuntu 24.04.3 LTS (aarch64, Parallels)

Ghostty config paths I am testing:

* macOS: `~/Library/Application Support/com.mitchellh.ghostty/config`
* Linux: `~/.config/ghostty/config`

The language configuration in the merged extension is:

```toml
# languages/ghostty/config.toml
name = "Ghostty"
grammar = "ghostty"
path_suffixes = ["ghostty/config"]
line_comments = ["#"]
scopes = ["source.ghostty"]
```

My understanding is that `path_suffixes = ["ghostty/config"]` should match any file whose full path ends with `ghostty/config`, which would include both paths above. While debugging, I also tried temporarily changing it to:

```toml
path_suffixes = ["ghostty/config", "config/ghostty/config"]
```

and I even created a test file at `com.mitchellh.ghostty/config` on Linux to see how that interacted with the suffixes. Those experiments did not fundamentally change the core issue described below, so I do not think this is only about the exact suffix string.

What I am actually seeing, on both macOS and Linux, and for both paths:

* If I start Zed "cold" and open the Ghostty config as the very first file (for example `zed ~/.config/ghostty/config` on Linux, or `zed "~/Library/Application Support/com.mitchellh.ghostty/config"` on macOS), the buffer often opens with the language shown as `Unknown` in the bottom-right selector. Ghostty is not applied automatically in that case.
* If Zed is already running with other files open, and I then open the same config paths (either from the command line or via Ghostty's `open_config` shortcut), the file is usually detected correctly as Ghostty and the language selector in the bottom-right shows "Ghostty".
* When detection fails and the language is `Unknown`, if I use the language selector in the bottom-right of the status bar to switch the buffer to "Ghostty", the syntax highlighting works as expected. That suggests that the grammar and highlight queries are fine, and the problem is only the initial language selection.

The same general pattern applies for both:

* `~/Library/Application Support/com.mitchellh.ghostty/config` on macOS
* `~/.config/ghostty/config` on Linux

On Linux I also wired Ghostty's `open_config` to Zed with:

```bash
xdg-mime default dev.zed.Zed.desktop text/plain
```

After that, `Ctrl+,` in Ghostty correctly opens the config file in Zed. However, the behavior is still split: when Zed is already "warm" with other buffers open, the file tends to be detected as Ghostty; when Zed is started just to open that file, it is more likely to come up as `Unknown`.

Because the behavior is inconsistent, I cannot claim a 100% deterministic repro, but roughly:

1. Opening a Ghostty config as the first and only file when Zed starts often gives me a buffer with language `Unknown`.
2. Once Zed is running and has other files open, opening the same paths tends to give me a buffer with language "Ghostty".

Given that:

* the `config.toml` actually loaded by Zed contains `path_suffixes = ["ghostty/config"]`,
* both of the real Ghostty config locations end with `ghostty/config`, and
* manually switching the language to "Ghostty" via the bottom-right selector always works and produces correct highlighting,

my impression is that this might be related to when extension languages are registered relative to when the first buffer is created, rather than only a mistake in the extension configuration. But I am not certain, and the inconsistency makes it a bit hard to pin down.

If there is something obviously wrong in how I am using `path_suffixes`, or if there is a known limitation around language detection at startup that I should work around in the extension, I am happy to adjust it. Otherwise this might be worth a closer look on the Zed side.

MrSubidubi:
Thank you very much for the detailed write-up! 

I think this might be because we hit slightly different code paths based on how "warm" your Zed instance is, which should obviously not be the case. 

Would you mind opening an issue over at https://github.com/zed-industries/zed with all the context here? Can't promise to get to this anytime soon, but would definitely like to have this tracked. Thanks!

PR #3927, opened 2025-11-19
Update elle URL to point at the original repository

acquitelol:


acquitelol:
Does the version need to change? The extension's contents are identical, it is just pointing to a different repository.

PR #3928, opened 2025-11-19
Update Pytest Language Server extension to v0.9.0

bellini666:


bellini666:
> Thanks!
> 
> Minor note for the next version: You'll most likely want to version the download you have somehow, as the check whether the binary is installed will always succeed after the first download I believe. So either the filename or foldername should depend on the release version.
> 
> You do also need to remove the older versions yourself currently. But the current code looks fine, so happy to merge as is - definitely an improvement, so thanks.

Thanks, will take a look at that for the next release 🙏

PR #3936, opened 2025-11-20
Update opencode to v1.0.80

thdxr:
Updating OpenCode extension to v1.0.80


benbrandt:
Newer version merged today

PR #3983, opened 2025-11-25
Add SFTP extension

andreyc0d3r:


andreyc0d3r (commented):
Add sub module reference

andreyc0d3r:
@cla-bot check

albertsko:
This would be huge! 🤞

maxdeviant (changes requested):
Thank you for the PR!

The way we're retrieving the language server in the extension isn't going to be compatible with different operating systems: https://github.com/andreyc0d3r/zed-sftp/blob/653fc16318e96846314dff7bc5a196f97e6f32f9/src/lib.rs#L15-L22

You'll want to publish the language server to npm so that it can be installed from there.

andreyc0d3r:
@maxdeviant  thanks, PR updated with changes to how the script resolves the path

andreyc0d3r:
> Thank you for the changes, but
> 
> > You'll want to publish the language server to npm so that it can be installed from there.
> 
> still holds - we do not copy these files into the extension, so you'll need to publish the server to NPM and then use the https://docs.rs/zed_extension_api/latest/zed_extension_api/fn.npm_install_package.html and similar APIs to install the server in the environment.

Thank you for sharing the extra info, I will attempt to follow your instructions and will get back with an update.

MrSubidubi:
Will close this as stale for the time being, feel free to ping me and happy to reopen!

PR #3986, opened 2025-11-25
Update neocmake to v0.1.1

k0tran:
Starting from v0.9.0 neocmakelsp releases switched from providing executables to providing archives. This update brings support for those.

You can read more about it here: https://github.com/k0tran/zed_neocmake/issues/20

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -1976,7 +1976,7 @@ version = "0.1.0"
 
 [neocmake]
 submodule = "extensions/neocmake"
-version = "0.1.0"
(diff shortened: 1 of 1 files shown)


merlinz01:
In case it was overlooked, there's not yet a v0.1.1 release or tag in the extension repo.

k0tran:
> In case it was overlooked, there's not yet a v0.1.1 release or tag in the extension repo.

Thanks! It was overlooked, but now it is there

PR #4025, opened 2025-11-29
Add go-snippets-plus extension

ksckaan1:
added new extension for golang snippets

https://github.com/ksckaan1/zed-go-snippets-plus

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1098,6 +1098,10 @@
 	path = extensions/go-snippets
 	url = https://github.com/ayberkgezer/go-zed-snippets
 
+[submodule "extensions/go-snippets-plus"]
+	path = extensions/go-snippets-plus
+	url = https://github.com/ksckaan1/zed-go-snippets-plus
+
 [submodule "extensions/golangci-lint"]
 	path = extensions/golangci-lint
 	url = https://github.com/zed-extensions/golangci-lint.git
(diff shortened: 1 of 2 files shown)


maxdeviant:
How does this differ from https://github.com/ayberkgezer/go-zed-snippets?

It would be great if we could contribute any missing features to that extension instead of creating a new one.

ksckaan1:
@maxdeviant Actually, there is no missing feature. However, the extension you mentioned does not use the standard snippet prefixes that others use.

VScode and the Nova editor use the snippets I've implemented. Therefore, I want to use the same prefixes for users switching from other editors.

The current extension uses the `go-` prefix for snippets. However, as a Go developer, I can't imagine typing the `go-` prefix before writing code in normal situations.

maxdeviant:
> @maxdeviant Actually, there is no missing feature. However, the extension you mentioned does not use the standard snippet prefixes that others use.
> 
> VScode and the Nova editor use the snippets I've implemented. Therefore, I want to use the same prefixes for users switching from other editors.
> 
> The current extension uses the `go-` prefix for snippets. However, as a Go developer, I can't imagine typing the `go-` prefix before writing code in normal situations.

That does seem a bit odd. We should see if the author of the `go-snippets` extension would accept a PR to remove those prefixes.

MrSubidubi:
With https://github.com/ayberkgezer/go-zed-snippets/pull/5 merged, I am gonna close this. Really appreciate you contributing the fix upsteam for everyone!

PR #4026, opened 2025-11-30
Add Panda+ Theme

lucianoratamero:


lucianoratamero:
@cla-bot check

lucianoratamero:
I've updated the paths and ran the validations locally, so it should be good to go :]

PR #4030, opened 2025-12-01
Add Web Component Language Server

break-stuff:


break-stuff:
Ah, thanks! Let me retry.

break-stuff:
@maxdeviant the Zed extension is a package in a monorepo. Is it possible to submodule the extension like that?

maxdeviant:
> @maxdeviant the Zed extension is a package in a monorepo. Is it possible to submodule the extension like that?

Yes! You can use the `path` field to point to a subdirectory inside of the submodule.

For example:

https://github.com/zed-industries/extensions/blob/1234fcd75ded39e373fccca578a110d6949fbda7/extensions.toml#L57-L60

break-stuff:
Replaced with #4181

PR #4067, opened 2025-12-04
Add Typing Sounds to Zed Extensions

evrett-k:
Simple server extension that plays a sound when typing, currently only one sound to be played with planned settings later.


abit-foggy:
@

Uploading 522303872-37dc9701-b7d3-43d7-b71e-e85cb33e128f.mp4…

abit-foggy:
holy, im geeked

PR #4082, opened 2025-12-08
Add BQN language extension

dunyakirkali:
The Zed [BQN](https://mlochbaum.github.io/BQN/) extension.

Based on [tree-sitter-bqn](https://github.com/shnarazk/tree-sitter-bqn).


dunyakirkali:
I found another extension so will be closing this PR: https://github.com/DavidZwitser/zed-bqn

PR #4096, opened 2025-12-09
Update SonarQube MCP Server to v1.4.0

nquinquenel:


maxdeviant (approved):
@nquinquenel Just FYI, it looks like the only change in this version is the version bump.

In the future, you only need to publish a new version of the extension if there are actual changes within the extension itself.

PR #4110, opened 2025-12-10
Add todo-tree slash command extension

atrtde:


alexandretrotel:
@cla-bot check

Mte90:
any updates for this?

alexandretrotel:
I'll update to v0.2.1 and we're good to go I guess.

alexandretrotel:
I've done the update, should I do anything else?

MrSubidubi (commented):

Unfortunately, I am not sure whether this extension misuses the slash command interface - slash commands for agents are intended as just that and nothing else, so I am unsure as to whether we want to merge have this in its current form.

From how I see it, this should rather be done via commands from extensions, which is on our radar, yet now obviosly not supported. 

Could you elaborate whether I am missing something here or whether that's the case?

alexandretrotel:
You’re right—using slash commands here is a bit of a workaround ^^. I just thought the feature was cool. I understand this isn’t the intended approach, especially since extension-based commands aren’t supported yet.

Would you recommend holding off on merging until there’s a proper approach, or is there a way you’d suggest I adapt this for now?

MrSubidubi:
> I just thought the feature was cool.

Understandably so, sincerely hope you had some fun hooking this up.

> Would you recommend holding off on merging until there’s a proper approach, or is there a way you’d suggest I adapt this for now?

Yes, I'll do just that and close out due to the mentioned reasons. What I think you could perhaps do is look into setting up tasks for this, which could stil produce some of what you are looking for here, and/or look into writing an LSP wrapper, although I myself am not sure whether that makes that much sense. But the stuff with tasks could definitely work and might be worth a shot!

And still, thanks for looking into this and getting this up in the first place!

alexandretrotel:
Thanks, I'll be thinking about that!

PR #4122, opened 2025-12-10
Add ZX extension

nurulhudaapon:
This extension add support for `.zx` file. `.zx` file is an extension to Zig language file with support for HTML like syntax. The extension currently provides syntax highlighting and LSP support using ZLS.

Repository: https://github.com/nurulhudaapon/zx

image

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -3505,3 +3505,7 @@
 [submodule "extensions/zwirn"]
 	path = extensions/zwirn
 	url = https://codeberg.org/polymorphicengine/zwirn-zed-extension.git
+
+[submodule "extensions/zx"]
+	path = extensions/zx
+	url = https://github.com/nurulhudaapon/zx.git
--- a/extensions.toml
+++ b/extensions.toml
@@ -3574,4 +3574,4 @@ version = "0.1.0"
 [zx]
 submodule = "extensions/zx"
 path = "editors/zed"
-version = "0.1.0-dev.364"
+version = "0.1.0-dev.366"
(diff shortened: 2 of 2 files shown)


nurulhudaapon:
@cla-bot check

MrSubidubi (changes requested):
Thanks for this!

Few things:

- We should not require the installation logic for zls and instead be able to use the zls provided by the Zig extension. We can add support for zls to run in these files by adding it to https://github.com/zed-extensions/zig/blob/main/extension.toml similar to how this is done for https://github.com/zed-extensions/emmet/blob/main/extension.toml.
- Please remove the icon theme, as this does only add support for this language and sadly we do not currently have a good way for merging these. I'd prefer for that to not be in there for now.


I do think the duplication for the debug locator is needed, if you have more context there, please let me know. That is something that has not yet occurred too often.

MrSubidubi:
I see movement upstream although not here - do you want to update this PR at some point or should I close for now? Happy to look at it at another time if needed. Thanks!

nurulhudaapon:
> I see movement upstream although not here - do you want to update this PR at some point or should I close for now? Happy to look at it at another time if needed. Thanks!

Yeah, we can close this for now. I’ll open another one after creating proper LSP for ZX. For other editor I have been using middleware since Zed doesn’t support middleware yet, will just build standalone LSP.

MrSubidubi:
Sounds good, thanks for the update!

nurulhudaapon:
HI @MrSubidubi, created a new PR. Now Ziex has it's own LSP!

https://github.com/zed-industries/extensions/pull/6307

PR #4130, opened 2025-12-11
Add odoo-lsp extension

Desdaemon:
This PR adds an extension to support the [Odoo LSP](https://github.com/Desdaemon/odoo-lsp).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -222,6 +222,10 @@
 	path = extensions/ayu-darker
 	url = https://github.com/k4yt3x/zed-theme-ayu-darker.git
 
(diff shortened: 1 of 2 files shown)


Desdaemon:
@cla-bot check

MrSubidubi (changes requested):
Thanks for this!

Will merge, but I'd appreciate if you could look into slightly modifying how you check for the currently installed version. https://github.com/zed-extensions/odin/blob/main/src/odin.rs is an example of this that does not require another file to be present.

Looks good otherwise, thanks!

PR #4131, opened 2025-12-12
Update Popping and Locking to v1.0.1

randoneering:
Changed focus colors to red so there was better visibility to the directory you were selecting (or file).


randoneering:
Ah! I apologize. Thank you for that.

PR #4139, opened 2025-12-14
Add Syscript language support

IsCoffeeTho:


IsCoffeeTho:
@cla-bot check

MrSubidubi (changes requested):
You'll need to change the grammar repository URL to be a `https` remote, otherwise this will not work. 

Besides that, looks good!

IsCoffeeTho:
> You'll need to change the grammar repository URL to be a `https` remote, otherwise this will not work.
> 
> Besides that, looks good!

Thanks for the feedback
Current grammar repository is `https://github.com/IsCoffeeTho/tree-sitter-syscript`
would it be more aligned if i add `.git` to the end, otherswise this is a resolved problem

MrSubidubi (changes requested):
Nice, thanks! Can you please update the submodule to the commit I should review? Thanks!

MrSubidubi (changes requested):
Appreciate it, thanks for adjusting (and sorry for the delay, was unfortunately sick).

Lastly, could you please remove the LSP from your `extension.toml` as well as all Rust code? Since this is effectively unused, it would be better if we were to remove it upfront. After, we should be good to go. Thanks!

MrSubidubi:
Closing as stale, since the [last comment](https://github.com/zed-industries/extensions/pull/4139#pullrequestreview-3843471929) is unadressed upstream. Happy to repoen with that adressed.

PR #4140, opened 2025-12-14
Add new dracula-flat-theme extension

WangWindow:


WangWindow:
@cla-bot check

MrSubidubi (changes requested):
Thanks for this!

Could you elaborate what this is about? This seems to be a fork of the original darkula theme, but indistinguishable in parts. I see some colors modified and perhaps a blur set for that, but could you elaborate more as to why you wanna publish this/what the difference is and why this is worth to be a separete extension?

WangWindow:
Dear @MrSubidubi 

Previously, while using the VSCode editor, I chose the Dracula theme. 

After switching to the Zed editor, I noticed that the color scheme of the [Dracula theme](https://zed.dev/extensions/dracula) here (as shown in figure 1, compared with VSCode's Dracula theme) is noticeably different from what I used before.

Figure 1


The main motivation for creating this new theme stems from my desire for a more unified color scheme. Additionally, I applied italic styles to the comment sections to better distinguish them. Figure 2 showcases the effect of my theme after readjusting the colors and transparency.

Figure 2

Of course, everyone has their own preferred style, and it does not mean that the adjusted theme is superior to the original. I simply wish to provide an additional option. 

With this in mind, perhaps I should rename my plugin to "Dracula Flat" to better differentiate it from the existing Dracula theme in Zed and to facilitate choices according to personal preferences for other users as well.

Thank you for your understanding and support!

PR #4145, opened 2025-12-15
add zcode extension

williamspiegel:
adds zcode extension for Xcode support: https://github.com/williamspiegel/Zcode

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -3438,6 +3438,10 @@
 	path = extensions/zabby
 	url = https://github.com/arne-fuchs/zabby
 
+[submodule "extensions/zcode"]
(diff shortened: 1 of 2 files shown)


williamspiegel:
@cla-bot check

williamspiegel:
@cla-bot check

MrSubidubi:

Unfortunately, that misuses the slash command API and it not something that we want to officially endorse - slash commands for agents are intended as just that and nothing else, so we will not merge this for the time being, sorry to disappoint. 

This will require the API to officially support commands from extensions, which is on our radar, yet now obviosly not supported. So we will close this for now - thank you for your efforts though anyway.

PR #4178, opened 2025-12-17
Add Markdoc extension

louiss0:
## Summary
- add Markdoc extension as HTTPS submodule pinned to v1.0.0
- register entry in extensions.toml with path="extension"
- run pnpm sort-extensions to keep metadata ordered

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,7 +66,7 @@ jobs:
       - name: Package extensions
         run: pnpm package-extensions
         env:
-          HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+          REF_NAME: ${{ github.ref_name }}
           SHOULD_PUBLISH: ${{ github.ref_name == 'main' }}
           S3_BUCKET: ${{ secrets.S3_BUCKET }}
           S3_ACCESS_KEY: ${{ secrets.S3_ACCESS_KEY }}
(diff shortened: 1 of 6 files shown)


louiss0:
@cla-bot check.

louiss0:
> You'll need to softlink your license in the directory the extension lives in and then update the submodule here. 
> 
> Also note that `cfg!(windows)` will not work in the WASM environment, you can get the current OS from https://docs.rs/zed_extension_api/latest/zed_extension_api/fn.current_platform.html Just as a note, the license thing is the only thing actually blocking here.

How do I do that?

MrSubidubi:
https://github.com/zed-extensions/csharp/blob/40c23be541fa445e47a184b2354f789550bbf85c/src/language_servers/roslyn.rs#L62-L85 is an example in another context from another repo. You can just query the OS and check for `os == zed::Os::Windows` basically

louiss0:
> https://github.com/zed-extensions/csharp/blob/40c23be541fa445e47a184b2354f789550bbf85c/src/language_servers/roslyn.rs#L62-L85 is an example in another context from another repo. You can just query the OS and check for `os == zed::Os::Windows` basically

I need to delete that folder!

MrSubidubi:
Sorry, I do not think I follow here. Instead of the `cfg!` directive, you need to use the value provided by the API.

MrSubidubi:
Closing as stale, but please do reopen if you decide to pick this up again! Thanks!

PR #4208, opened 2025-12-21
Update css-modules-kit to v0.1.1

mizdra:
release note: https://github.com/mizdra/css-modules-kit/releases/tag/release-1766310088497#:~:text=css%2Dmodules%2Dkit%2Dzed%400.1.1

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -622,7 +622,7 @@ version = "0.0.6"
 
 [css-modules-kit]
 submodule = "extensions/css-modules-kit"
-version = "0.1.0"
+version = "0.1.1"
 path = "crates/zed"
 
 [csv]


mizdra:
I have a question. When updating packages, how do you determine whether the Pull Request author is trustworthy? Is there a mechanism to verify Pull Request authors?

I believe merging a Pull Request created by someone other than the extension maintainer could lead to serious security issues. For example, imagine an attacker creates a malicious extension repository (https://github.com/css-modules-kit/css-modules-kit) and says: “Hey, Zed maintainers. I've changed the extension repository from https://github.com/mizdra/css-modules-kit to https://github.com/css-modules-kit/css-modules-kit. Can you merge this?”

Is there a mechanism in place to prevent such attacks?

MrSubidubi:
As for your question - we will generally not change upstream repositories without prior discussion. So if someone comes in and would replace your submodule URL with his own, we'd either not merge, ping you and discuss this first before proceeding. Exceptions would be where we determine upstream to be unresponsive, but then we'd fork into `zed-extensions` organization, which we obviously consider trustworthy.

Hope that makes sense, feel free to ask if it does not!

mizdra:
Thank you for answering my question. I'm relieved to hear there are systems in place to prevent such attacks.

What made me a bit uneasy is that the only mechanism to prevent attacks seems to be manual checks. Manual checks are prone to human error. Layered defense is important for preventing attacks. I think it would be good to have automated checks in addition to manual ones.

For example, it would be good to have CI that detects PRs modifying the extension repository and requires author approval. https://github.com/DefinitelyTyped/DefinitelyTyped has a similar mechanism. That might be a useful reference.

MrSubidubi:
We are currently in the process of setting up security policies for the entirety of Zed Industries, and we'll most likely revisit this soon. I'll mention your input once the topic comes up, thank you for providing it.

PR #4246, opened 2025-12-24
Update oxc to v0.4.1

nrayburn-tech:


nrayburn-tech:
@maxdeviant any idea why CI might be failing? I tried to run the same steps locally and I'm not seeing the same error. I may be using some different versions than CI though (Rust versions, zed-extension CLI version, etc.)

chungweileong94:
> @maxdeviant any idea why CI might be failing? I tried to run the same steps locally and I'm not seeing the same error. I may be using some different versions than CI though (Rust versions, zed-extension CLI version, etc.)

I'm actually getting the same error on my fresh install of rust environment, it was due to missing `wasm32-wasip2` target. I'm not too familiar with the rust development, but I think you need to define the `wasm32-wasip2` target as part of `rust-toolchain.toml`.

https://github.com/zed-industries/zed/blob/main/rust-toolchain.toml

PR #4251, opened 2025-12-25
Add Asciidoc Zed extension

louiss0:
This PR adds a new Zed Asciidoc Extension! It has language support and snippets! With a better Asciidoc Parser behind the scenes!


MrSubidubi:
Thanks for this! Aside from us already having an Asciidoc extension at https://zed.dev/extensions/asciidoc, the upstream repo seems to be gone, so will close. Otherwise, we'd need a compelling reason to gi with this in the first place anyway, so if you want to resubmit, please elaborate the need for a separate extension. Thanks!

PR #4266, opened 2025-12-26
Add AutoHotkey extension

alfredomtx:
## Summary

Adds AutoHotkey v1 language support for Zed editor.

## Features

- Syntax highlighting for `.ahk` files
- Comment toggling (`Ctrl+/` or `gcc` in vim mode)
- Document outline and go-to-symbol (`Ctrl+Shift+O`)
- Auto-indentation
- Bracket matching
- Run scripts directly from the editor (with task configuration)

### Supported Syntax

| Category | Elements |
|----------|----------|
| Comments | Line (`;`), block (`/* */`), doc (`/** */`) |
| Literals | Strings, numbers (int, hex, float), arrays |
| Operators | Assignment, comparison, arithmetic, logical, bitwise |
| Keywords | `if`, `else`, `while`, `loop`, `for`, `switch`, `case`, `try`, `catch`, `throw`, `return`, `class`, `new`, etc. |
| Functions | Definitions, calls, parameters with defaults |
| Commands | 38 built-in commands (`MsgBox`, `Send`, `Sleep`, `Run`, `Gui`, etc.) |
| Variables | Regular, built-in (`A_ScriptDir`, `A_Now`, `A_TickCount`, `Clipboard`, `ErrorLevel`, etc.), `%var%` dereferencing |
| Objects | Property access, method calls, array indexing |
| Hotkeys | Modifier combinations (`^!+#`) |
| Directives | `#Include`, `#NoEnv`, `#SingleInstance`, etc. |
| Escape sequences | `` `n ``, `` `t ``, `` `r ``, etc. |

### Advanced Syntax

- **Hotstrings:** `:options:trigger::replacement` syntax with option parsing
- **Classes:** Full OOP with `extends`, static members, magic methods (`__New`, `__Delete`, `__Get`, `__Set`, `__Call`)
- **Object literals:** `{key: value}` syntax with identifier, string, and numeric keys
- **Exception handling:** `try`/`catch`/`finally` blocks with `throw`
- **Switch statements:** `switch`/`case`/`default` control flow
- **Ternary expressions:** `condition ? value1 : value2`
- **Control flow:** `for...in` loops, `break`/`continue`, labeled statements

## Repository

https://github.com/alfredomtx/tree-sitter-autohotkey

## License

MIT

PR #4269, opened 2025-12-27
Add autohotkey-debugger extension

alfredomtx:
## Summary
Adds the AutoHotkey Debugger extension for Zed.

**Features:**
- Debug AutoHotkey v1 scripts with breakpoints
- Step through code (step in, step over, step out)
- Variable inspection during debugging
- Uses the [autohotkey-debug-adapter](https://github.com/alfredomtx/autohotkey-debug-adapter) DAP implementation

This complements the existing `autohotkey` tree-sitter extension (#4266) to provide a complete AutoHotkey development experience in Zed.

**Repository:** https://github.com/alfredomtx/zed-autohotkey-debugger
**License:** MIT

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -209,10 +209,10 @@
 [submodule "extensions/autohotkey"]
 	path = extensions/autohotkey
 	url = https://github.com/alfredomtx/tree-sitter-autohotkey.git
+
 [submodule "extensions/autohotkey-debugger"]
 	path = extensions/autohotkey-debugger
 	url = https://github.com/alfredomtx/zed-autohotkey-debugger.git
-
 
 [submodule "extensions/autumnal-marscape"]
 	path = extensions/autumnal-marscape
--- a/extensions.toml
+++ b/extensions.toml
@@ -18,10 +18,6 @@ version = "0.1.1"
 submodule = "extensions/activitywatch"
 version = "0.1.2"
 
-[autohotkey-debugger]
-submodule = "extensions/autohotkey-debugger"
-version = "0.1.0"
-
 [ada]
 submodule = "extensions/ada"
 version = "0.4.0"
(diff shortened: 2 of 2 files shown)


MrSubidubi:
Thanks for this! Is there any reason to not add this to the extension you want to add in https://github.com/zed-industries/extensions/pull/4266?

alfredomtx:
> Thanks for this! Is there any reason to not add this to the extension you want to add in #4266?

Hi @MrSubidubi, just to keep as different repos, as they serve different purposes. Although they complement each other, it's easier to maintain/update

alfredomtx:
@MrSubidubi is there anything else pending to be able to merge?

MrSubidubi:
I think we would still prefer if you were to add this to the already published extension in an update - I do not really see what you mean by 

> it is easier to maintain/update

but feel free to elaborate on that, and we genreally prefer for extensions to provide as much as possible, especially for language extensions. They do serve different purposes, that said, only together will the experience be quite good for users. Also, we generally tend to ship debuggers with the language wherever we can and I'd honestly prefer if were were to not make this an exception here, especially given that you own the two repositories.

Feel free to provide arguments as well if you disagree, but how I currently see it, the cost of adding it to the other repository is minimal, as there is no Rust code involved so far, and everything could just go in a `lib` folder or whatever name you prefer there.

alfredomtx:
@MrSubidubi I had the "separation of concerns" mindset as my reason to keep it separated, and because the debugger repository is a fork of another repository. But I completely understand and agree with your point, for a language extension, it makes much more sense to have everything into one for a better user experience.

I'm closing this PR as I open a new one to update the main extension with the debugger merged into it, thank you for your attention, wish you a great day ☺️

Here's the PR to update the existing extension: https://github.com/zed-industries/extensions/pull/4383

PR #4298, opened 2025-12-30
Add semgrep language server

zeelsheladiya:
This PR adds the Semgrep extension to the Zed extensions registry.

**Semgrep** is a fast, open-source static analysis tool for finding bugs and enforcing code standards at editor, commit, and CI time. This extension integrates the semgrep lsp directly into Zed to provide real-time diagnostics and security findings as you type.

Repository: https://github.com/zeelsheladiya/zed-semgrep Extension ID: semgrep

## Features

- **Zero-Config Setup:** Automatically detects semgrep on the system. If missing, it attempts to installing it automatically via brew (macOS) or pip (Linux) so it works out of the box.
- **Real-time Diagnostics:** Highlights bugs and security vulnerabilities directly in the editor.
- **Multi-Language Support:** Supports JavaScript, TypeScript, Python, Go, Java, C, C++, Rust, JSON, Terraform, and more.
- **Customizable:** Fully configurable via Zed's settings.json. Users can override the binary path, pass custom arguments (e.g., to use a specific ruleset), and set environment variables.
- **Smart Defaults:** Runs with sensible default rules (p/security-audit, p/secrets) so users see value immediately without needing a local configuration file.

## How to Test

1. Clone the extension repository: `git clone https://github.com/zeelsheladiya/zed-semgrep`
2. Open Zed and go to Extensions -> Install Dev Extension.
3. Select the cloned `zed-semgrep` directory.
4. Open a file in a supported language (e.g., a Python file with known vulnerable code like `eval(user_input)`).
5. Verify that Semgrep findings appear as diagnostics in the editor.

## Screenshots

https://github.com/user-attachments/assets/2b363a80-b473-4d1c-9762-d83dc4152c23


zeelsheladiya:
@cla-bot check

MrSubidubi:
That said, you can provide support for this through the command API: https://docs.rs/zed_extension_api/latest/zed_extension_api/process/struct.Command.html Using that, users have control over this and can opt-in or opt-out

zeelsheladiya:
Thanks for the detailed feedback, that makes sense, and I understand the concern 👍

I’ll remove the automatic installation logic so the extension no longer installs Semgrep into the user’s global environment without explicit consent. For the initial version, I’ll update it to return a clear error when the Semgrep binary is not found, instructing users to install it manually, similar to how the Nginx extension handles this.

I’ll also look into adding optional, opt-in support via the Command API, so users can explicitly choose to install or manage Semgrep through a command rather than having it happen implicitly.

I’ll push an updated commit shortly. Thanks again for the guidance!

zeelsheladiya:
I’ve updated the PR to remove the auto-installation script entirely:
- The extension no longer attempts to run brew or pip. It simply looks for `semgrep` in the user's `$PATH` using Zed's `worktree.which` API.
- Clear Error Messaging: If `semgrep` is missing, it now returns a clear error message instructing the user to install it manually, matching the behavior of other extensions like `Nginx`.

Issue #1406, opened 2024-09-13
Lean4


femtomc:
### Check for existing issues

- [X] Completed

### Language

Lean4

### Tree Sitter parser link

https://github.com/Julian/tree-sitter-lean

### Language server link

https://github.com/leanprover/lean4/tree/master/src/Lean/Server

### Misc notes

There's also a React app to support the Infoview window: https://www.npmjs.com/package/@leanprover/infoview

blackbird314:
@uargos 

Hi, how about the progress of tree-sitter-lean? I made a [lean4 zed extension](https://github.com/blackbird314/zed-lean4) but the current tree-sitter is not ideal.

drdozer:
I would love good lean4 support within zed, supporting a similar interactive environment to that in vs code. I hate to be that person, but have you tried a modern agentic coder to build some of the translations and glue? It's the sort of thing they seem to be quite good at particularly if you can get them to incrementally check their work against a library of test cases.

lnay:
I dug a bit into this topic today. What I found is that the infoview is only part of the issue here (which on its own could be possible in an external window if you built a process in the middle of zed and the lean4 language server).

The lean4 language server expects the LSP spec to be violated for specific messages and really needs non-standard messages for more than just the infoview.

My takeaway was that there's little point pursuing this seriously before extensions have the ability to provide hooks for certain buffer events and send custom LSP messages. No amount of workarounds with extra language server shims is going to get around this.

lnay:
Relevant conversation (involving a zed team member too): https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/Lean.20.2B.20Zed.20integration/near/505671509

ofek:
It looks like support was introduced in #2482 that uses [owlx56/zed-lean4](https://github.com/owlx56/zed-lean4) which is being [actively](https://github.com/zed-industries/extensions/pull/4340) maintained.

MrSubidubi:
Yup this is now available: https://zed.dev/extensions/lean4 🎉

ofek:
I opened a follow-up https://github.com/zed-industries/extensions/issues/5059

PR #4340, opened 2026-01-02
Update lean4 to v0.1.0

owlx56:


owlx56:
Thanks for the review! Here is the context regarding the snippets and the installation process:

1. About the 9.3k lines of snippets: These snippets provide Unicode Abbreviation support (e.g., converting `\alpha` to `α`), which is essential for writing Lean 4. While VSCode and Neovim use custom extension code to handle these mappings, Zed’s current WASM-based extension API doesn't yet support custom input methods. Providing these as standard JSON snippets is the most robust way to achieve this functionality in Zed. The line count reflects the 1,000+ unique symbols supported by Lean 4, with each requiring several lines of JSON metadata.

2. About Auto-installation & Paths: In VSCode, users are usually prompted via a notification before the extension runs an installation script(`elan-init.sh`/`elan-init.ps1`). In the Zed environment, the plugin handles the download of the official `elan-init` binary if elan is not detected.

To ensure this is safe and transparent for the user:

**Official Tooling**: It uses `elan`, the standard Lean toolchain manager (identical in philosophy to `rustup`).

**User-Local Only**: Everything is installed into `~/.elan` in the user's home directory. It does not require root/sudo access and does not modify system-wide folders (like `/usr/bin`).

**What's installed**: It downloads the elan manager, the lean compiler, and the lake build system in `~/.elan/bin`.

**Environment**: The installer typically appends the `~/.elan/bin` path to the user's shell profile (e.g., `.zshrc` or `.bashrc`).

**Clean Uninstall**: the user can completely remove the installation and delete path by simply running `elan self uninstall`.

This approach balances the current UI limitations of Zed plugins with the need for a functional Lean development environment.

owlx56:
@MrSubidubi 

1. I'm not sure which scope to use... `lean.json`, `lean4.json` or `lean 4.json`?
2. Year, it will modify the user environment. Perhaps I can wait until the Zed extension supports a “custom prompt” feature, and then let users decide whether to install?
3. An unrelated question: I couldn’t find any instruction about `[[capabilities]]` at `extension.toml` in Zed docs. Did I miss something?

MrSubidubi:
Sorry for the wait again, was sick this week.

1. I think `lean 4.json` should be the proper scope, I see you also already did that. Did it work?
2. You could adapt the approach the java extension takes and have a key for this in the language server settings. Let that be false by default and make it opt-in with docs. Example of the Java extension: https://github.com/zed-extensions/java?tab=readme-ov-file#quick-start `jdk_auto_download` would be it. But you are free to also wait here until we improve things in that area
3. We lack docs for this... I stumbled across this a few days ago, made a note of it to look into that soon. https://github.com/zed-extensions/java/blob/3392cb7db0d48d08a1522e4c0716aefd05a42f3e/extension.toml#L25-L28 is an example of how this would work.

owlx56:
1. `lean 4.json` is right. It works.
2. Sure, I will try this way.

ofek:
Thanks for the update! Should https://github.com/zed-industries/extensions/issues/1406 be closed or is it being kept open intentionally until the infoview item in the [TODO list](https://github.com/owlx56/zed-lean4/blob/2d71b3e1faa3559b058ccf7d5bfd0de0d2e5ad48/README.md#todo-list) is complete? I see that there's a [discussion](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/Lean.20.2B.20Zed.20integration/near/505671509) about the challenges of implementing that in Zed.

owlx56:
> Thanks for the update! Should #1406 be closed or is it being kept open intentionally until the infoview item in the [TODO list](https://github.com/owlx56/zed-lean4/blob/2d71b3e1faa3559b058ccf7d5bfd0de0d2e5ad48/README.md#todo-list) is complete? I see that there's a [discussion](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/Lean.20.2B.20Zed.20integration/near/505671509) about the challenges of implementing that in Zed.

Considering that the todo list feature is difficult to complete in a short time, I personally think we can close the issue.

ofek:
Was this PR closed accidentally?

owlx56:
> Was this PR closed accidentally?

Hi, I just opened a new pr with action workflow.

ofek:
Ah, I didn't notice https://github.com/zed-industries/extensions/pull/4935 until now. Thank you!

PR #4348, opened 2026-01-03
Add Pathy extension

gokulp01:
## Summary

This PR adds the Pathy Zed extension to the community registry.

Pathy provides filesystem path completions inside Python string literals (via a lightweight sidecar LSP server), helping users quickly autocomplete paths in common cases like `open("./")` and `Path("./")`.

## Key Features

- Path completions in Python strings (e.g., `open("./")`, `Path("./")`)
- Context-aware behavior with a safe fallback for explicit path prefixes (`./`, `../`, `~`, absolute paths)
- Does not replace your primary Python LSP — runs as a secondary server focused on completions only
- Cross-platform support via packaged server binaries (downloaded and cached by the extension)  
- Configurable (base directory strategy, ignore patterns, max results, hidden files, etc.)
- Works with external libraries like pandas etc. 

## Links

- Extension repo: https://github.com/gokulp01/pathy

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2390,6 +2390,10 @@
 	path = extensions/path-of-exile
 	url = https://github.com/egibs/poe.zed.git
 
+[submodule "extensions/pathy"]
+	path = extensions/pathy
+	url = https://github.com/gokulp01/pathy.git
+
 [submodule "extensions/pbxproj"]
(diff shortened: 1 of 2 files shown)


gokulp01:
@cla-bot check

PR #4352, opened 2026-01-03
Remove unavailable repos

eagafonov:
`git submodule update` fails because of a few repos are not available anymore:

- https://github.com/brook-code-theme/zed-theme.git
- https://github.com/one-thing-theme/zed.git
- https://github.com/xaviduds/theme-lince.git

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -362,10 +362,6 @@
 	path = extensions/brainfuck
 	url = https://github.com/JosephTLyons/zed-brainfuck.git
 
-[submodule "extensions/brook-code-theme"]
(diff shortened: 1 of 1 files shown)


MrSubidubi:
Thank you for looking into this!

We are in active contact with the (former) owners of these repositories and will take appropriate action once we hear back from them. Due to the holidays, we have not taken immediate action, but will over the course of the next week should no action be taken from their side.

In any case, while I appreciate your efforts, this will be something we as Zed staff will be doing, so I am gonna go ahead and close your PR for the time being. Still, thank you for bringing this to our attention again!

PR #4353, opened 2026-01-03
Add marc-dracula

marcorentap:


MrSubidubi:
Thank you for this, but we do not think it is a sufficient change to warrant a distinct extension, given that this basically modifies only a few properties. You can specify these yourself and/or share these using theme-overrides, https://zed.dev/docs/themes#theme-overrides, and we do not think this is sufficient enough to warrant a distinct publish for this.

Still, thank you for getting this up!

PR #4360, opened 2026-01-04
Add gruvbox-material-neovim theme

RiverMatsumoto:
Ported over the gruvbox-material theme from [f4z3r's gruvbox-material.nvim](https://github.com/f4z3r/gruvbox-material.nvim/tree/master) for neovim to look as similar as possible. I understand there are existing gruvbox material themes that exist but this one is supposed to look as similar as possible to the neovim one.


MrSubidubi (approved):
Thanks for providing a reasoning behind this!

PR #4374, opened 2026-01-05
Add warm-light theme

YangGangUEFI:
Warm Light theme for the Zed code editor.

https://github.com/YangGangUEFI/warm-light-theme

image


YangGangUEFI:
> Thanks for this! We do not allow extension IDs with `zed` in them, as all extensions are Zed extensions, so you'll need to change the name of your theme (or at least the ID) in that regard. Thanks!

Hi, thank you for reminding me. I have updated the ID to `warm-light-theme`.

PR #4396, opened 2026-01-07
Add npm-package-json-checker language server

e-simpson:
Adds an LSP that highlights outdated packages in a package.json and shows the changelogs

https://github.com/e-simpson/zed-npm-update-checker

screenshot

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2270,6 +2270,10 @@
 	path = extensions/not-material-theme
 	url = https://github.com/iamawatermelo/zed-hct-theme-maker
 
+[submodule "extensions/npm-package-json-checker"]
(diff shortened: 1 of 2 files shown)


e-simpson:
Thanks @MrSubidubi!
I was wondering what the process is to apply to become the recommended extension popup for package.json files?

MrSubidubi:
We currently have no process for this, mostly because this does not happen for language servers and only for language support in general (we tend to have only one extension for a given file format). 

I think we could look into this at a later point should your extension gain reasonable traction, however, not right now. Feel free to reach out again at some point should this be the case. Thanks!

e-simpson:
I see, sounds good thank you!

PR #4401, opened 2026-01-07
Add todo-tree extension using Zed tasks

atrtde:
This PR reintroduces the todo-tree extension using **Zed tasks** instead of agent slash commands.

Context:
- The previous PR (#4110) was closed because slash commands are intended only for agents.
- This implementation follows the recommendation to use tasks instead.

Changes:
- Removed all slash-command usage
- Implemented todo-tree functionality via Zed tasks and guides for users to do it
- Extension remains non-agent-based and aligns with current platform expectations

Happy to adjust based on feedback (again)!

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -62,6 +62,10 @@
 	path = extensions/aira
 	url = https://github.com/talison-cardoso/aira-zed
 
+[submodule "extensions/aizen-theme"]
+	path = extensions/aizen-theme
+	url = https://github.com/vivy-company/zed-aizen-theme.git
+
 [submodule "extensions/alabaster"]
 	path = extensions/alabaster
 	url = https://github.com/tsimoshka/zed-theme-alabaster.git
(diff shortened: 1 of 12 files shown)


alexandretrotel:
Sure, you're right. Firstly, I thought it would be better not to use slash commands, as they do not follow the intended usage. So, in this case, less is more. Secondly, the extension provides syntax highlighting, and that's true—it’s the only “true” feature for now. Zed Task commands are a plus, and indeed, they’re manual (which is a bad thing for developer experience); I agree. The core of the extension will exist once Zed extensions allow extending the current IDE UI. The LSP path seems to be the best one, but I’m not skilled enough yet to go down that road (maybe later), and I’m also lacking time (the semester is starting, and I’ve got to study).

MrSubidubi:
Yeah, sorry for pushing you back and forth here. I sincerely appreciate your time that went into this, and sorry I was always the road block at the end. I hope you can resonate with my reasoning here, and us lacking proper support for external commands is definitely something that does bother me, but it will probably require some time to get this right.

I wish you all the best for your ongoing semester and best of luck with exams, feel you there as I have to juggle Zed and my studies simultaneously as well. 

Given this, I'll have to close, sorry again, but thank you for being this reasonable and understanding.

alexandretrotel:
That's okay!

PR #4427, opened 2026-01-09
Add logcat language support

vinnom:
I'm moving from VSCode to Zed and I use editor a lot for reading Android Logcat files, so I missed a extension for highlighting it.

I hope I didn't mess any thing.

For reference, I aimed highlighting color captures based on theme I use locally (Ayu Dark). This might behave different on other themes.


vinnom:
@cla-bot check

vinnom:
> Thanks for this, looks good! Only thing, could you perhaps change the extension ID to just `logcat`?

sure! I updated the extension ID to logcat, but kept the submodule as extension/logcat-highligter due to the directory naming. Would you suggest to keep everything as "logcat" only? I had named as "-highlighter" as a pattern, I have highlighter on "name" as well to prevent people thinking this might be a complete utility to capture/analyze logcat.

MrSubidubi:
CI will yell at you if you don't 😄 Quickly took care of it

PR #4444, opened 2026-01-10
Add tan-jade theme

Ukendio:


MrSubidubi (changes requested):
Please make sure to not touch all files in this repository and change their file modes.

Also, there are several things missing, see https://zed.dev/docs/extensions/developing-extensions#extension-license-requirements and https://zed.dev/docs/extensions/developing-extensions#publishing-your-extension

Also also, language like https://github.com/Ukendio/Tan-Jade/blob/758392c8f287a77a2c5394c0b0e3b73eaccda1df/extension.toml#L5 is not something we want to publish here in this repository. While we are not strictly against publishing such themes and/or trolling people with such, the language within the repository needs to be improved in order for this to be published. 

Closing given all of the above, please reopen with another PR that adresses thee issues. Still, thanks for taking the time to get this up and running.

PR #4446, opened 2026-01-10
Add Symposium extension

nikomatsakis:
Q: In the toml file I have

```toml
[symposium]
submodule = "extensions/symposium"
path = "zed-extension/prod"
version = "1.1.1"
```

...but I'm not sure what this version number represents. The actual extension is up to 1.3 by now, but it seems silly to keep this version up-to-date..?

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -3206,6 +3206,10 @@
 	path = extensions/symbols
 	url = https://github.com/sebastiandotdev/zed-symbols.git
 
+[submodule "extensions/symposium"]
+	path = extensions/symposium
+	url = https://github.com/symposium-dev/symposium.git
+
 [submodule "extensions/syntax"]
 	path = extensions/syntax
 	url = https://github.com/syntaxfm/syntax-zed-theme.git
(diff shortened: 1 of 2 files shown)


nikomatsakis:
@cla-bot check

benbrandt:
@nikomatsakis the issue with the versioning as you have it: we likely won't have the behavior you want.
We currently update agent servers when the extension updates, so if the extension never updates, we won't pull latest again.

Until there is an acp registry it would likely be better to have the versions tied together

nikomatsakis:
@benbrandt ok, it seems odd to have the submodule so that it can "auto-update" but I still have to update this file with a PR?

benbrandt:
The submodule doesn't autoupdate

nikomatsakis:
OK, well, that explains *that*

nikomatsakis:
I've updated, @benbrandt, though \[the setup\] still seems a bit "un-DRY" to me, even if submodules don't auto-update.

benbrandt:
thanks. I guess part of the issue is that your extension could serve more than just your agent server. all of this will definitely be revisited in the future when we focus some more on the extension story for sure.

nikomatsakis:
@benbrandt Got it. I've updated the version number and I believe this is ready to mergre!

PR #4484, opened 2026-01-13
Add http-client extension

cramhead:
This adds the http-client extension to the extensions listing.
This is the first extension I've made.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1630,6 +1630,10 @@
 	path = extensions/http
 	url = https://github.com/tie304/zed-http.git
 
+[submodule "extensions/http-client"]
+	path = extensions/http-client
+	url = https://github.com/cramhead/http-client
+
 [submodule "extensions/hubl"]
 	path = extensions/hubl
 	url = https://github.com/blueambr/zed-hubl.git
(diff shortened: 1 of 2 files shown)


bobbymannino:
i'm not a maintainer or anything but i think your missing a couple things. in `extensions.toml` you need to add your extension, i just copied and pasted one and edited the fields. then you also need to run `pnpm sort-extensions`. I think this may bump your extension, looking forward to using this!

cramhead:
> i'm not a maintainer or anything but i think your missing a couple things. in `extensions.toml` you need to add your extension, i just copied and pasted one and edited the fields. then you also need to run `pnpm sort-extensions`. I think this may bump your extension, looking forward to using this!

Thanks @bobbymannino. I don't know how I gapped on that. I've updated the PR.

cramhead:
@MrSubidubi no problem on the delay. Life happens sometimes. 
Given your statement, I expect the approach is that users will install the http support independently if they want syntax highlighting.  I didn’t read that was the approach. 
I’ll make that change shortly

cramhead:
@MrSubidubi Removed the http syntax highlighting. 
Eventually, I'd like to consider adding highlighting that would include syntaxlighting for request, response and comments, but that's something for the future.

PR #4489, opened 2026-01-14
Update GLSL to v0.2.0 and Proto to v0.3.1

amustaque97:
Closes: https://github.com/zed-industries/zed/issues/46820


Thinkseal:
@yeskunall you wanted to be pinged for this

MrSubidubi (changes requested):
Thanks for doing this!

Can you please also bump the Proto one here to 0.3.1?

PR #4505, opened 2026-01-16
Update wat to v0.2.2

g-plane:


BlackAsLight:
Is there an eta on when this will be merged? I'd prefer to not go back to zero syntax highlighting.

PR #4532, opened 2026-01-18
Add Meson Muon extension

jobpaardekooper:
Add an alternative extension that adds language support for `meson.build` files with syntax highlighting and the `Muon LSP` to replace `MesonLSP`. Submitting this extension since the current [`extensions/meson`](https://github.com/hqnna/zed-meson) extension is unmaintained and was archived. The existing `extensions/meson` extension uses [`MesonLSP`](https://github.com/JCWasmx86/mesonlsp) for the language server but this project is also unmaintained and archived. This extension uses the LSP feature of [`Muon`](https://muon.build).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1598,6 +1598,10 @@
 	path = extensions/ink
 	url = https://github.com/yuna0x0/zed-ink.git
 
+[submodule "extensions/intellij-light-theme"]
+	path = extensions/intellij-light-theme
+	url = https://github.com/chandruscm/intellij-light-theme-zed.git
+
 [submodule "extensions/intellij-newui-theme"]
 	path = extensions/intellij-newui-theme
 	url = https://github.com/kpitt/zed-theme-intellij-newui.git
--- a/extensions.toml
+++ b/extensions.toml
@@ -1625,6 +1625,10 @@ version = "0.0.7"
 submodule = "extensions/ink"
 version = "0.0.2"
 
+[intellij-light-theme]
+submodule = "extensions/intellij-light-theme"
+version = "0.1.0"
+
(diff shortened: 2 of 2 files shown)


jobpaardekooper:
Hello @MrSubidubi both proposed solutions are fine by me, but I think it would be best for now to leave the older extension available for a little longer.

The original extension uses [MesonLSP](https://github.com/JCWasmx86/mesonlsp) as the backing LSP which is archived and no longer maintained but still very feature rich and working near perfectly on the newest meson files.

My new extension uses the LSP mode of [Muon](https://muon.build/) as the backing LSP. This is a bit of a newer project and their focus is not solely on the LSP/analyzer side. I wanted to create this extension so it is easy to try out Muon even though it has a lot of features. The more people using it the better as it will also bring encouragement for it to be further developed. Muon is very actively maintained and clearly the future for Meson language servers.

So because of this context, I think it might be better to have both available for now. What do you think?

As far as I don't think there is a way to provide custom settings for extensions in Zed, correct? Otherwise I could support both LSPs in my new extension and replace the old one, letting users choose their preferred LSP. I have not seen the custom settings for extensions in the Zed docs but I might be wrong. Please provide a docs link in case I missed something.

jobpaardekooper:
Sounds good! I will try to update my extension so users can switch between Muon LSP and mesonlsp and then come back here in a few days with the updated ID and version etc.

MrSubidubi:
Wonderful, thank you!

jobpaardekooper:
@MrSubidubi I updated everything as you mentioned. Changed the extension ID and and version. Please review the new PR (because this one got auto closed due to my force push when syncing).

#4980

PR #4534, opened 2026-01-18
Update jsonl to v0.1.1

vim-zz:
- Update extensions/jsonl submodule to the latest commit on main
  - Switch submodule URL to SSH (git@github.com:vim-zz/jsonl-syntax-zed.git) for consistency with
    owned repo
  - Only .gitmodules and the extensions/jsonl submodule pointer are changed

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -565,7 +565,7 @@ version = "0.0.5"
 
 [code-stats]
 submodule = "extensions/code-stats"
-version = "0.2.7"
+version = "0.3.0"
 
 [codebabel-ztheme-dark]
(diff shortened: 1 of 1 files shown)


MrSubidubi (changes requested):
You'll need to bump the version of your extension both in the extension and here so that we can roll out the update

vim-zz:
@MrSubidubi any idea why the Zed editor doesn't updates the extension the new version?

MrSubidubi:
You failed to update the extension properly and I missed that because I ignored the bot message above (will now fail) and the set of changes fit our usual pattern. Fixed via https://github.com/zed-industries/extensions/pull/4697

PR #4535, opened 2026-01-18
Add supertheme4 theme

superkacper4:
# supertheme4

A dark theme for ZED editor inspired by [Sublime Text's Marina](https://github.com/hnatiukr/zed-mariana-theme/) theme, specifically designed for web development.

## Overview

**supertheme4** changes small quirks of [Sublime Text's Marina](https://github.com/hnatiukr/zed-mariana-theme/) making it easier for the web development and fixes a few issues.

## Screenshots

image

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -3506,7 +3506,7 @@ version = "0.1.0"
 
 [supertheme4]
 submodule = "extensions/supertheme4"
-version = "0.0.4"
+version = "0.0.3"
 
 [surrealql]
 submodule = "extensions/surrealql"


MrSubidubi (changes requested):
Thanks! Could you please elaborate where your theme differs from the existing Zed theme you took inspiration from and why this would warrant a separate release? Thanks!

superkacper4:
Sure 😄 So I fixed a few bugs that were in the original (eg, the diagnostics pop-up not being readable), and I changed the theme to be dimmer (the strong colors in the original made eyes more tired).

I am testing the current version by overriding the original in the settings, and I must say I see the value, so I decided to share it 😄
 
I am also planning to experiment with it further and add new variants, which would be much easier with the dedicated theme.

PR #4537, opened 2026-01-18
Add `Bearded Theme` theme

BeardedBear:
Here is my official addition of Bearded Theme, built from https://github.com/BeardedBear/bearded-theme

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -302,6 +302,10 @@
 	path = extensions/bearded-icon-theme
 	url = https://github.com/sethstha/bearded-icons-theme.git
 
+[submodule "extensions/bearded-theme"]
+	path = extensions/bearded-theme
+	url = https://github.com/BeardedBear/bearded-theme-zed.git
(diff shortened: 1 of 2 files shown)


BeardedBear:
@cla-bot check

BeardedBear:
> Thanks for this, looks good!
> 
> One question: I see you are the official author of the bearded theme and in such cases we occasionally allow original authors to take over the previous fork. I'd have to chat with folks internally regarding this first.
> 
> Would you be interested in that? Or are you fine with the existing `bearded` theme co-existing and stay with the `bearded-theme` ID?

I'd feel uncomfortable taking over the existing fork, since others took the time to port my themes because they liked them. It's totally fine to let each fork coexist - that's the beauty of open source! Within the next few days, I'll also be releasing an official port of my icon pack Bearded Icons), which will create the same situation. I already have a working version locally that just needs some minor tweaks. ;)

PR #4549, opened 2026-01-19
Add go-template-lsp extension

Jason-Abbott:
There is an existing gotmpl but that’s not a full LSP. This heavily refactors and extends https://github.com/yayolande/go-template-lsp and exposes it as a Zed extension


Jason-Abbott:
@cla-bot check

PR #4554, opened 2026-01-20
Add hledger language support

juev:
Plain-text accounting support with syntax highlighting and LSP integration.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ concurrency:
   cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
 
 env:
-  ZED_EXTENSION_CLI_SHA: bb591f1e653a492c2d6c85353181e5151dbf8ac7
+  ZED_EXTENSION_CLI_SHA: 03d8e9aee95ea6117d75a48bcac2e19241f6e667
 
 jobs:
   package:
--- a/.github/workflows/danger.yml
+++ b/.github/workflows/danger.yml
@@ -1,6 +1,7 @@
 name: Danger
 
 on:
+  merge_group:
   pull_request:
     branches: [main]
     types:
@@ -8,6 +9,8 @@ on:
       - synchronize
       - reopened
       - edited
+      - labeled
+      - unlabeled
 
 jobs:
   danger:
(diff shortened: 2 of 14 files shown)


juev:
@cla-bot check

juev:
@MrSubidubi Hello!

The language part uses the same tree-sitter-ledger grammar, but there are differences that justify a separate language definition:

  1. File extensions — hledger uses .hledger files which aren't covered by the ledger extension
  2. Comment syntax — hledger only supports ; and # comments (ledger also supports %, |, *)
  3. Highlight queries — customized for hledger-specific constructs (balance assertions, cost/price operators, etc.)

  The main value of this extension is the hledger-lsp language server. I'm open to restructuring if you'd prefer — for example, adding .hledger suffix to the existing ledger extension and shipping only the LSP here. What approach would
  you prefer?

MrSubidubi:
If you are fine with it, I'd love for you to propose the changes in the existing extension upstream and see whether these are accepted. I'd then come to a decision with you here given what happens upstream, if you are fine with that.

juev:
Thanks for the review! I've investigated the feasibility of using the ledger extension's grammar, and I'd like to explain why I believe a separate language definition is necessary despite sharing the same tree-sitter grammar.

The core issue isn't the grammar — it's the language configuration.

Both extensions use tree-sitter-ledger, but hledger and ledger have semantic differences that affect the editor experience:

1. **Comment syntax incompatibility**
  - ledger extension: ["; ", "# ", "% ", "| ", "* "]
  - hledger: only ;  and #

In hledger, %, |, and * are not comment markers. If we use the ledger language, these characters would be incorrectly highlighted as comments, and editor features like "toggle comment" would insert wrong characters.

2. **File associations**
  - ledger extension doesn't support .hledger files
  - Adding .hledger to ledger is possible, but then users would get incorrect comment handling

3. **Separate ecosystems**
  - ledger uses ledger-language-server
  - hledger uses hledger-lsp (different tool, different capabilities)

**What I could do**:

I could submit a PR to mrkstwrt/zed-ledger to add .hledger suffix, but that would give hledger users incorrect comment behavior. The alternative would be asking ledger extension to conditionally change comment markers based on file extension, which adds complexity and may not be welcome upstream.

**Proposed path forward**:

The current structure — shared tree-sitter-ledger grammar with separate language definitions — follows the same pattern as similar dialects in other ecosystems (e.g., TypeScript/JavaScript share concepts but have separate language definitions).

Would you be open to accepting this as a separate extension, given these semantic differences? I'm happy to discuss alternative approaches.

PR #4557, opened 2026-01-20
Add settings-sync extension (v0.1.0)

vittorius:
Adds the [settings-sync](https://github.com/vittorius/zed-settings-sync) extension to Zed

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2974,6 +2974,10 @@
 	path = extensions/seti-icons
 	url = https://github.com/d1y/zed-seti-icons.git
 
+[submodule "extensions/settings-sync"]
+	path = extensions/settings-sync
+	url = https://github.com/vittorius/zed-settings-sync.git
+
 [submodule "extensions/severance-theme"]
 	path = extensions/severance-theme
 	url = https://github.com/Zev18/severance-zed.git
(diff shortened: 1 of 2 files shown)


MrSubidubi:
Thanks for this! We know this is a heavily requested feature and there have been many pushes for this internally to get this in. 

Yet, whilst this is a rather requested feature, I do not think the LSP approach proposed here is a good approach for this and something that we want to officially endorse via an extension - we do want to really have this, but definitely not using an extension. I would argue that the Discord presence extension provides a different purpose than what this extension does, and would argue that there is less risk involved with having that extension as opposed to syncing settings here. Users might unknowingly leak API keys and such whilst syncing these, and I think for such reasons it would be better if we were to provide this via a native solution at some point.

You and others are free to use this as a dev extension, totally fine with me, but please understand that we will not ship this as an extension avaiable from the extension store.  I hope you can resonate with my decision here, feel free to reach out if you have concerns regarding it. Still, thank you for getting this up and ready!

vittorius:
@MrSubidubi Hey, thank you for the detailed response. Yes, I understand your concerns and the shortcomings of this extension's approach (caused by the limited power of the extensions in general). 
Thanks for the feedback anyway. I will be happy to see this as a native Zed functionality someday and more than willing to participate in contributing to it if there is any kind of collaboration initiative related to settings sync (like the recent Git Together, for instance).

PR #4565, opened 2026-01-21
Add xmake language support

jeleferai:
- Add xmake as a Zed extension with full XMake build system support
- Includes xmake_ls language server integration, syntax highlighting, and build tasks
- Extension repository: https://github.com/xmake-io/xmake-zed


waruqi:
> Thanks for this!
> 
> Could you quickly elaborate why the separate language definition for Lua is needed? Would it not suffice if we were to update the file suffixes in the Lua extension? Or do you want to have this for the language server so that it only is invoked for the correct files?
> 
> Also, just as a note: `make_file_executable` is already a noop on Windows, so no need to gate that in the extension code.

Although xmake is based on the Lua runtime, it extends many interfaces and restricts some native Lua APIs. In addition, it has its own configuration syntax and a large number of APIs, which are not available in native Lua.

Currently, there is also a separate xmake language category on GitHub.

https://github.com/github-linguist/linguist/blob/738a27e1e2f26f061a24e94f85cd37ae9be71105/lib/linguist/languages.yml#L8818

MrSubidubi:
@jeleferai I lack permissions to push to your branch and it seems that the `package` job suffers from a former GitHub outage. Can you please merge main so I can re-trigger CI? Thanks!

albifronsss:
> ... and please update the submodule here.

Apologies for struggling with this, haven't touched Zed or done much coding lately since this PR. Submodule is updated now.

MrSubidubi (changes requested):
All good, no worries. I'd love to help you with it, however, can't because the repo is bound to an org.

As per the latest CI error - please remove the `embeddings.scm` in your repo and see if that fixes things - we removed support for that recently. If the extension compiles locally for you, it might very well be that that query file is the issue. 

Thanks!

albifronsss:
> Nice, that was it. Thanks for bearing with me, the quick follow-ups and the extension!

Thank you for the patience and guidance throughout! Really appreciate it.

PR #4581, opened 2026-01-22
Add Riverpod and Freezed snippets

OppositeDragon:


OppositeDragon:
@cla-bot check

OppositeDragon:
Hello @MrSubidubi, I reviewed the apparent error, and found why it is marking the snippets as broken.
I need the character `$` to be used no only to place a cursor, but also to be used as the actual character.

For example: `abstract class Person with _$Person`
[Freezed Docs](https://pub.dev/packages/freezed#creating-a-model-using-freezed)

So when I do: `"abstract class ${1:Class} with _$${1:Class} {"`
the first `$` after 'with' is being marked as needing a number(int), thus failling parsing.

I tried this as a workaround, but also fails
`"abstract class ${1:Class} with _\u0024${1:Class} {"`

Do you think something can be made to make this possible, or is there a different direction I can take?

MrSubidubi:
You should be able to just escape the dollar sign, I think by just adding a backslash but could be wrong. There is documentation somewhere I believe, if not happy to add that. Dollar signs are definitely supported already though.

OppositeDragon:
Sorry for the late response. 
I haven't been able to find the documentation you mentioned.

I did try to use backslah to escape the `$`(dollar sign), but zed warns againt it, and even if I try to reinstall/build dev extension like that, it fails.


image

If at all possible, could you point me to the documentation, or a workaround for this.

Thanks.

MrSubidubi:
Well, you'll need to properly escape within the JSON string (https://stackoverflow.com/a/27516892 for more context I suppose) so that we can parse and then escape the symbol within Zed. So two backslashes should probably do it

OppositeDragon:
Thank you for taking the time and pointing me in the right direction.

You were right, `\\` (double backslash) did it.

I'll update the snippets and retry submitting.

I appreciate your help.

MrSubidubi:
(And sorry for the delay, was unfortunately sick)

OppositeDragon:
Thank you @MrSubidubi. Hope you are feeling much better now.

Although they are for Dart, and some of these snippets can be used in pure Dart projects, most are specific to Flutter (specially in `riverpod_snippets`). So, I didn't add `-dart` because I did not want users to think these only contain snippets for dart and not flutter.

MrSubidubi:
Appreciate the context. Can we go for `flutter` then or would you have similar concerns there?

Also, to be sure, just talking about the ID here, Name can stay as it currently is

OppositeDragon:
What about `riverpod-dart-flutter-snippets` and `freezed-dart-flutter-snippets`?

MrSubidubi:
Perfectly fine with me!

MrSubidubi (approved):
Thank you and thanks for bearing with me!

OppositeDragon:
Thank you!!!

PR #4588, opened 2026-01-23
Add RDF (SPARQL, Turtle, TriG support) and properties lang extensions

rpietzsch:


rpietzsch:
@cla-bot check

rpietzsch:
@cla-bot check

MrSubidubi (changes requested):
Thanks for this! Aside from this missing some steps from https://zed.dev/docs/extensions/developing-extensions#publishing-your-extension, I see we already have properties support provided by the java extension as well as support for turtle files from the turtle extension.

Could you elaborate why you deem separate extension necessary? We try to avoid language duplications where possible to have fewer extensions that then provide better support, so I wonder if we can just go with the SPARQL support here or even provide this in an upstream extension? Curious about what you think

MrSubidubi:
Will close for the time being as stale, but please do ping me to reopen if you decide to pick this up again. Thanks!

PR #4611, opened 2026-01-24
Add iwonder snippets

ityme:
A code snippet extension pre-loaded with a collection of essential code blocks. 
Currently focused on Python, with support for additional languages planned for future updates.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,6 +35,11 @@ jobs:
           node-version: "20"
           cache: "pnpm"
 
+      - uses: actions-rust-lang/setup-rust-toolchain@1780873c7b576612439a134613cc4cc74ce5538c # v1
+        with:
+          toolchain: "1.90"
+          target: "wasm32-wasip2"
+
       - name: Cache zed-extension CLI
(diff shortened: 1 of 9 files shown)


ityme:
@cla-bot check

ityme:
@MrSubidubi  
According to the CI logs, the failure was due to the extension ID naming convention. I've corrected the ID and pushed the update. Kindly take another look when convenient.

ityme:
> The ID/name still includes `Zed` - please think of another ID/name here please

I've applied the update by renaming the extension from ~~`snippet-zed-iwonder`~~ to `snippet-iwonder`.

ityme:
> Looks better now!
> 
> Last question: Is there a specific reason for the name, and/or can we suffix the `snippets` part instead? Not insisting, just curious, as we usually suffix snippet extensions as oppposed to prefixing as done here.

Thanks for the feedback!
Regarding the naming style, I personally tend to follow a `Noun-First` principle. I prefer placing the most essential word at the beginning, followed by descriptive qualifiers, as I find it most helpful to identify what an object is at first glance.

PR #4631, opened 2026-01-26
Add PHTML extension

WITCZAKB:


WITCZAKB:
> Thanks for this!
> 
> If the sole purpose of this is to support PHP highlighting in `*.phtml` files, can we not just update the PHP extension with that suffix?

Yes it's totally a better solution. I tried and everything works fine directly with the php extension.
I opened a new PR directly on the php extension.

PR #4656, opened 2026-01-28
Add nu-lint extension

vmeurisse:
Add an extension to integrate [nu-lint](https://codeberg.org/wvhulle/nu-lint) for linting Nushell scripts


vmeurisse:
@cla-bot check

PR #4663, opened 2026-01-29
Add Kubernetes MCP server

GGORG0:
Adds the [Kubernetes MCP Server extension](https://github.com/GGORG0/zed-mcp-server-kubernetes), which itself is an adapter for https://github.com/containers/kubernetes-mcp-server/

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1930,6 +1930,10 @@
 	path = extensions/mcp-server-grafana
 	url = https://github.com/sd2k/zed-mcp-grafana.git
 
+[submodule "extensions/mcp-server-kubernetes"]
+	path = extensions/mcp-server-kubernetes
+	url = https://github.com/GGORG0/zed-mcp-server-kubernetes
+
 [submodule "extensions/mcp-server-linear"]
 	path = extensions/mcp-server-linear
 	url = https://github.com/LoamStudios/zed-mcp-server-linear.git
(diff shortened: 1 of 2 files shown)


GGORG0:
@cla-bot check

MrSubidubi:
And apologies for the slow response, was unfortunately out sick and the queue grew a bit out of control 😅

GGORG0:
Hi, so the upstream documentation is a bit unclear, but from what I can see the server gets configured through a config file in `~/.config`, and it gets the actual Kubernetes keys from your system's `~/.kube/config`, so it seems that there isn't really anything to configure through the editor.

Please do correct me if I'm wrong, as, as I said, the docs aren't really clear on how to configure it, and I don't have much experience with MCP - either way, I've never had the need to ness with the config, it just works after it picks up my Kubernetes keys.

MrSubidubi:
Thanks for the explanation, makes sense then. How would you feel about adding an instruction that states this, so that users are aware of this being the case? E.g. https://github.com/Ataraxy-Labs/weave-zed/commit/f4f4e8935a7a5cf0fd88193aab0a1ef030166a5e added something similar. Not insisting, more curious how you would feel about that

MrSubidubi:
Closing as stale, feel free to repoen with the feedback adressed/answered. Thanks!

PR #4674, opened 2026-01-30
Update mistral-vibe to v2.0.2

Nemtecl:
Release notes:

https://github.com/mistralai/mistral-vibe/releases/tag/v2.0.2

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -2123,7 +2123,7 @@ version = "0.0.1"
 
 [mistral-vibe]
 submodule = "extensions/mistral-vibe"
-version = "2.0.1"
+version = "2.0.2"
 path = "distribution/zed"
 
 [mnemonic]


Nemtecl:
Hi @MrSubidubi 👋 

We’re already listed in the ACP registry!
I think we could stop opening PRs to Zed and ask you to remove mistral-vibe from extensions, but before we do, I’d love to confirm a couple of things:

- For current extension users, is there a smooth way to encourage migration from the extension to the registry?
- Since we’ve been in the registry from the start, I might have missed something: Are updates to Mistral Vibe in the registry handled automatically? (It seems like it, just want to double-check!)

benbrandt:
Sorry for the confusion @Nemtecl :D 

Updates are automatic, registry auto-updates and we will check for new registry version every ~1hr or so, so it will get out there.

We need to go through a deprecation phase for extensions, all things to get new agents point to the registry and we also have some callouts on the extension page.

Nemtecl:
Thank you all for your replies, fixed the conflicts 🙏

PR #4682, opened 2026-01-30
Add Oso extension

arusahni:
[Oso](https://www.osohq.com/) is an Authorization as a Service platform. A core part of our product is Polar, an authorization DSL. This extension adds initial support for the LSP, as well as syntax highlighting via our tree-sitter grammar.


maxdeviant (approved):
Thanks!

PR #4689, opened 2026-01-31
Add hybrid theme

camiha:
A dark color scheme for Zed based on [w0ng/vim-hybrid](https://github.com/w0ng/vim-hybrid).

### DIFF
--- a/.github/workflows/actionlint.yml
+++ b/.github/workflows/actionlint.yml
@@ -13,6 +13,7 @@ on:
 jobs:
   actionlint:
     runs-on: namespace-profile-2x4-ubuntu-2404
+    if: github.repository_owner == 'zed-industries'
     env:
       SHA: 03d0035246f3e81f36aed592ffb4bebf33a03106
       VERSION: 1.7.7
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,7 @@ jobs:
           clean: false
           fetch-depth: 0
 
-      - uses: pnpm/action-setup@41ff72655975bd51cab0327fa583b6e92b6d3061 # v4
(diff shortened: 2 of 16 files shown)


camiha:
@cla-bot check

camiha:
@MrSubidubi Thanks for the feedback!

I've addressed both points.
* Changed the extension ID to `hybrid-theme`
* Added the extension ID to the repo (https://github.com/camiha/zed-hybrid/blob/main/extension.toml)

MrSubidubi (commented):
Hm, could you quickly elaborate on the chosen license?

camiha:
@MrSubidubi 
The license is MIT, which is the same license use by the original.

The original vim-hybrid incorporates color palletes from vim-colors-solarized, jellybeans.vim, and Tomorrow-Night, so the original source file includes their copyright notices: https://github.com/w0ng/vim-hybrid/blob/master/colors/hybrid.vim#L426C1-L451C16

I've preserved the copyright notices as required by the MIT License and added my own for this Zed port.

MrSubidubi:
Sounds good, thank you for the context.

The license text fails both our as well as GitHubs validation for the license matching though. Could you look into updating the wrapping according to https://github.com/zed-industries/extensions/blob/d7ea01a5a691101d13b187cd15d8aabfbf2cf123/src/lib/test-licenses/test-mit-license, which passes? I just checked and the text seems to be the same, however, with that, both GitHub and us here will know that its MIT.

After, please update the submodule here in the PR to the commit with the modified text (if that works for you!). Thanks!

MrSubidubi:
Closing as stale (we since updated the license matching to be less prone to wrapping but the error seems to persist). Feel free to resubmit with the issue fixed. Thanks!

PR #4693, opened 2026-02-01
Add sonder theme

jmg-duarte:


MrSubidubi (changes requested):
Thanks for this!

A few things are missing here:
- You need to add the submodule, see https://zed.dev/docs/extensions/developing-extensions#publishing-your-extension for a detailed guide.
- Your extension will need a license.
- Please use the `sonder-theme` ID instead and change this both here as well as in your repo.

Once that is in, please update the submodule here and we'll take another look. Thanks!

jmg-duarte:
> Please use the sonder-theme ID instead and change this both here as well as in your repo.

This is a new (arbitrary) requirement not documented anywhere (that I could find), how can I help document it?

MrSubidubi (approved):
I/we tend to do this for themes, I tend to do this for themes with shorter names where it is not immediately clear this is a theme otherwise and yes, it is also more of a recent change. If you'd like to add a small note to this, https://github.com/zed-industries/zed/blob/8ff52904a28b2eb6ccbcf8da20d14ff72bd5272b/docs/src/extensions/developing-extensions.md?plain=1#L28-L34 or below might be a good place, noting that this obviously only applies to themes that want to be published at some point. 

All else looks good, thanks for the follow-up!

PR #4699, opened 2026-02-02
ci: Add merge queue trigger to danger job

MrSubidubi:
This is needed, but perhaps does not work - checking here if it does


MrSubidubi:
Closing because the trigger works, but throws errors: https://github.com/zed-industries/extensions/actions/runs/21591867479/job/62213785112

PR #4713, opened 2026-02-02
Add Idris 2 language extension

dylanbraithwaite:
Adds tree sitter and LSP support for Idris 2. 

I see there was previously a dead Idris 2 plugin which has been removed from the registry in #2860. I'm not sure if this is likely to cause any problems with name collisions, but happy to tweak the extension ID a bit if so.


dylanbraithwaite:
@cla-bot check

PR #4717, opened 2026-02-03
Add greycat language support

maxleiko:
Add [GreyCat](https://get.greycat.io/) extension


maxleiko:
@cla-bot check

maxleiko:
@cla-bot check

MrSubidubi (changes requested):
Thanks for this!

Please make https://github.com/maxleiko/zed-greycat-extension/blob/3e965fee1311cf456d30279604b81a288940e378/src/lib.rs#L49 error instead when greycat cannot be found - we should not try invoking something that does not exist. You can instead throw an error that would be more helpful to users. Thanks!

PR #4752, opened 2026-02-06
Add mise-en-place language support

FrancoTampieri:
Adds syntax highlighting and configuration support for mise-en-place (mise) files.

## Features
- Full TreeSitter syntax highlighting for all mise sections and properties
- Complete support for tasks, hooks, settings, tools, environment variables
- Official mise JSON schema included for future autocompletion
- Zero external dependencies
- MIT License

## Links
- Repository: https://github.com/TTlab-Research/mise


FrancoTampieri:
@cla-bot check

FrancoTampieri:
> Thanks for this! Could you please elaborate how/why this is needed to be available as a separate extension? From what I can see, this is just a configuration within a TOML file, so not sure where the advantage is of having a dedicated extension for this is.
> 
> Thanks!

Thank you @MrSubidubi for taking the time to look at this: I really appreciate it!

You're right that mise files are TOML under the hood. Without this extension, Zed would simply treat .mise.toml, mise.local.toml, and .mise.local.toml as standard TOML files with no domain context no awareness of what the sections, keys, or values actually mean in the mise ecosystem.

The extension adds a layer of semantic understanding on top:

- File recognition — proper association for all mise file variants

- Semantic highlighting — mise-specific constructs like task names and properties (run, depends, description), environment variables (by UPPER_CASE convention), Tera template expressions ({{ env.VAR }}), and special version strings ("latest", "lts", "system")

- JSON Schema — autocompletion and inline validation for the full mise configuration surface

A similar case would be Cargo.toml or pyproject.toml syntactically plain TOML, but with dedicated support that makes them much more usable in context.

As for the roadmap, I'd love to eventually surface mise tasks inside Zed's task system so users can run mise run <task> directly from the editor. I'm still in the early stages of exploring this: I understand that Zed's task runner isn't currently exposed to extensions, so this may need to wait for the platform to evolve. If you have any guidance on that direction, I'd be very grateful!

Thanks again for your time and feedback.

PR #4784, opened 2026-02-09
Add copilot-usage extension

chrispetrou:
Adds the `copilot-usage` extension, which provides a `/copilot-usage` slash command that displays GitHub Copilot usage stats (inline suggestions, chat, premium  requests) in the assistant panel.

**Note:** Requires a `GITHUB_TOKEN` or `COPILOT_USAGE_TOKEN` environment variable with a classic PAT (no scopes needed).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -170,6 +170,10 @@
 	path = extensions/arkts
 	url = https://github.com/liuyanghejerry/zed-arkts.git
 
+[submodule "extensions/arturo"]
+	path = extensions/arturo
+	url = https://github.com/DaZhi-the-Revelator/zed-arturo.git
+
 [submodule "extensions/asciidoc"]
 	path = extensions/asciidoc
 	url = https://github.com/andreicek/zed-asciidoc.git
--- a/extensions.toml
+++ b/extensions.toml
@@ -57,7 +57,7 @@ version = "0.1.0"
 [agnix]
 submodule = "extensions/agnix"
 path = "editors/zed"
-version = "0.12.1"
+version = "0.14.0"
 
 [aiken]
 submodule = "extensions/aiken"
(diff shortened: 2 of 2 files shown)


chrispetrou:
@cla-bot check

chrispetrou:
@cla-bot check

chrispetrou:
@cla-bot check

chrispetrou:
Hey @MrSubidubi ! Thanks for the feedback!

That's a fair point, the usage data is primarily useful to the user rather than the assistant and that was also my intention from the start. The main reason it's a slash command is that it's currently the only extension surface available. There's no way to render a panel, status bar item, or popover from an extension, so a slash command is the closest thing to a lightweight in-editor dashboard without leaving Zed.

That said, there are some cases where it could be relevant to the assistant — for example, if a user is approaching their quota limit, the assistant could be aware of that and adjust suggestions accordingly (e.g., being more concise to reduce completions usage, or suggesting to pause Copilot on boilerplate code). But I'll admit that's a secondary benefit, not the primary use case.

If the extension API eventually supports richer UI surfaces, this would be a more natural fit as a status bar widget or sidebar panel.

Also another thing that personally bugs me is that the API I'm using to get the data is an undocumented GitHub API, so it might change at some point...

My initial intension was to create something like what vscode has to show the user copilot usage but contributing to zed directly while using an undocumented GitHub API would be risky (_too much work that would probably be rejected due to the undocumented API_) and thus decided to create it as an extension instead. But yes I totally get your point!

MrSubidubi:
Well, thank you for your detailed explanation!

As you mentioned, this will require the API to officially support commands from extensions, which is on our radar, yet now obviously not supported. So we will have to close this for now - thank you for your efforts though anyway. 

PR #4386, opened 2026-01-06
Add Aztec extension

ametel01:
Aztec-nr smart contract support for Zed

PR #4787, opened 2026-02-09
Add Aztec-Noir (prev. 'Zoir') extension

DROOdotFOO:
## Summary
- Adds Aztec-Noir (prev. 'Zoir') extension v0.3.0

### Features:
- Noir zero-knowledge language support for Zed, focused around the Aztec Network
- Tree-sitter grammar for syntax highlighting
- nargo language server integration

### Repositories:
- Extension: https://github.com/DROOdotFOO/aztec-noir
- Tree-sitter grammar: https://github.com/DROOdotFOO/tree-sitter-noir

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -3970,6 +3970,10 @@
 	path = extensions/zoegi-theme
 	url = https://github.com/nikitapashinsky/zoegi-theme.git
 
+[submodule "extensions/zoir"]
+	path = extensions/zoir
+	url = https://github.com/Hydepwns/zoir.git
+
 [submodule "extensions/zokrates"]
 	path = extensions/zokrates
 	url = https://github.com/threonyl/zed-zokrates.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -306,6 +306,10 @@
 	path = extensions/ayu-themes-glass
 	url = https://github.com/jansol/zed-ayu-glass.git
 
+[submodule "extensions/aztec-noir"]
+	path = extensions/aztec-noir
+	url = https://github.com/DROOdotFOO/aztec-noir.git
+
 [submodule "extensions/azure-context-server"]
 	path = extensions/azure-context-server
 	url = https://github.com/oWretch/zed-extension-azure-mcp
@@ -4665,10 +4669,6 @@
 [submodule "extensions/zoegi-theme"]
 	path = extensions/zoegi-theme
 	url = https://github.com/nikitapashinsky/zoegi-theme.git
-
-[submodule "extensions/zoir"]
-	path = extensions/zoir
-	url = https://github.com/Hydepwns/zoir.git
 
 [submodule "extensions/zokrates"]
 	path = extensions/zokrates
--- a/extensions.toml
+++ b/extensions.toml
@@ -4017,6 +4017,10 @@ version = "0.0.1"
 submodule = "extensions/zoegi-theme"
 version = "0.4.3"
 
+[zoir]
+submodule = "extensions/zoir"
+version = "0.1.3"
+
 [zokrates]
 submodule = "extensions/zokrates"
 version = "0.1.0"
--- a/extensions.toml
+++ b/extensions.toml
@@ -310,6 +310,10 @@ version = "1.1.2"
 submodule = "extensions/ayu-themes-glass"
 version = "1.0.0"
 
+[aztec-noir]
+submodule = "extensions/aztec-noir"
+version = "0.3.0"
+
 [azure-context-server]
 submodule = "extensions/azure-context-server"
 version = "1.0.0"
@@ -4725,10 +4729,6 @@ version = "0.0.1"
 [zoegi-theme]
 submodule = "extensions/zoegi-theme"
 version = "0.4.3"
-
-[zoir]
-submodule = "extensions/zoir"
-version = "0.1.3"
 
 [zokrates]
 submodule = "extensions/zokrates"


DROOdotFOO (commented):
lgtm

MrSubidubi (changes requested):
Thanks for this!

Could you please elaborate how this is different to the existing Noir extension and/or what it provides on top of that? We usually try to avoid having duplicate language support extension where possible, hence the question here. Thanks!

DROOdotFOO:
The existing Noir extension uses https://github.com/hhamud/tree-sitter-noir which hasn't been updated since October 2023 and has an open issue about highlighting not working out of the box.

Zoir uses a https://github.com/Hydepwns/tree-sitter-noir that is actively maintained with comprehensive test coverage. Additional improvements:
  - Full editor query support (brackets, indents, outline, runnables, textobjects)
  - LSP settings support (custom args, initialization options, workspace configuration)
  - Explicit Windows handling with helpful error message
  - Tested grammar covering all major Noir language features

DROOdotFOO:
Hey mr @MrSubidubi what are your thoughts?

DROOdotFOO:
hey @MrSubidubi weekly bump looking for clarity here thanks.

DROOdotFOO:
Hey @maxdeviant or @notpeter trying to get a response here. Aztec network's mainnet is live and a functional and maintained Noir extension would be a value add.

MrSubidubi:
Sorry for the lack of response, rather busy at the moment.

I'd still be curious why you cannot just publish these fixes upstream over at https://github.com/shuklaayush/zed-noir. That way, all existing users would benefit from the improvements as opposed to needing to find an extension where things are improved. It also seems that extension is in fact actively maintained. Could you please elaborate on that?

DROOdotFOO:
The repository only appears recently updated because it was months for a response.

Zoir extension is meaningfully different and it feels not within FOSS values for you to compel me to collaborate with someone I do not wish to.

It is not only for general Noir but specifically for supporting the cryptocurrency L2 Aztec. 
This is outside of the intent of the other repo. 

If you would like for me to make this delineation more clear I'm happy to! @MrSubidubi

MrSubidubi:
Sorry for the subpar response time from my side, got pulled into an urgent project and am still catching up on all fronts sadly. 

> it feels not within FOSS values for you to compel me to collaborate with someone I do not wish to.

That is a valid opinion to have. At the same time, we try to maintain a minimal set of extension for the languages supported in the editor to provide an optimal experience for users when installing an extension (and removing the need to search for an extension that actually works for the case) or, more particular, already having one installed. By updating existing extensions, we ensure the IMO optimal experience for users, as they get a better experience without always having to look whether there's a new alternative available that better suits their needs. This is always a trade-off between the preference you mentioned and the approach we follow.

> It is not only for general Noir but specifically for supporting the cryptocurrency L2 Aztec.
> This is outside of the intent of the other repo.

Thanks for the valuable context that I do appreciate.

> If you would like for me to make this delineation more clear I'm happy to! @MrSubidubi 

I think it would be very nice if you could do that! Speficially sounds good for the extension ID as well as the name to make the intent more clear as well as make it easier for users to find this extension. Thus, please ensure that you update the naming appropriately, making the connection to Aztec clearer here.

Also, please remove https://github.com/DROOdotFOO/zoir/blob/eecaa14eeffca94c5653ccba776d7783357c6baf/src/lib.rs#L134-L142 in favor of reading this from `settings.binary.args` (or a similar name), as this will be the expected method for users to configure custom arguments for the language server.


To note, I also wanna mention https://github.com/zed-industries/extensions/pull/4386 that I'll close in favor of this one, as your extension is more fleshed out and aligns more with the guidelines we lay out for extensions at the current time. However, the other PR did go the extra step of invoking the Aztec LSP through the CLI if available. I am not as familiar with the ecosystem obviously, but wanted to point it out at this stage.

DROOdotFOO:
@MrSubidubi thanks for the review — addressed both items in the latest push:

**1. Naming (extension ID + display name)** 
Renamed: id = "aztec-noir", name = "Aztec Noir". Also renamed the source repository 
DROOdotFOO/zoir → DROOdotFOO/aztec-noir for consistency.

**2. Custom LSP arguments**
Removed the ad-hoc settings.get("args") block in favor of the structured xLspSettings.binary.arguments field. 

3. Bumped to 0.3.0 across extension.toml, Cargo.toml, and the registry extensions.toml entry.

While I was at it I also wired up the sibling fields: 
binary.path (user-specified nargo binary, takes precedence over PATH lookup and auto-download) and, 
binary.env (passed through to the spawned process) — the full CommandSettings shape is now honored.

**Re: the Aztec LSP CLI fallback from #4386**
Great shout, we can accomplish in the next PR. 
The Aztec ecosystem's aztec-nargo (forked nargo with Aztec-specific macro expansion) deserves dedicated detection logic and a separate review cycle, so I'd rather handle it with the Aztec team on its own once this lands. 
Tracking: https://github.com/DROOdotFOO/aztec-noir/issues/3

Let me know if there's anything else!

DROOdotFOO:
hey @MrSubidubi check it out

DROOdotFOO:
Bi-weekly ping @MrSubidubi ... any feedback?

DROOdotFOO:
hey again asking for review @MrSubidubi

DROOdotFOO:
hey @MrSubidubi would be super sweet if we could get some feedback here

PR #4796, opened 2026-02-10
Update iwe to v0.0.3

gimalay:
Updates the iwe extension submodule to version 0.0.3, which adds Windows platform support.

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -847,7 +847,7 @@ version = "1.0.0"
 [design-tokens]
 submodule = "extensions/design-tokens"
 path = "extensions/zed"
-version = "0.1.19"
+version = "0.1.21"
 
 [desktop]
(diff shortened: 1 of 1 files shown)


MrSubidubi (approved):
Thanks!

Regarding

```rust
if platform != zed::Os::Windows {
    zed::make_file_executable(&binary_path)?;
}
```

we handle this on the Zed-side already, it's a noop for Windows platforms. Just a note and definitely not a blocker here.

kravlost:
I still get an error: 

```
Language server iwe:

from extension "IWE" version 0.0.3: no asset found matching "iwe-iwe-v0.0.60-x86_64-pc-windows-msvc.zip"
```

gimalay:
The fix is here https://github.com/zed-industries/extensions/pull/4940

PR #4824, opened 2026-02-11
Update united-gnome to v0.3.1

Smyrnis:
In this update there are two new themes the cherry blossom gnome and the dark gnome. I also got rid of some of the italic formats (comments e.t.c).


MrSubidubi (approved):
Thanks!

PR #4827, opened 2026-02-11
Add ChipKit extension

Leonui:
Zed extension providing file associations and syntax highlighting, currently supporting:

- Verilog (.v)
- SystemVerilog (.sv, .svh, .vh)
- Tcl (.tcl)
- SDC (.sdc)
- UPF (.upf)
- CPF (.cpf)
- SKILL (.il, .ils, .sk)
-

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -518,6 +518,10 @@
 	path = extensions/cherri
 	url = https://github.com/videah/zed-cherri.git
 
+[submodule "extensions/chipkit"]
+	path = extensions/chipkit
+	url = https://github.com/Leonui/zed-chipkit.git
+
 [submodule "extensions/chocolate"]
 	path = extensions/chocolate
 	url = https://github.com/brunozalc/zed-chocolate
--- a/extensions.toml
+++ b/extensions.toml
@@ -522,6 +522,10 @@ version = "0.5.0"
 submodule = "extensions/cherri"
 version = "0.1.0"
 
+[chipkit]
+submodule = "extensions/chipkit"
+version = "0.1.0"
+
 [chocolate]
 submodule = "extensions/chocolate"
 version = "0.0.2"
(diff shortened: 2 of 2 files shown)


MrSubidubi (changes requested):
Thanks for this!

We already have support for plenty if not all of these extensions, so I am somewhat confused as to what this adds - could you elaborate on this perhaps?

Leonui:
> Thanks for this!
> 
> We already have support for plenty if not all of these extensions, so I am somewhat confused as to what this adds - could you elaborate on this perhaps?

Thanks for you reply! I admit Zed already covers Verilog, SystemVerilog, and Tcl in existing extensions. The overlap is intentional for convenience (one install for chip designers), but the core value of this extension is the languages Zed extensions do not currently support:

- SDC: I've built a custom tree-sitter grammar for this (https://github.com/Leonui/tree-sitter-sdc), though it's not employed in this release. It will be integrated in a future update. For now, SDC files get basic Tcl-based highlighting and tclsp diagnostics.  
- UPF / CPF: power intent formats used in low-power chip design. Also not currently supported in Zed.
- SKILL: scripting language for Cadence EDA tools, mapped to a Scheme grammar for now, but I might also add a custom tree-sitter for this in a future release.

Please let me know if anything else is needed.

MrSubidubi:
While I appreciate the QoL aspect of including these languages, I think for now I'd appreciate if we were to remove these from the extension and only provide the actual "new" languages, if that makes sense.

Leonui:
> While I appreciate the QoL aspect of including these languages, I think for now I'd appreciate if we were to remove these from the extension and only provide the actual "new" languages, if that makes sense.

OK I can trim it to only have these new languages. Could you please suggest how can I proceed? Should I open a new pull request? Thanks!

MrSubidubi:
You can update this PR (more specifically, the submodule here) as needed and then re-request a review once you are done.

Leonui:
Thank you! I have updated a trimmed version and please review the changes.

PR #4830, opened 2026-02-12
Add Umple language support

DraftTin:
Adds [Umple](https://www.umple.org) language support (syntax highlighting, diagnostics, completion, go-to-definition).

  - Extension repo: https://github.com/umple/umple.zed
  - LSP server: [`umple-lsp-server`](https://www.npmjs.com/package/umple-lsp-server) (auto-downloaded from npm)
  - Tree-sitter grammar: https://github.com/umple/umple-lsp/tree/master/packages/tree-sitter-umple
  - License: MIT

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -3994,6 +3994,10 @@
 	path = extensions/umka
 	url = https://github.com/michabay05/zed-umka.git
 
+[submodule "extensions/umple"]
+	path = extensions/umple
+	url = https://github.com/umple/umple.zed.git
+
 [submodule "extensions/underground-theme"]
 	path = extensions/underground-theme
 	url = https://github.com/i-amdroid/zed-underground-theme.git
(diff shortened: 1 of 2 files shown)


DraftTin:
@cla-bot check

MrSubidubi (changes requested):
Thanks for this and sorry for letting this slip through the cracks for so long!

I think overall this looks good, however, please remove the `worktree.which("node")` check - we want to install the LSP in the environment provided by Zed and not the user-provided one, thus, I'd prefer if you were to remove this and just use that environment always. Thank you!

DraftTin:
Updated — removed the `worktree.which("node")` check as requested. The extension now always uses `zed::node_binary_path()`.

Also bumped to v0.1.0 and synced the grammar rev to the latest upstream.

MrSubidubi (approved):
Thanks!

PR #4837, opened 2026-02-12
Add PaletteSwap extension

jsvensson:
Syntax highlighting and LSP support for PaletteSwap .pstheme files.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -3961,3 +3961,6 @@
 [submodule "extensions/zwirn"]
 	path = extensions/zwirn
 	url = https://codeberg.org/polymorphicengine/zwirn-zed-extension.git
+[submodule "extensions/paletteswap"]
+	path = extensions/paletteswap
+	url = https://github.com/jsvensson/paletteswap-zed
--- a/extensions.toml
+++ b/extensions.toml
@@ -2605,6 +2605,10 @@ version = "0.0.3"
 submodule = "extensions/palenight"
 version = "0.0.1"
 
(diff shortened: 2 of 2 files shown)


jsvensson:
@cla-bot check

MrSubidubi (changes requested):
Thanks for this! Could you elaborate the use case for this? From what I could read, this seems to be in early development still and reuses the HCL grammar. Could we not instead just have the language server provided here and use the language as provided by another extension?

jsvensson:
> Thanks for this! Could you elaborate the use case for this? From what I could read, this seems to be in early development still and reuses the HCL grammar. Could we not instead just have the language server provided here and use the language as provided by another extension?

Heya! The actual code structure is largely done, actually. I don't see this moving beyond the existing HCL grammar.

The Zed extension was largely a "because I could" thing, I can't say I'm too well-versed in how to structure this as an extension. If this isn't the proper way to get an LSP + default file extension associated with a format as an extension, then I'm all ears.

MrSubidubi:
Sorry for dropping the ball here, was pulled into another project.

Basically, you could just provide the language server, have it use the other extension language and then deal with the file association within the language server code. So you'd need two extensions for this to work properly, with the advantage that we do not need to maintain two languages for that.

Since CI is still failing after months, will close for now, but happy to take another look if you want to invest more here. I suppose you use this as a dev extension anyway so I hope it fulfills your use case. And again, sorry for the slow response

PR #4840, opened 2026-02-13
Add MiniZinc language support

Dekker1:
This adds a (currently pretty simple) extension for the [MiniZinc](https://www.minizinc.org/) language, providing syntax highlighting and bracket completion for `mzn`, `dzn`, and `fzn` files. It uses the tree-sitter parsers developed by the MiniZinc developers.


Dekker1:
@cla-bot check

PR #4848, opened 2026-02-14
Add Tree-sitter test language support

kylegoetz:
Tree Sitter tests are their own language, and this adds highlighting and outline pane support for them via a new extension.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,7 +35,7 @@ jobs:
           node-version: "20"
           cache: "pnpm"
 
-      - uses: actions-rust-lang/setup-rust-toolchain@v1
+      - uses: actions-rust-lang/setup-rust-toolchain@1780873c7b576612439a134613cc4cc74ce5538c # v1
         with:
           toolchain: "1.90"
           target: "wasm32-wasip2"
--- a/.gitmodules
+++ b/.gitmodules
@@ -1822,6 +1822,10 @@
 	path = extensions/lydia
 	url = https://github.com/dimitrisnl/lydia-zed-theme
 
+[submodule "extensions/mach"]
+	path = extensions/mach
+	url = https://github.com/octalide/mach-zed.git
+
 [submodule "extensions/macos-classic"]
 	path = extensions/macos-classic
 	url = https://github.com/huacnlee/zed-theme-macos-classic.git
@@ -2949,6 +2953,10 @@
 [submodule "extensions/risor"]
 	path = extensions/risor
 	url = https://github.com/rubiojr/zed-risor
+
+[submodule "extensions/robot-framework"]
+	path = extensions/robot-framework
+	url = https://github.com/mendes-jose/zed-robot.git
 
 [submodule "extensions/roc"]
 	path = extensions/roc
--- a/extensions.toml
+++ b/extensions.toml
@@ -1017,7 +1017,7 @@ version = "0.1.0"
 
 [erlang]
 submodule = "extensions/erlang"
-version = "0.1.3"
+version = "0.2.0"
 
 [everforest]
 submodule = "extensions/everforest"
@@ -1851,6 +1851,10 @@ version = "0.6.1"
 [lydia]
 submodule = "extensions/lydia"
 version = "0.0.4"
+
+[mach]
+submodule = "extensions/mach"
+version = "0.1.0"
 
 [macos-classic]
 submodule = "extensions/macos-classic"
@@ -2791,7 +2795,7 @@ version = "0.1.0"
 
 [powershell]
 submodule = "extensions/powershell"
-version = "0.4.0"
+version = "0.4.1"
 
 [prisma]
 submodule = "extensions/prisma"
@@ -2991,6 +2995,10 @@ version = "0.0.5"
 [risor]
 submodule = "extensions/risor"
 version = "0.0.2"
+
+[robot-framework]
+submodule = "extensions/robot-framework"
+version = "0.0.1"
 
 [roc]
 submodule = "extensions/roc"


kylegoetz:
@cla-bot check

kylegoetz:
@cla-bot check

kylegoetz:
> You'll need to ensure the ID here and within your extension repository match, please update it in your repo and then update the submodule here. Also, please remove the Rust code from your extension, currently you won't need this and your extension will still function properly.

I have done this now (assuming I understood the request correctly). Thanks for the guidance.


> Lastly, do we need this very generic `txt` file suffix match or is there perhaps a better suffix we can match on?

Unfortunately, .txt is the extension used by the tree-sitter community to write tests. In all official TS language repositories, they use this, and so all community-created ones (that I am aware of) also use this. 

For example, this is the official Python grammar's corpus tests produced by the Tree Sitter maintainers themselves: https://github.com/tree-sitter/tree-sitter-python/tree/master/test/corpus

Third party tree sitter grammars like the one for Haskell also do this, and so does my own Unison tree sitter grammar.

The only workaround I could imagine is if you could pattern match on the whole path: `test/corpus/**/*.txt`. The test command used for testing grammars is `tree-sitter test` and it *only* looks for files in `$language_root/test/corpus` (and its subdirectories).

kylegoetz:
Also, for what it's worth, the Zed manual on developing extensions has, as its example setup, an id and repo that do not match name.

Screenshot 2026-02-18 at 3 01 30 PM

Should I open a ticket for this? Or am I not understanding which "id" is meant to match the repo name? Thanks!

MrSubidubi:
> Unfortunately, .txt is the extension used by the tree-sitter community to write tests. In all official TS language repositories, they use this, and so all community-created ones (that I am aware of) also use this.

Appreciate the context!

> The only workaround I could imagine is if you could pattern match on the whole path: test/corpus/**/*.txt. The test command used for testing grammars is tree-sitter test and it only looks for files in $language_root/test/corpus (and its subdirectories). 

Yeah unfortunately that is not supported, although I have it on my list to change (currently sick, hence lagging very much behind on things I wanna do, but I made a note of it). For the time being, could we pan to a similar approach as https://github.com/dargor/zed-python-requirements?tab=readme-ov-file#configuration where we just hint in the README? I definitely will not say its optimal, but I am not sure whether generally matching txt files would be better here :/

> Should I open a ticket for this? Or am I not understanding which "id" is meant to match the repo name? Thanks!

Think you are misunderstanding this - your ID does not have to match your repo name (it can, but it does not have to). The ID is what we assign here and what you'll be able to find your extension under within Zed and our webpage (e.g. https://zed.dev/extensions/java).

So, for this case, feel free to just change it to `tree-sitter-test` in your repo. 

We'll gladly accept docs improvements here though! So if you think the wording is bad, feel free to shoot us a PR and I'll take a look 🙂

kylegoetz:
> For the time being, could we pan to a similar approach as https://github.com/dargor/zed-python-requirements?tab=readme-ov-file#configuration where we just hint in the README? I definitely will not say its optimal, but I am not sure whether generally matching txt files would be better here :/

Does this mean I would put in my own extension's README file a code block

```
"file_types": {
    "Tree Sitter Corpus Tests": ["**/tests/corpus/**/*.txt"],
  },
```

and instruct the user to insert this into their Zed's settings.json? Am I meant to leave the `path_suffixes = ["txt"]` in my language's `config.toml`?

I'm unclear how these settings interact. In any case, I've tried variations of the above ("./tests/corpus/**/*.txt", "tests/corpus/**/*.txt", "tests/**/*.txt", etc.) in my Zed settings.json and the syntax highlighting doesn't work automatically for any of my tests.

While waiting for a response, I'm going to force push the latest updates to this PR to include new tasks.

MrSubidubi:
> Does this mean I would put in my own extension's README file a code block

Yeah, that would be the idea. 

> Am I meant to leave the path_suffixes = ["txt"] in my language's config.toml?

Nope, that would be a replacement for that.

> I'm unclear how these settings interact.

User settings always take precedence and I believe glob support was only added after extension `file_suffix` support sadly. Makes very little sense to me, sadly is the way it is. 

> In any case, I've tried variations of the above ("./tests/corpus//*.txt", "tests/corpus//.txt", "tests/**/.txt", etc.) in my Zed settings.json and the syntax highlighting doesn't work automatically for any of my tests.

That is a real issue and definitely something I'd like to see resolved before I push you in that direction. Did you get it to work in the meantime once? https://github.com/zed-industries/zed/blob/943cd106c9a2f50d8c781f15f780ff2be7a6b894/.zed/settings.json#L50-L55 would hint to me you did everything right, hence a bit surprised to not see it working honestly :(

kylegoetz:
@MrSubidubi Here is my `$PROJECT/.zed/settings.json` file in its entirety:

```
// Folder-specific settings
//
// For a full list of overridable settings, and general information on folder-specific settings,
// see the documentation: https://zed.dev/docs/configuring-zed#settings-files
{
  "file_types": {
    "Tree Sitter Corpus Tests": ["./test/corpus/abilities.txt"],
  },
}

```

Then when I open `test/corpus/abilities.txt` there is no highlighting. It's just black text on white background.

Here is my ~/config/zed/settings.json:

```
// Zed settings
//
// For information on how to configure Zed, see the Zed
// documentation: https://zed.dev/docs/configuring-zed
//
// To see all of Zed's default settings without changing your
// custom settings, run `zed: open default settings` from the
// command palette
{
  "git_panel": {
    "tree_view": true,
  },
  "languages": {
    "Unison": {
      "hard_tabs": true,
    },
  },
  "git": {
    "inline_blame": {
      "min_column": 0,
      "show_commit_summary": true,
    },
  },
  "ui_font_size": 16,
  "buffer_font_size": 16,
  "theme": {
    "mode": "system",
    "light": "One Light",
    "dark": "One Dark",
  },
  "prettier": {
    "tabWidth": 2,
    "semi": false,
    "singleQuote": true,
    "arrowParens": "avoid",
  },
  "file_types": {
    "Tree Sitter Corpus Tests": ["test/**/*.txt"],
  },
}

```

No matter which TXT file I open in `test/corpus` there is no highlighting.

In my extension's codebase, `languages/tree-sitter-corpus-tests/config.toml`, 

`name = "Tree Sitter Corpus Tests"`

So I have the key of the KVP under file_types correct.

Also has no effect.

MrSubidubi:
```json
  "file_types": {
      "Tree Sitter Corpus Tests": ["**/test/corpus/**/*.txt"],
    },
```
seems to make things work for me - could you check?

If it does, can you please make the updates and then also ensure the language follows the naming convention of `Tree-sitter`?

MrSubidubi:
Hey, I'll close this one due to inactivity for now, but I very much would like to get this in. I am very happy to reopen this, please just ping me once you decide to pick this up again. Thanks and have a nice week!

PR #4855, opened 2026-02-14
Update ferret to v0.0.8

itsfuad:


itsfuad:
I wish there were auto merge. Waiting for extensions to be available is frustrating

itsfuad:
I could not anyhow add icon to the file extension from the language extension, So I added icon theme. 
image

It's just a copy of material icon theme with only added "ferret"'s icon. That is also the reason why I did not create it as separate extension. It has nothing new beside ferret's icons. 

I can remove the icon theme if you want. Or if somehow I can provide fallback icons to the file type, please let me know.

itsfuad:
@MrSubidubi Done. I removed icon files from this extensions and opened a pull request to materials icon repo to add `Ferret` icons.
Thanks for the guidance.

MrSubidubi (changes requested):
Thank you. Last request: The LSP will currently not work for any but your machine from what I can tell. Can you make this machine-agnostic please?

I think you will want `Worktree::which` for this to see whether the LSP is available in the environment and return an appropriate error if its not. Then we should finally be good to go, sorry for dragging this a bit. Thanks!

itsfuad:
Ah thanks for the feedback. I'll fix it after a few hours. Thanks for the patience.

itsfuad:
@MrSubidubi Kindly see if anything left now.

MrSubidubi (approved):
Looks good now, thanks!

I _think_ you can remove the setttings for enabling the language server - this is already configurable within Zed itself so you would not necessarily need it in the extension.

However, completely fine for me if you want to remove that at a later point or keep it, up to you.

With that said, good to merge! Do you want to remove that prior or should I hit the green button here?

itsfuad:
You can merge it now. 
I have to update the lsp for more, so I'll remove it on later update. Thanks!

MrSubidubi (approved):
Thanks for bearing with me here! 

And I hope you could also partially see why we do review these, hope that was more helpful than annoying after all.

Have a great weekend!

PR #4871, opened 2026-02-15
Add basilisk-qcc extension bootstrap

VatsalSy:
## Summary
- Register `basilisk-qcc` for first-time publication in the central Zed extensions index
- Point the extension submodule to `VatsalSy/zed-qcc` at the commit that renames the manifest ID to `basilisk-qcc`

## Changes
- Add `extensions/basilisk-qcc` as a new git submodule
- Add `[basilisk-qcc]` entry in `extensions.toml` with `version = \"0.3.0\"`
- Add matching `.gitmodules` entry for the new submodule URL
- Remove the earlier `zed-qcc` bootstrap entry/path from this branch

## Testing
- Not run (metadata/bootstrap change only)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -274,10 +274,6 @@
 	path = extensions/barbenheimer
 	url = https://github.com/jayvicsanantonio/barbenheimer-zed-theme.git
 
-[submodule "extensions/basilisk-qcc"]
-	path = extensions/basilisk-qcc
-	url = https://github.com/VatsalSy/zed-qcc.git
-
 [submodule "extensions/base16"]
 	path = extensions/base16
 	url = https://github.com/bswinnerton/base16-zed.git
@@ -285,6 +281,10 @@
 [submodule "extensions/basher"]
 	path = extensions/basher
 	url = https://github.com/zed-extensions/bash.git
+
+[submodule "extensions/basilisk-qcc"]
+	path = extensions/basilisk-qcc
+	url = https://github.com/VatsalSy/zed-qcc.git
 
 [submodule "extensions/batman"]
 	path = extensions/batman
--- a/extensions.toml
+++ b/extensions.toml
@@ -277,10 +277,6 @@ version = "0.218.1"
 submodule = "extensions/barbenheimer"
 version = "1.3.1"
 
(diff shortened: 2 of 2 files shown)


VatsalSy:
@MrSubidubi thanks for the thoughtful comment. 

1. I have been thinking the same that we `should not download language server packages in the managed Zed node runtime`. 
I’m switching this extension to ship a pinned, bundled qcc-lsp server artifact and launch it directly. That removes runtime package installation and manual LSP wiring; users would only need to configure their `qcc`/Basilisk paths. It is not ready yet though. Perhaps, over this weekend....

2. Now, perhaps, the more important: On scope, Basilisk C (https://basilisk.fr/) is a niche ecosystem. If you prefer, I can keep this as a local/dev extension for now.

VatsalSy:
@MrSubidubi - Re 1. if you have a different suggestion, please do let me know. 

To be fair, the only reason I wrote this up was I had to turn off zed's clang LSP everytime I switched to Basilisk C from native c due to many false positives. And to bring in hover docs for new users (students).

MrSubidubi:
I think 1 sounds fine, so either you upload the server package to NPM or make it available as a downloadble executable somewhere. That is what basically all other language servers do.

Also, totally fine with the niche usecase. That said, do I understand you correctly that the language is basically C? If so, could we not just add this as a C language server and then you disable Clangd in settings? Curious as to whether that'd be an option. https://zed.dev/docs/configuring-languages#choosing-language-servers

VatsalSy:
@MrSubidubi I think I tried that approach in the early days of zed. 
Basilisk C does have a set of extra C-like constructs useful for writing discretisation schemes on Cartesian grids. 

Ok. I am changing this PR into draft stage for now. I will first try to add this add this as a C language server once again perhaps with some help :D .. Thanks for the tip. I appreciate the work (and the IDE as well).

MrSubidubi:
Hey, I'll close this for the time being due to inactivity, but am very happy to reopen once you want to have this published. In the meantime, others can install this as a Dev extension. In any case, feel free to ping me here or open another PR if you intend to publish this again. Thanks!

PR #4882, opened 2026-02-16
Add printf language injection support

ozacod:
Adds printf format string syntax highlighting for C, C++, Go, and other languages.

Before:
before

After:

after

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -3985,3 +3985,6 @@
 [submodule "extensions/zwirn"]
 	path = extensions/zwirn
 	url = https://codeberg.org/polymorphicengine/zwirn-zed-extension.git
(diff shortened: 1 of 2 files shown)


MrSubidubi (commented):
Thanks for this!

I am not sure how helpful this is - as of recent, we have semantic highlighting, which could achieve similar things, could it not? Besides us needing the upstream queries for this obviously

ozacod:
To be able to highlight semantically an injected language, language server of this specific language is needed. Since we don't have one for printf we are bound to tree-sitter. If this PR is merged, we need to update "injections.scm" for the languages(c,c++,go) for correct injection. (To achieve the highlighting on the second image, I have also changed the "injections.scm" of C++.)

PR #4883, opened 2026-02-17
Add witchesbrew-theme

shoenot:
Hi, 
This is a theme I created for Neovim that I've ported over to Zed, since I'm using Zed more now. The Neovim theme has quite a few users so I thought people might be interested in the zed version too.


shoenot:
@cla-bot check

PR #4889, opened 2026-02-17
Add GitGraph extension

dwnmf:
- add GitGraph extension as a new submodule: extensions/gitgraph
- register it in extensions.toml with path crates/gitgraph-zed
- initial version:

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4005,3 +4005,6 @@
 [submodule "extensions/zwirn"]
 	path = extensions/zwirn
 	url = https://codeberg.org/polymorphicengine/zwirn-zed-extension.git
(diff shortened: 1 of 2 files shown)


dwnmf:
@cla-bot check

MrSubidubi (changes requested):

Unfortunately, I am not sure whether this extension misuses the slash command interface - slash commands for agents are intended as just that and nothing else, so I am unsure as to whether we want to merge have this in its current form.

From how I see it, this should rather be done via commands from extensions, which is on our radar, yet now obviosly not supported.

Could you elaborate whether I am missing something here or whether that's the case?

MrSubidubi:
Closing as stale

PR #4766, opened 2026-02-08
Add Salesforce language support

avinkrisv:
Description:
  Adds Rustforce extension providing Salesforce language support:

  - Apex syntax highlighting (.cls, .trigger files)
  - SOQL syntax highlighting (.soql files)
  - SOSL syntax highlighting (.sosl files)
  - Embedded SOQL/SOSL highlighting within Apex code
  - Bracket matching and auto-indentation for Apex

  Uses tree-sitter grammars from [aheber/tree-sitter-sfapex](https://github.com/aheber/tree-sitter-sfapex).

Issue #1239, opened 2024-08-12
Support for APEX language


javiaparicio:
### Check for existing issues

- [X] Completed

### Describe the feature

Add support to APEX language, by Salesforce.

### If applicable, add mockups / screenshots to help present your vision of the feature

APEX documentation: https://developer.salesforce.com/docs/atlas.en-us.apexcode.meta/apexcode/apex_intro_what_is_apex.htm

jbenge1:
https://github.com/jbenge1/zed-apex-extension 

Not sure why I can't open a PR to get this into the repo

javiaparicio:
Thanks for the contribution! As far as I checked, you need to fork the extensions project, add your changes there and create the PR.

https://zed.dev/docs/extensions/developing-extensions#publishing-your-extension

Rihan9:
Is someone still working on this? I know at least 20 peaople that will love to discard VSC and migrate to zed if this extention was available

jbenge1:
I can take a look at this again.

brianevanmiller:
I've already started the process, will try to make some progress on it: https://github.com/brianevanmiller/zed-sfapex

aayushProvusInc:
Did any progress happened? i see last activity is quite long back, do you guys still using Zed for apex?

brianevanmiller:
@aayushProvusInc @Rihan9  - I finally got a working version of the Apex code highlighting (SOQL & SOSL working as well) with version 0.1.0

[The PR was submitted](https://github.com/zed-industries/extensions/pull/4877) today to be an actual Zed extension, but in the meantime, you can download and Install Dev Extension on the main project folder here: https://github.com/brianevanmiller/zed-sfapex

Damecek:
Looks like there are multiple PRs trying to implement this, at different levels of completeness. I would suggest to try collaborate on single one and provide as much of Salesforce DX as possible together. 
Feel free to join this [discord](https://discord.gg/YdvjUGpe) server to collaborate.

Damecek:
@MrSubidubi this can be now closed. 🥳

MrSubidubi:
Good point, thanks for the reminder and efforts @Damecek !

PR #4890, opened 2026-02-17
Add Salesforce language support

Damecek:
## Summary
Add a new Zed extension for Salesforce DX development.

## Extension
- **ID:** `salesforce`
- **Name:** `Salesforce DX`
- **Repository:** https://github.com/Damecek/zed-salesforce-extension
- **Version:** `0.0.2`

## Included MVP functionality
- Salesforce DX language recognition:
  - Apex: `.cls`, `.trigger`, `.apex`
  - SOQL: `.soql`
  - SOSL: `.sosl`
  - Salesforce logs: `.sflog`, `.log`
- Baseline Tree-sitter syntax highlighting for the above languages
- Apex language server wiring via Salesforce Apex Jorje LSP (`apex-jorje-lsp.jar`), launched through Java

## Notes
- Grammar sources are pinned to a specific commit for deterministic builds.
- The extension includes clear runtime expectations for Java/SFDX workspace setup.

## Verification performed
- `cargo check` passes
- `cargo build --release --target wasm32-wasip1` passes
- Metadata includes required fields (including `repository`)

### DIFF
--- a/.github/workflows/actionlint.yml
+++ b/.github/workflows/actionlint.yml
@@ -13,6 +13,7 @@ on:
 jobs:
   actionlint:
     runs-on: namespace-profile-2x4-ubuntu-2404
+    if: github.repository_owner == 'zed-industries'
     env:
       SHA: 03d0035246f3e81f36aed592ffb4bebf33a03106
       VERSION: 1.7.7
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,7 +35,7 @@ jobs:
           node-version: "20"
           cache: "pnpm"
 
-      - uses: actions-rust-lang/setup-rust-toolchain@v1
+      - uses: actions-rust-lang/setup-rust-toolchain@1780873c7b576612439a134613cc4cc74ce5538c # v1
         with:
           toolchain: "1.90"
           target: "wasm32-wasip2"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,7 +35,7 @@ jobs:
           node-version: "20"
           cache: "pnpm"
 
-      - uses: actions-rust-lang/setup-rust-toolchain@150fca883cd4034361b621bd4e6a9d34e5143606 # v1
+      - uses: actions-rust-lang/setup-rust-toolchain@2b1f5e9b395427c92ee4e3331786ca3c37afe2d7 # v1
         with:
           toolchain: "1.90"
           target: "wasm32-wasip2"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,7 @@ jobs:
           clean: false
           fetch-depth: 0
 
-      - uses: pnpm/action-setup@b906affcce14559ad1aafd4ab0e942779e9f58b1 # v4
+      - uses: pnpm/action-setup@fc06bc1257f339d1d5d8b3a19a8cae5388b55320 # v4
         with:
           version: 9
 
@@ -35,7 +35,7 @@ jobs:
           node-version: "20"
           cache: "pnpm"
 
-      - uses: actions-rust-lang/setup-rust-toolchain@a0b538fa0b742a6aa35d6e2c169b4bd06d225a98 # v1
+      - uses: actions-rust-lang/setup-rust-toolchain@150fca883cd4034361b621bd4e6a9d34e5143606 # v1
         with:
           toolchain: "1.90"
           target: "wasm32-wasip2"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ concurrency:
   cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
 
 env:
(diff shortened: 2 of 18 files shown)


Damecek:
addresses https://github.com/zed-industries/extensions/issues/1239

MrSubidubi (changes requested):
Will block this for the moment on https://github.com/zed-industries/extensions/pull/4766#pullrequestreview-3842097032, hope that's fine with you. Thank you for raising your and the other PR on the other PRs so that the backlog is a bit smaller 🙂 

In any case, if we are to proceed with this PR, please remove the Python wrapper, I think this should no longer be needed as of https://github.com/zed-industries/zed/pull/48587

Damecek:
Fine by me. Let's see how it will end with the other PR. 

Definitely have zed-industries/zed#48587 on radar. Just waiting once it gets out of pre-release there will be no need for the proxy.

Damecek:
@MrSubidubi This is now only salesforce extension with open PR. Do we still want to wait for zed-industries/zed#48587? It is being postponed over and over again. Now it is stable and once that will be released I will promptly remove the Python proxy.

MrSubidubi:
Sorry for the slight delay here, should we not by now be able to remove the proxy without issues?

Also, how would you feel about changing the extension ID to just `salesforce`?

Damecek:
The fix which is now mitigated by the python proxy is still being postponed, now it is scheduled to [v0.227.1-pre](https://github.com/zed-industries/zed/commit/641c58d04ced013e1cdd20bc97edae84bc9c4bf1). Once it is in `Latest` release and not `Pre-release` I am comfortable to remove the proxy, until than it simply won't work without it. 

Also I've deliberately selected the name to be `salesforce dx` just in case Salesforce company would want to invest in theirs zed extension, then the obvious choice for them would be name it simple `salesforce`. I just want to avoid any future potential clashes.

MrSubidubi:
> The fix which is now mitigated by the python proxy is still being postponed, now it is scheduled to [v0.227.1-pre](https://github.com/zed-industries/zed/commit/641c58d04ced013e1cdd20bc97edae84bc9c4bf1). Once it is in Latest release and not Pre-release I am comfortable to remove the proxy, until than it simply won't work without it.

It has been in the main branch since v0.224.x at least, which you can see if you click on the three dots next to the current label. Thus, you should be just able to remove said proxy.

> Also I've deliberately selected the name to be salesforce dx just in case Salesforce company would want to invest in theirs zed extension, then the obvious choice for them would be name it simple salesforce. I just want to avoid any future potential clashes.

Are there any hints in that regard? As for now, I'd totally be fine with you getting that ID, but if you are against that, not insisting.

Damecek:
> > The fix which is now mitigated by the python proxy is still being postponed, now it is scheduled to [v0.227.1-pre](https://github.com/zed-industries/zed/commit/641c58d04ced013e1cdd20bc97edae84bc9c4bf1). Once it is in Latest release and not Pre-release I am comfortable to remove the proxy, until than it simply won't work without it.
> 
> It has been in the main branch since v0.224.x at least, which you can see if you click on the three dots next to the current label. Thus, you should be just able to remove said proxy.

If I understand the release lifecycle correctly, it is still not propagated to the end users on the latests release, as the changes are tagged to be part of pre-release. The fix was just postponed to v0228.0-pre.

Please correct me if I am wrong.

MrSubidubi:
You are indeed wrong. The fix has already been rolled out to end users weeks ago in v0.224.x. You cal also see this within https://zed.dev/releases/stable/0.224.4

Damecek:
I see, thanks for clarification. 

Will remove the proxy promptly so we can merge this.

Damecek:
@MrSubidubi proxy is gone.

MrSubidubi (changes requested):
Thanks for making the change! 

I do have to note that https://github.com/Damecek/zed-salesforce-extension/blob/260ac97f24fb95981de733aff4ff4b0b1719c460/src/lib.rs#L66 is neither what we want nor will it work - the `vendor` directory will not be copied into the extension directory. Can we resolve this differently? 

Also, I'd still like to discuss the topic of 
> > Also I've deliberately selected the name to be salesforce dx just in case Salesforce company would want to invest in theirs zed extension, then the obvious choice for them would be name it simple salesforce. I just want to avoid any future potential clashes.
> 
> Are there any hints in that regard? As for now, I'd totally be fine with you getting that ID, but if you are against that, not insisting.

Damecek:
@MrSubidubi both resolved. renamed to salesforce and removed the local jar, implemented download with cache.

Damecek:
@MrSubidubi is there anything I can improve in order for this to be merged? Thanks for your time

MrSubidubi (changes requested):
Think we are close to merging here, sorry for the long delay, got pulled into an urgent project.

Lastly, could you please remove `find_in_shell_path` and all call sites? This will only lead to cryptic errors and you should only rely on `worktree.which` for this, which will resolve a proper name should one exist.

Damecek:
@MrSubidubi updated per your request. thanks 🚀

MrSubidubi (approved):
Alright, lets go - thank you for bearing with me on this and the extension! 🚢

PR #4909, opened 2026-02-19
Add MySQL MCP server

toonvd:
This extension uses https://github.com/benborla/mcp-server-mysql and adds built-in SSH tunnel functionality.
Use with caution.


toonvd:
@cla-bot check

toonvd:
Weird, I just did. @cla-bot check

toonvd:
You are correct, I rushed into the situation I needed myself at that moment. In this case, the upstream MCP does not support SSH tunnels (they claim they do but then provide you with a bashscript to set it up). I will remove the option and provide instruction on how to set up an SSH tunnel if needed.

MrSubidubi:
Sounds good!

toonvd:
Changes have been made and tested :)

MrSubidubi (approved):
Nice, that looks better, thank you for the quick follow-up!

Also, https://github.com/toonvd/zed-mysql-mcp/blob/16ccf02c9577a239015390c73b4da4d9bb991610/configuration/installation_instructions.md#requirements the node point is not strictly true I think, as we ship a node runtime with Zed, right? Might be wrong here.

Either way, don't consider this to be a blocking issue and something for a follow-up, hence will go ahead and get this in. Thank you!

toonvd:
Correct, seems my ai agent inserted some weird extras in the README  :| , I didn't review them well. I fixed it already so it can join a possible next release. Thanks for the merge!

PR #4913, opened 2026-02-20
Add jinja-universal

Else00:
Adds `jinja-universal` to the Zed extension registry at version `1.0.0`.

## Why this extension

`jinja-universal` provides Jinja2 syntax highlighting with host-language injection across a large set of Zed-supported languages (currently 353 in the published default set), including examples like:

- `.yaml.jinja`, `.toml.jinja`, `.sql.jinja`, `.md.jinja`
- filename-based templates like `Justfile.jinja`

## How it differs from existing Jinja extensions

This is complementary to the existing Jinja entries:

- `html-jinja` (`0.1.0`): HTML-first approach (`grammar = "html"`), focused on HTML/Jinja templates.
- `jinja2` (`0.0.1`): single Jinja2 language config with generic `.jinja/.jinja2/.j2` detection and HTML-oriented injection.

`jinja-universal` instead ships generated per-language Jinja variants, so host-language highlighting is preserved across many non-HTML template targets as well.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1606,6 +1606,10 @@
 	path = extensions/jetbrains-themes
 	url = https://github.com/artemevsevev/zed-theme-jetbrains.git
 
+[submodule "extensions/jinja-universal"]
+	path = extensions/jinja-universal
+	url = https://github.com/Else00/zed-jinja-universal.git
+
 [submodule "extensions/jinja2"]
 	path = extensions/jinja2
 	url = https://github.com/ArcherHume/jinja2-support.git
--- a/extensions.toml
+++ b/extensions.toml
@@ -1635,6 +1635,10 @@ version = "0.0.2"
 submodule = "extensions/jetbrains-themes"
 version = "0.3.1"
 
+[jinja-universal]
+submodule = "extensions/jinja-universal"
+version = "1.0.0"
+
 [jinja2]
 submodule = "extensions/jinja2"
 version = "0.0.1"


Else00:
@MrSubidubi Hi, this PR is now green (CI passes) and ready for review. Could you please take a look when you have time? Thanks!

MrSubidubi (commented):
Well, that is a _lot_ of languages for one extension, although I do see at the end it reuses the grammar everywhere as described by your readme as well. 

Can I ask for some more input here and perhaps how this is done in VSCode? I see the existing extensions only support HTML and this obviously supports more, still curious though as to how this is handled for other editors (and what we can do to make this more minimalistic perhaps). Thanks!

Else00:
Thanks for raising this, very fair point.

You’re right that this PR includes a lot of language variants.  
The key implementation detail is that it still uses a single Jinja2 grammar; the larger count is mostly file-detection + host-language injection metadata, not many separate parsers.

On packaging strategy: I explicitly considered multiple distribution profiles before opening this PR:
- a minimal profile (native-only),
- a standard profile (native + extension),
- a full profile (including extra/manual languages).

I chose the standard profile for the registry submission because it gives broad real-world coverage while staying deterministic:
- excludes `extra` languages,
- excludes entries without deterministic auto-detection,
- keeps one shared Jinja grammar across all variants.

I also checked the practical UX tradeoff: the main downside of broader coverage is a larger manual language-picker list. Runtime grammar cost does not scale the same way, and manual picker usage is usually occasional (mostly when auto-detection is missing/incorrect). The picker is searchable, which helps here.

For VSCode comparison: I’m not mirroring a single official "jinja-universal" extension. Most support there is HTML-first, with broader multi-host coverage typically achieved through file associations in community extensions.

If you prefer a narrower first merge, I can switch this PR to the minimal profile and follow up with incremental expansions.

Else00:
@MrSubidubi ping

Else00:
@MrSubidubi Following up on this: would you prefer that I reduce this PR to the minimal/native-only profile first, and then submit broader coverage incrementally in follow-up PRs?

I’m happy to do that if it fits the registry better.

Else00:
@MrSubidubi ping

PR #4920, opened 2026-02-21
Add Dark Castle theme

malucard:
A theme I made for my usage, will update when necessary. Very dark, but colorful and highly readable.


malucard:
@cla-bot check

PR #4923, opened 2026-02-21
Add Isabelle language support

MCB-SMART-BOY:
## Summary
Add a new Isabelle extension entry to the Zed official extensions registry.

## Included changes
- Add submodule: `extensions/isabelle` -> `https://github.com/MCB-SMART-BOY/isabelle-zed.git`
- Add `extensions.toml` entry:
  - `submodule = "extensions/isabelle"`
  - `path = "zed-extension"`
  - `version = "0.2.0"`

## Notes
- Extension repo: https://github.com/MCB-SMART-BOY/isabelle-zed
- Extension id: `isabelle`
- Default runtime mode uses Isabelle native `vscode_server` integration.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1550,6 +1550,10 @@
 	path = extensions/ir-black
 	url = https://github.com/sametaylak/ir-black-zed-theme
 
+[submodule "extensions/isabelle"]
+	path = extensions/isabelle
+	url = https://github.com/MCB-SMART-BOY/isabelle-zed.git
+
 [submodule "extensions/isle"]
 	path = extensions/isle
 	url = https://github.com/eagr/zed-isle.git
--- a/extensions.toml
+++ b/extensions.toml
@@ -1578,6 +1578,11 @@ version = "0.1.2"
 submodule = "extensions/ir-black"
 version = "0.0.1"
 
+[isabelle]
+submodule = "extensions/isabelle"
+path = "zed-extension"
+version = "0.2.0"
+
 [isle]
 submodule = "extensions/isle"
 version = "0.0.1"


MCB-SMART-BOY:
@cla-bot check

MCB-SMART-BOY:
@cla-bot check

MCB-SMART-BOY:
Correction (formatting only):

Updated this PR to address the previous merge/package blockers:

- Synced branch with latest upstream `main` and resolved conflicts
- Kept `.gitmodules` and `extensions.toml` sorted/valid
- Bumped `[isabelle]` entry to `version = "0.2.3"`
- Updated `extensions/isabelle` submodule pointer to `MCB-SMART-BOY/isabelle-zed@d1c875b`

On the extension side, `zed-extension` packaging now succeeds with the current grammar configuration (`tree-sitter-isabelle`), and the registry entry in this PR matches that update.

MCB-SMART-BOY (commented):
ok?

MrSubidubi (changes requested):
Thanks for this and sorry for the delay in review here!

Could you pleaes briefly elaborate on why the reading from the text files within the worktree is needed? I am unfamiliar with how isabelle is used/set up and thus would like to have a brief explanation on why we need to have this in the extension and why e.g. the language server cannot do that themselves.

Thanks!

MCB-SMART-BOY:
@MrSubidubi Thanks for the review.

The extension only reads Isabelles `ROOT` / `ROOTS` files. These are Isabelle project configuration files: they tell Isabelle which session to use and which directories belong to that project.

We need this in the extension because the information is required before starting the server process. In native mode we launch `isabelle vscode_server` directly, and the relevant values must be passed as startup arguments (`-l` / `-d`) when the process is started. If we wait until after launch, Isabelle may not discover the correct local session/import setup.

So this is only a small startup-time probe for Isabelles standard project metadata, to avoid forcing users to configure these arguments manually for each worktree.

In a proxy/bridge setup this could in principle be handled by the server layer, but in native mode there is no extra wrapper: the extension is starting Isabelles own server directly, so the extension has to determine those arguments first.

I hope this is clear — English is not my strongest language, so I used AI assistance to help condense this reply.

MrSubidubi (changes requested):
Very fine you used AI for translating and I appreciate you doing so. Sorry for the very late reply from my side as well.

Could you please ensure that the command errors if isabelle is not found in the environment? You should not assume the user has it installed. Instead, please provide an appropriate error message that the user can work with.

I think all else is fine overall, if there's anything you can strip out, do not hesistate to do so, but from what I can see here, looks fine overall

MCB-SMART-BOY:
Oh, thanks. I'm preparing for a new isabelle rewrite by Rust. The old one is based on PIDE and it's not suitable for modern editor like zed. Thus I have tried adding an lsp into the new one.
Sorry to waste your time. I'll send new plugins with my new isabelle-rs later.(Might be one year later?)

MrSubidubi:
No worries and no time waste! Am I right to understand you that we can close this then for the time being?

MCB-SMART-BOY:
Sure！



---Original---
From: "Finn ***@***.***&gt;
Date: Tue, Jun 2, 2026 02:42 AM
To: ***@***.***&gt;;
Cc: ***@***.******@***.***&gt;;
Subject: Re: [zed-industries/extensions] Add Isabelle language support (PR #4923)


MrSubidubi left a comment (zed-industries/extensions#4923)
 
No worries and no time waste! Am I right to understand you that we can close this then for the time being?

MrSubidubi:
Alright, will do just that!

Still, I very much appreciate your communication and the effort that went into this - I very much appreciated you being upfront with the use of AI and was one of the most pleasant experiences I had here. Thanks for that and all the best!

MCB-SMART-BOY:
Thanks a lot for the kind words and for taking the time to review this PR.

Closing this for now makes sense. I will revisit the Isabelle extension later with a cleaner implementation, especially around proper detection/error reporting when "isabelle" is not available in the environment.

Thanks again for your patience and helpful feedback!

MrSubidubi:
My pleasure, looking forward to that!

PR #4928, opened 2026-02-22
Add Gruvbox Material Dark theme

GaussianWonder:
This is **yet another port of** [sainnhe's Gruvbox Material](https://github.com/sainnhe/gruvbox-material) VSCode theme.

## Features

- Lighter inlay hints
- [zed-comment](https://github.com/thedadams/zed-comment) support
- closer to the original

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1358,10 +1358,6 @@
 	path = extensions/gruvbox-material
 	url = https://github.com/tokiory/zed-gruvbox-material.git
 
-[submodule "extensions/gruvbox-material-dark"]
-	path = extensions/gruvbox-material-dark
-	url = https://github.com/GaussianWonder/zed-gruvbox-material.git
-
 [submodule "extensions/gruvbox-material-mix"]
 	path = extensions/gruvbox-material-mix
 	url = https://github.com/hyperdevstuff/zed-gruvbox-material.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1358,6 +1358,10 @@
 	path = extensions/gruvbox-material
 	url = https://github.com/tokiory/zed-gruvbox-material.git
 
+[submodule "extensions/gruvbox-material-dark"]
+	path = extensions/gruvbox-material-dark
+	url = https://github.com/GaussianWonder/zed-gruvbox-material.git
+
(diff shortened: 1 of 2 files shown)


GaussianWonder:
@cla-bot check

MrSubidubi (changes requested):
Thanks for this!

I see it's another port, you say it's another port, so I am a bit curious: Is there a particular reason why you did not try contributing this to a theme upstream/add a variant there with these things added? Otherwise, could you please elaborate why this is not an option? 

Thanks!

GaussianWonder:
All themes have slight variations of coloring and slightly different font style choices for syntax. To me this difference looked like a deliberate choice, which pushed me away from contributing to existing releases.

I also did not consider the option of "adding" a theme as opposed to "overriding".

## Why does this PR exist

[gruvbox_material](https://github.com/tokiory/zed-gruvbox-material)'s colors may match, but its linking to syntax makes it look too aqua, teal or red depending on context.

[gruvbox material dark](https://github.com/hyperdevstuff/zed-gruvbox-material) is the closest, but some syntax element colors were swapped, or changed to other accent colors.
> the dark mix variant only looks more vibrant.

Coming from the VSCode extension, the `Port` variant is the one bearing the most similarity (in terms of pattern matching by syntax+colors).

> Having said all of this, I also did change the `namespace` from _pink old-rose_ to _boney white_ and `attributes` from _pink old-rose_ to _sunlit clay_, so mine is not a purist replica either.

## So what now?

I can do one of two things:

1. undo my changes and expose a literal replica and attach my variation as an override in README
2. try to push this in some other existing repo

> either way there will be a clutter of gruvbox themes, either in the theme selector, or in the extension browser.

What's your opinion on this?

MrSubidubi:
Sorry for the slow response.

> either way there will be a clutter of gruvbox themes, either in the theme selector, or in the extension browser.

Yeah, but I'd argue we don't have to make it worse.

If that works for you, I'd love for you to try improving things upstream, if that does not pan out well, I am happy to merge this here. How does this sound?

MrSubidubi:
Closing for now as stale due to the long queue, feel free to get back and we can discuss this further. Thanks!

PR #4931, opened 2026-02-22
Update tombi to v0.2.1

ya7010:
Reference discussion: https://github.com/tombi-toml/tombi/discussions/1516

- Remove TOML grammer


MrSubidubi (approved):
Yup that seems reasonable. Thank you very much!

PR #4939, opened 2026-02-22
Add firrtl-source-locator language server

MrAMS:
A Zed companion LSP extension for Verilog/SystemVerilog that parses FIRRTL/Chisel source locator comments (`@[...]`) and jumps back to Scala source locations.

## Features

- `textDocument/definition`
  - Jump from anywhere inside one locator comment to all mapped Scala locations.
  - Always returns a multi-target list for one `@[...]` block (for picker-based selection in Zed).
  - Supports inherited-path tokens like `:108:21`.
  - Supports multi-column tokens like `:257:{27,31,48,72}`.
- `textDocument/hover`
  - Shows expanded locator information for shorthand tokens.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1082,6 +1082,10 @@
 	path = extensions/firebase-security-rules
 	url = https://github.com/ChemisTechlabs/zed-firebase-security-rules.git
 
+[submodule "extensions/firrtl-source-locator"]
+	path = extensions/firrtl-source-locator
+	url = https://github.com/MrAMS/zed-firrtl-source-locator.git
+
 [submodule "extensions/fish"]
 	path = extensions/fish
 	url = https://github.com/hasit/zed-fish.git
--- a/extensions.toml
+++ b/extensions.toml
@@ -1096,6 +1096,10 @@ version = "1.0.0"
 submodule = "extensions/firebase-security-rules"
 version = "0.1.1"
 
+[firrtl-source-locator]
+submodule = "extensions/firrtl-source-locator"
+version = "0.1.0"
+
 [fish]
 submodule = "extensions/fish"
 version = "0.0.8"


MrAMS:
All :green_circle: Ready to go!

MrAMS:
Hi @MrSubidubi 

Thanks for the review and the detailed explanation!

I completely understand the standard approach of downloading the server asset. In fact, I used that exact method in a previous extension I contributed [zed-auto-file-header](https://github.com/MrAMS/zed-auto-file-header), where it fetches the LSP server directly from GitHub.

My reasoning for choosing to compile on the host machine for this specific PR is that this LSP is extremely simple and tiny. Maintaining pre-packaged assets for multiple platforms feels a bit tedious and introduces an unnecessary dependency on GitHub releases. Building it directly on the host seems reasonable for its size and naturally provides excellent cross-platform support out of the box. Additionally, I didn't spot a strict rule against host compilation in the official Zed extension development guide, which is why I went with this approach initially.

That being said, I respect your point regarding code auditing and the team's reluctance to endorse on-host compilation. If you insist that packaging and downloading the server asset (or using `$PATH`) is a hard requirement for this to be merged, I am completely willing to update the PR accordingly when I have some free time later.

Let me know how you'd like to proceed!

MrSubidubi:
That is a reasonable point of view, thanks for the context! 

I do still think that having it as a GitHub release, although a bit tedious, has two major advantages:
- It is more evident to the user that we are in fact currently downloading the LSP as opposed to compiling it on their machine. While compilation might be quick, users will not be aware of our conversation here and might be confused what is happening.
- We cannot assume that we can compile this on the host machine - you do handle this in your code, but if users are informed about Rust errors they will not immediately know what is going on and I think it is a much more pleasant experience to not have to do that and instead expect the author (you in this case) to do it for them.

I hope that you can see where I am coming from with this, and thus I'd ask you to just go the normal route here. Thanks for being understanding and your input here, appreciate that nonetheless.

MrAMS:
@MrSubidubi Thanks for such a detailed explanation! I have updated the PR. Plz review at your convenience.

PR #4942, opened 2026-02-22
Add opaline theme

IgorAlexey:
Adapted from [opaline-theme](https://github.com/d4rkr00t/opaline-theme) VSCode Extension, quite good!


IgorAlexey:
My bad! it should be compliant now.

PR #4943, opened 2026-02-23
Add Arduino language support

mambucodev:
## Summary
- Adds Arduino language support for `.ino` files with dedicated tree-sitter-arduino grammar
- Auto-downloads `arduino-language-server` and `arduino-cli` from GitHub releases
- Requires `clangd` on the system PATH (shows install instructions if missing)
- Syntax highlighting, code outline, brackets, indentation, and text objects

## Test plan
- [ ] Install extension and open a `.ino` file
- [ ] Verify status bar shows "Arduino" instead of "C++"
- [ ] Verify syntax highlighting for Arduino builtins (`setup`, `loop`, `digitalWrite`, etc.)
- [ ] Verify LSP starts (requires `clangd` installed)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -150,6 +150,10 @@
 	path = extensions/arctic-depth
 	url = https://github.com/MarvellinusVincent/Arctic-Depth-Zed.git
 
+[submodule "extensions/arduino"]
+	path = extensions/arduino
+	url = https://github.com/mambucodev/zed-arduino.git
+
 [submodule "extensions/ariake"]
 	path = extensions/ariake
 	url = https://github.com/artivilla/zed-ariake-theme
@@ -4021,6 +4025,3 @@
 [submodule "extensions/zwirn"]
 	path = extensions/zwirn
 	url = https://codeberg.org/polymorphicengine/zwirn-zed-extension.git
-[submodule "extensions/arduino"]
-	path = extensions/arduino
-	url = https://github.com/mambucodev/zed-arduino.git
--- a/extensions.toml
+++ b/extensions.toml
@@ -154,7 +154,7 @@ version = "1.0.0"
 
 [arduino]
 submodule = "extensions/arduino"
-version = "0.0.1"
+version = "0.1.0"
 
 [ariake]
 submodule = "extensions/ariake"


mambucodev:
@cla-bot check

mambucodev:
forgot to say: sorry for usage of claude, but it was a big boost in the development, debugging and workflow process to make this extension real. everything is still fully tested by me, only on linux though. I'd need more people to test on different OSes.

mambucodev:
> * I see the auto-installation of the arduino CLI - could you elaborate on that? I think I would prefer here if we were to not do this by default and instead let users configure this using their settings (that means, not download by default, allow download if users set something like `allowCliDownload` under `arduino-language-server.settings` to true).

`arduino-language-server` requires `arduino-cli` as a runtime dependency (passed via `-cli` flag) — it won't work without it. The extension auto-downloads it if not found on PATH. I can gate this behind a setting like `allowCliDownload` if you'd prefer.

> * Similarly, could you elaborate why/how the config yaml is used? In a similar fashion, I'd like to prevent writing a file unprompted or would at least like to know why this would be needed.

The language server requires a `-cli-config` flag pointing to a valid config file or it won't start. The extension creates a minimal default only if one doesn't already exist and the user hasn't specified a custom path in settings. Happy to gate this behind a setting too.

If you still need me to change something specific I can gladly do that, I just thought like that the extension wouldn't be as seamless.

MrSubidubi:
Yeah I do fully agree that the extension would not be as seamless with this, no doubt about that. At some point, I definitely would like for more seamless configuration to be available here.

For me, this always comes down to an issue of how much we want to do implicitly - users already reasonable have an issue with us downloading language servers automatically, and we try to avoid additional downloads where possible. This is an issue created first and foremost by us, at the same time I try to not make the implicitness worse than it already is. 

Somewhat the same holds for the creation of files, while not strictly prohitibet, having a file created without prior notice does not seem like the best solution to me.

Happy to land on a compromise though - how would you feel about having the CLI downloaded by default with opt-out behavior possible whereas we do not create the config in the worktree by default? Or would we then be at a point where it would make more sense to have both these things be opt-in? I would very much prefer for at least the file creation to be gated behind a setting.

As for other things, please run `pmpn sort-extensions` here once. Thanks!

mambucodev:
That compromise works perfectly, I was actually going to propose the same opt-out approach for the CLI myself.

Quick clarification though: the `arduino-cli.yaml` is created in the extension's WASM sandbox directory, not in the user's project worktree. But I'm happy to gate it behind a setting anyway.

I'll add two settings under `arduino-language-server.settings`:
- `autoDownloadCli` (default `true`) — opt-out for CLI auto-download
- `autoCreateConfig` (default `false`) — opt-in for config auto-creation

Both will show clear error messages with instructions when the dependency isn't available. Will also run `pnpm sort-extensions`.

P.S. I'll also indicate the opt-out feature and everything in the extension's repo README of course, let me know if there's anywhere else I should indicate it.

PR #4954, opened 2026-02-23
Update sentry-mcp to v0.1.1

fabric0de:
Release notes:

https://github.com/fabric0de/sentry-mcp-server-zed/releases/tag/v0.1.1


MrSubidubi:
You can also return an error if `env::current_dir` fails - this does never happen in practice and is what all other extensions also do.

fabric0de:
Thanks for the review — agreed.

I updated the lookup to use only `env::current_dir()` and removed all other fallback paths.
`resolve_sentry_entrypoint` now:

1. returns an error immediately if `env::current_dir()` fails, and
2. returns an explicit error if `node_modules/@sentry/mcp-server/dist/index.js` is not found under that cwd.

This aligns with the pattern used in other extensions and keeps the behavior simple and predictable.

PR #4977, opened 2026-02-24
Add VHS80 theme

tahayvr:
New PR for VHS80 with correct id format

https://github.com/tahayvr/vhs80-zed/


tahayvr:
> The submodule here seems borked, all else looks good! Do not forget to run `pnpm sort-extensions` after adding the submodule. Thanks!

Forgot to add .gitmodules 
Thanks

PR #4980, opened 2026-02-24
Re-add Meson language support

jobpaardekooper:
Add an extension to support the Meson build system after the existing Meson zed extension was no longer maintained and removed.

Please view #4532 for the full context of adding this extension.

@MrSubidubi

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2178,6 +2178,10 @@
 	path = extensions/mermaid
 	url = https://github.com/gabeidx/zed-mermaid.git
 
+[submodule "extensions/meson"]
+	path = extensions/meson
+	url = https://github.com/jobpaardekooper/zed-meson.git
+
 [submodule "extensions/metal"]
 	path = extensions/metal
 	url = https://github.com/computer-graphics-tools/metal-analyzer.git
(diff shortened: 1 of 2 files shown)


jobpaardekooper:
Hello again, I have updated the extension with you suggestions above. You can now do something like:

```
{
  "languages": {
    "Meson": {
      "language_servers": ["mesonlsp", "!muon"]
    }
  }
}
```

As far as I am concerned, the extension is ready to be published 😁

I was wondering if there is a way to make `["mesonlsp", "!muon"]` the default setting when using this extension? Because now both will be enabled by default.

Thanks for the suggestion @MrSubidubi

MrSubidubi:
> I was wondering if there is a way to make ["mesonlsp", "!muon"] the default setting when using this extension? Because now both will be enabled by default.

We lack a good way for this to come from the side of extensions currently, but you can shoot us an upstream PR in the Zed repo to change the default settings and I'll take a look!

MrSubidubi:
Could you also update the submodule here to the latest commit? I can otherwise do that tomorrow

jobpaardekooper:
Ok, should be updated now I think

jobpaardekooper:
Thanks for the feedback. I think we can merge this now and I will do a follow-up update.

PR #4983, opened 2026-02-25
Add LikeC4 language support

Lenivvenil:
## Summary

Adds [LikeC4](https://likec4.dev) architecture-as-code language support for Zed:

- **Tree-sitter grammar** — syntax highlighting, bracket matching, auto-indentation, code outline
- **LSP integration** — official [@likec4/language-server](https://www.npmjs.com/package/@likec4/language-server) providing completions, diagnostics, hover, go-to-definition, find references, and rename
- **File detection** — `.likec4` and `.c4` extensions

**Extension repo:** https://github.com/Lenivvenil/zed-likec4
**Grammar repo:** https://github.com/Lenivvenil/tree-sitter-likec4

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4453,3 +4453,6 @@
 [submodule "extensions/zwirn"]
 	path = extensions/zwirn
 	url = https://codeberg.org/polymorphicengine/zwirn-zed-extension.git
+[submodule "extensions/likec4"]
+	path = extensions/likec4
+	url = https://github.com/Lenivvenil/zed-likec4.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1950,6 +1950,10 @@
 	path = extensions/lights-out
 	url = https://github.com/Ajiva-D/lights-out-theme-zed.git
 
+[submodule "extensions/likec4"]
+	path = extensions/likec4
+	url = https://github.com/Lenivvenil/zed-likec4.git
+
 [submodule "extensions/lilypond"]
 	path = extensions/lilypond
 	url = https://github.com/nwhetsell/lilypond-zed-extension
@@ -4453,6 +4457,3 @@
 [submodule "extensions/zwirn"]
 	path = extensions/zwirn
 	url = https://codeberg.org/polymorphicengine/zwirn-zed-extension.git
-[submodule "extensions/likec4"]
-	path = extensions/likec4
-	url = https://github.com/Lenivvenil/zed-likec4.git
--- a/extensions.toml
+++ b/extensions.toml
@@ -1981,6 +1981,10 @@ version = "0.0.1"
 submodule = "extensions/lights-out"
 version = "0.0.1"
 
+[likec4]
+submodule = "extensions/likec4"
+version = "0.0.1"
+
 [lilypond]
 submodule = "extensions/lilypond"
 version = "0.0.9"


Lenivvenil:
@cla-bot check

Lenivvenil:
@cla-bot check

Lenivvenil:
@cla-bot check

Lenivvenil:
@cla-bot check

Lenivvenil:
@cla-bot check

MrSubidubi (changes requested):
Thanks for this!

I very much would like for you to remove https://github.com/Lenivvenil/zed-likec4/blob/c094b7d8d0c213799df843ec1ad58a8c59483385/src/lib.rs#L19-L72 - this should not be needed, as this should be handled by the NPM package install. Can you elaborate on this perhaps?

Lenivvenil:
Thanks for the review!

The ESM resolve hook is unfortunately necessary because Zed ships Node.js v24, which enforces strict ESM resolution. Without the hook, the server crashes immediately on startup:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module
  '.../node_modules/vscode-languageserver/node'
  imported from '.../node_modules/@likec4/language-server/dist/bundled.mjs'
Did you mean to import "vscode-languageserver/node.js"?
```

**Root cause:** `@likec4/language-server` is an ESM package (`"type": "module"`) that imports `vscode-languageserver/node` (without `.js` extension). `vscode-languageserver@9.0.1` is a CJS package with no `exports` field in its `package.json`. Node v24 requires explicit `.js` extensions for subpath imports of such packages — this is an upstream issue in `vscode-languageserver`.

The hook does two things:
1. **Retries failed imports with `.js` suffix** — fixes the `ERR_MODULE_NOT_FOUND` for `vscode-languageserver/node` → `vscode-languageserver/node.js`
2. **Stubs optional peer deps** (`bundle-require`, `esbuild`) — these are statically imported by `@likec4/language-server` but only used for loading `likec4.config.ts` files, not for LSP operations. `zed::npm_install_package` does not install peer dependencies.

I verified this on macOS with the exact Node binary Zed bundles (`node-v24.11.0-darwin-arm64`). The server cannot start without the hook.

Once `vscode-languageserver` ships an `exports` field (or `@likec4/language-server` switches to explicit `.js` imports), the hook can be removed.

MrSubidubi:
Hm, is that not something you could then fix upstream for the language server? While we obviously do not want to ship a broken language server, I still would prefer for this to be fixed at the root as opposed to here in the extension.

Lenivvenil:
You're right — I've filed an upstream issue to address this at the root: https://github.com/likec4/likec4/issues/2840

The situation is actually broader than just the ESM hook: since `@likec4/language-server` v1.48.0, the standalone entry point (`bundled.mjs`) was removed entirely. The bin script now just prints an error and exits. The intended replacement is `likec4 lsp --stdio` from the full `likec4` CLI package, which pulls in 50+ dependencies (vite, playwright, esbuild, etc.) — not viable for a Zed extension.

I've proposed restoring a self-contained bundled entry point upstream. Once that lands, the ESM resolve hook can be removed entirely from this extension.

In the meantime, would you be open to merging with the hook as a temporary workaround? I can add a TODO comment linking to the upstream issue. The hook is minimal and well-scoped — it only intercepts module resolution, doesn't modify module content.

Lenivvenil:
Great news — the upstream issue has been resolved! @davydkov released [`@likec4/lsp`](https://www.npmjs.com/package/@likec4/lsp), a fully-bundled standalone LSP server with zero npm dependencies, specifically to address this.

I've switched the extension to `@likec4/lsp` and **removed the ESM resolve hook entirely**. The integration is now clean — just `node [script] --stdio`, nothing else.

Updated the PR accordingly.

Lenivvenil:
Hey @MrSubidubi, just a gentle ping — the ESM resolve hook has been fully removed as you suggested. The extension now uses `@likec4/lsp`, a standalone zero-dependency LSP package that was released upstream specifically to address this. Would love your feedback when you get a chance!

MrSubidubi (approved):
Looks good now, thank you very much for your efforts and a big sorry from my side for the delay in review!

PR #4989, opened 2026-02-25
Update tombi to v0.2.2

ya7010:
Revert to v0.2.0

See: https://github.com/tombi-toml/tombi/issues/1523

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -3751,7 +3751,7 @@ version = "0.1.0"
 [tombi]
 submodule = "extensions/tombi"
 path = "editors/zed"
-version = "0.2.1"
+version = "0.2.2"
 
 [toml]
 submodule = "extensions/toml"


ya7010:
Thank you.

Yeah, I think it's fine for the Tombi extension to focus on LSP.

This issue is resolved on the Zed side, I will update the Tombi extension.

MrSubidubi:
Would be awesome if we could gather a bit more informatio on the root cause - it should work with the normal TOML extension. Could this be related to missing path suffixes at https://github.com/zed-extensions/toml/blob/main/languages/toml/config.toml or whether it had something to do with the language ID missing for TOML in the `extension.toml`, as e.g. https://github.com/zed-extensions/toml/blob/main/languages/toml/config.toml does? Could you give that a shot if you find the time? I'll also add it as a note to my list. Thanks!

PR #4991, opened 2026-02-25
Add Fe language support

micahscopes:
This adds support for the [Fe language](https://fe-lang.org/getting-started/what-is-fe/)


MrSubidubi (changes requested):
You'll need to add a license upstream, see https://zed.dev/docs/extensions/developing-extensions#extension-license-requirements and also adjust the extension ID so that it matches the ID here. Then, update the submodule here and we should be good to go!

MrSubidubi:
Closing as stale, but happy to take another look with the license issue resolved!

PR #5021, opened 2026-02-27
Update hledger to v0.1.2

juev:


juev:
@MrSubidubi 

I noticed that Zed can now fully handle semantic tokens. As a result, the extension no longer requires the use of tree-siter and its grammars.

Therefore, these issues have been removed from the extension and comments on configuring the extension with coloring have been added.

MrSubidubi:
I still do not think this is a good nor sustainable solution - first and foremost, many people will not have semantic highlighting enabled and thus will not have highlights for that case. Similarly, some people might not even use language servers (or well, for as long as they do not trust the workspace, they will have no highlighting).

Also, many of our other features require some sort of grammar to be present (e.g. indents or outline items). Thus, I'd prefer if you were to add the grammar back so that people who want to use your extension have a pleasant experience no matter their setup. Hope that makes sense

juev:
The main functionality of this extension is to launch and provide LSP server functions.
And coloring and addition and verification, everything is done on the server side.
Previously, I used the trisiter as a fallback for expansion. And that’s only because a little earlier there was no full implementation of the use of semantic tokens from lsp.
Now everything is there and there is no reason to use the previous grammars further.
Moreover, when using both at the same time, conflicts often arise that have to be resolved programmatically. Complicating the code.
Now this is gone.

PR #5027, opened 2026-02-27
Add Xcode MCP bundle extension

tommyming:
This extension contains 2 MCP Servers:

1. XCodeBuildMCP
2. Apple's Xcode MCP Server (available since Xcode 26.3)

Will provide installed display photos when available.

- [x] Test XCodeBuildMCP 
- [ ] Test Xcode MCP Server

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4130,6 +4130,10 @@
 	path = extensions/wxml
 	url = https://github.com/BlockLune/zed-wxml.git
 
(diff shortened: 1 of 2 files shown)


MrSubidubi:
Given that this has been sitting as a draft for quite some time, will close. Also, we generally prefer one MCP per extension, that said, I think in the context it might make sense to have both available. 

Happy to reopen/look at another PR!

PR #5033, opened 2026-02-28
Add wren language server

jossephus:
Add [wren-lsp](https://github.com/jossephus/wren-lsp) extension. 

description:

zed currently has wren extension which provides wren language support with out the lsp experience. wren-lsp is a new lsp for wren, and this extension provides lsp + language support for wren.


jossephus:
@cla-bot check

jossephus:
Hi @MrSubidubi 
Thanks for the review. Yeah I agree and i have removed the language support.

my first intention was that a person should only be required to install a single extension to get the whole lsp + language experience but the duplication part was also bothering me too.

jossephus:
> 
> That said, I consider neither of these blocking. Do you want to fix these as part of this or do you want to throw these into an update soon?

Its better for me if i can update them now. I will update it in few mins. Thanks

jossephus:
@MrSubidubi made both changes in the latest commit.

MrSubidubi:
I made a note for the feature to have recommended extensions (what you mentioned before) and will do my best to keep you posted once I pick this up. Note that it might take some time, but I'll try to push this forward at some point.

PR #5039, opened 2026-02-28
Add please-build extension

michaelmccabe:
This is an extension that provides Language Server Protocol (LSP) support for Please (plz) build files (e.g. BUILD). Please is a cross-language build platform.

See https://please.build

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2902,6 +2902,10 @@
 	path = extensions/playdate
 	url = https://github.com/notpeter/playdate-zed-extension.git
 
+[submodule "extensions/please-build"]
+	path = extensions/please-build
+	url = https://github.com/michaelmccabe/please-build.git
+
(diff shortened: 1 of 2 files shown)


michaelmccabe:
@cla-bot check

MrSubidubi:
Thanks for this and sorry for the delay in review here!

I think this extension in its current state is too barebones to warrant a publish: The language definition itself has no highlights and the requirements to install the language server are suboptimal, we should at least search the environment for the server if not install it ourselves through the extension. Also, the extension does not comply with the 
[extension publishing prerequisites](https://zed.dev/docs/extensions/developing-extensions#extension-publishing-prerequisites), which would something else to consider here.

Happy to look at it another time if the extension has matured a bit and these issues were taken care of. Right now, will close, still, thanks for getting it up in the first place!

PR #5041, opened 2026-02-28
Add Agda language server

MagiHotline:
Adds Agda language extension with LSP support integrated v0.0.1 

Although an Agda extensions for Zed already exists, it comes without the LSP. I decided to integrate it with this PR providing support to the ALS (Agda Language Server).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4277,3 +4277,6 @@
 [submodule "extensions/zwirn"]
 	path = extensions/zwirn
 	url = https://codeberg.org/polymorphicengine/zwirn-zed-extension.git
+[submodule "extensions/agda-lsp"]
+	path = extensions/agda-lsp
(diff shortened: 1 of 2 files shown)


MagiHotline:
I would have liked to open a PR on other [original agda extension for Zed](https://github.com/haohanyang/agda-zed) but it looks like the author is not really interested to develop it further? Just by looking at the issues, he doesn't seem to respond.

MrSubidubi:
Hm, could you try doing that anyway? If they do not respond/merge, we can evaluate whether we want to move ownership of the extension, as we usually prefer for just one languaage support extension to exist. If they do actually not respond/react to a PR, we can look into having you take over the extension so that you can maintain it from then on, where existing users will benefit more from.

MagiHotline:
The original author accepted my PR to the Agda extension, so I am closing this one. Thanks.

PR #5074, opened 2026-03-03
Add i18n-key-translator language server

zhaoqi90s:


zhaoqi90s:
@cla-bot check

MrSubidubi (changes requested):
Thanks for this!

In order to have this published, you'll need to at least
- remove the bundled language server and download it from NPM instead - see https://github.com/zed-extensions/astro/blob/main/src/astro.rs for an example of how this might look
- remove the slash command support, as we consider this a misuse of the API. While we do know that we are missing APIs there, using slash commands is not a way we want to nor will officially endorse.


Also, you've got quite a lot of Rust code within your extension, much of which I'd suspect is actually not needed. Please check what you can remove there, as most of this should probably be handled by the language server. Thanks!

MrSubidubi:
Closing as stale

PR #5075, opened 2026-03-03
Add bloc extension

felangel:
Adds the initial 0.1.0 release of the official [bloc](https://bloclibrary.dev) extension.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,7 +35,7 @@ jobs:
           node-version: "20"
           cache: "pnpm"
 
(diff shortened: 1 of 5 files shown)


felangel:
@cla-bot check

felangel:
> Thanks for this!
> 
> This overall looks good, however, I think we should not be using `latest_github_release`, as this will fail rather often and leave people withour your language server.
> 
> How would you feel about using https://docs.rs/zed_extension_api/latest/zed_extension_api/fn.github_release_by_tag_name.html instead? That way, you would have to publish a new extension version with every new release of the language server, however, we would ensure that language server installation actually always succeeds. What do you think?

Yeah good catch, will fix shortly!

felangel:
@MrSubidubi can you take another look? Thanks!

PR #5077, opened 2026-03-03
Add Dogxi theme

dogxii:
Add my zed editor theme


dogxii:
@cla-bot checked

MrSubidubi (changes requested):
You'll need to remove the `-zed` suffix in the extension ID (and the name as well), then update the submodule and everything related here.

PR #5082, opened 2026-03-03
Update arturo to v1.0.1

DaZhi-the-Revelator:
# Update to Arturo extension from 0.5.5 to 1.0.1

### REPL / Interactive Evaluation

Full Zed REPL integration via a native Jupyter kernel (`arturo-kernel`):

- Evaluate cells with `Ctrl+Shift+Enter` (Windows/Linux) or `Cmd+Shift+Enter` (macOS)
- Cell separator: `; %%`
- Stateful sessions — variables and functions defined in one cell are available in subsequent cells
- stdout and stderr streamed directly into the REPL panel
- Interrupt running execution with `Ctrl+C`
- See `arturo-kernel/README.md` for build and install instructions

### Dynamic Signature Generation

A self-updating signature index that replaces the old static list of embedded functions (see `language-server/seed-cache.json` for the full list of seed functions shipped with the extension)

  - **Signature Indexer (`lib/signature-indexer.js`)**: Core `SignatureIndexer` class with full cache lifecycle management
  - **Stale-While-Revalidate Pattern**: Cache is loaded instantly on startup; background task refreshes it every 24 hours without blocking the editor
  - **Offline-First Design**: Ships with a seed cache of the 80 most popular functions so the extension works perfectly without internet from day one
  - **User Function Indexing**: Automatically scans and indexes all user-defined functions across workspace files — their signatures appear in Signature Help, Hover, and Completion alongside the 521+ built-ins, with parameter types extracted from annotations
  - **Delta Updates**: Uses SHA hashes from Arturo's `.nim` source files on GitHub to detect which library files have changed, fetching only the updated signatures rather than doing a full refresh each cycle
  - **Graceful Degradation**: Falls back to the last known good cache if any network fetch fails; falls back to full refresh if delta data is unavailable

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -175,7 +175,7 @@ version = "0.2.0"
 
 [arturo]
 submodule = "extensions/arturo"
-version = "0.5.5"
+version = "1.0.1"
 
 [asciidoc]
 submodule = "extensions/asciidoc"


DaZhi-the-Revelator:
I'm not sure why it is saying extensions.toml is not changed.

MrSubidubi:
You were missing a change here

MrSubidubi (approved):
Scratch that, I got confused by the `lib_embedded` file. Feel free to remove that in the next release. Thanks

PR #5107, opened 2026-03-04
Add fuzzy ruby language server

candidosales:
- Adds support to the new Ruby LSP server ([fuzzy-ruby-server](https://github.com/doompling/fuzzy_ruby_server)) as a Zed plugin

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ concurrency:
   cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
 
 env:
-  ZED_EXTENSION_CLI_SHA: 03d8e9aee95ea6117d75a48bcac2e19241f6e667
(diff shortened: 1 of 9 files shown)


candidosales:
@MrSubidubi could you review it? 🙇

candidosales:
@MrSubidubi, thanks for the quick review. I didn't know that we could add multiple LSP servers per language. Could you show me an example?

I think it would be better to keep them separate so I can add the fuzzy binary in the next version, giving the developer a better experience with this plugin (so he doesn't need to install the fuzzy binary manually to use it).

MrSubidubi:
Hm, I think I'd appreciate if you at least checked in with Vitaly on this over at https://github.com/zed-extensions/ruby. You will also find an example there for multiple language servers within an extension. While we normally do not strictly enforce for all language servers to live within one repository, we have done precisely that for the Ruby extension and the last proposed addition for a language server also went that route. Thank you.

candidosales:
@MrSubidubi I got it, I'll submit a PR there. Thanks for pointing out the direction :)

PR #5112, opened 2026-03-05
Update hledger to v0.1.3

juev:


MrSubidubi (changes requested):
https://github.com/zed-industries/extensions/pull/5021#issuecomment-3984586782 still holds - your initial reply did not really answer any of my concerns. If we want for this to just be the language server, we should just be removing all language support upfront.

PR #5126, opened 2026-03-06
Add Houdini VEX language support

igor-elovikov:
Support for VEX Language used in Houdini software: https://www.sidefx.com/docs/houdini/vex/lang.html
It's a very simple C-like shader language with couple of quirks 

No LSP, only grammar: https://github.com/igor-elovikov/tree-sitter-vex and snippets for builtins (with help provided in tooltips)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1574,6 +1574,10 @@
 	path = extensions/hot-dog-stand
 	url = https://github.com/CoasterFan5/hotdog-stand-zed.git
 
+[submodule "extensions/hou-vex"]
+	path = extensions/hou-vex
+	url = https://github.com/igor-elovikov/hou-vex-zed.git
+
 [submodule "extensions/hp42s"]
 	path = extensions/hp42s
 	url = https://codeberg.org/brechanbech/tree-sitter-hp42s.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1575,7 +1575,7 @@
 	url = https://github.com/CoasterFan5/hotdog-stand-zed.git
 
 [submodule "extensions/hou-vex"]
-	path = extensions/hou-vex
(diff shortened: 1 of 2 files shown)


igor-elovikov:
@cla-bot check

MrSubidubi (changes requested):
Thanks for this and sorry for the way too long wait!

Could you please elaborate why the snippet file is that large - is this supposed to be a replacement for a proper language server?

Furthermore, could we update the language ID to `houdini-vex` to be a bit more descriptive? Thanks!

igor-elovikov:
Thanks! no worries!

sorry, didn't quite get about language id, I've renamed **extension** id and language folder name. Language config has only the human readable `name` (I kept it as "VEX")

for snippets you're right and it mirrors actually how for example vscode extensions for vex authored (just providing snippets for builtins)
It's a bit niche language but it also has a lot of small quirks so you can't actually reuse any existing servers and nobody bothers writing one. VEX has a lot of builtins and it's very useful to autocomplete them with a bit of help. I just parse official specs and create snippets so I can see suggestions and small tooltips:
Screenshot 2026-06-13 at 10 08 24

it does feel a little like exploiting snippets 
if you fell this is not right I can remove it but it is useful and Houdini users already used to that

MrSubidubi (approved):
Lets keep it for now, appreciate the context! And thank you!

.gitmodules
```diff
@@ -1990,6 +1990,10 @@
 	path = extensions/hot-dog-stand
 	url = https://github.com/CoasterFan5/hotdog-stand-zed.git
 
+[submodule "extensions/hou-vex"]
```
MrSubidubi:
```suggestion
[submodule "extensions/houdini-vex"]
```

PR #5128, opened 2026-03-07
Update flexoki theme to v2.0.2

thedadams:


thedadams:
I am looking for a little guidance on the CI failures because (1) this is an update of a existing extension, (2) there is a license in this repo, and (3) I do not own this extension just trying to help get it updated.

MrSubidubi:
The license requirement was only added after that extension was added. To solve this, you will need to softlink or add a separate license into the folder of the extension - we only consider licenses that live in the extension directory. We want to be explicit about this, hence this decision. This will require a change to the upstream repository.

thedadams:
Thanks! I will close this PR and open a new one after I get the license added. It may take a few weeks to get the upstream PR merged.

PR #5140, opened 2026-03-07
Add MJML language support

pataruco:
Adds [MJML](https://mjml.io) email markup language support to Zed.

## Features: 

- Syntax highlighting with MJML-specific tag differentiation (reuses tree-sitter-html)
- Built-in LSP with real-time diagnostics: nesting validation, required attributes, unknown tag detection with typo suggestions, and singleton enforcement
- Bracket matching, auto-indentation, document outline, CSS injection for `<mj-style>` blocks
- Word-aware navigation for hyphenated tag names like `mj-section`

Repository:  https://github.com/pataruco/zed-mjml

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2434,6 +2434,10 @@
 	path = extensions/mistral-vibe
 	url = https://github.com/mistralai/mistral-vibe.git
 
+[submodule "extensions/mjml"]
+	path = extensions/mjml
+	url = https://github.com/pataruco/zed-mjml.git
+
 [submodule "extensions/mlir-tablegen"]
 	path = extensions/mlir-tablegen
 	url = https://github.com/feichai0017/mlir-zed.git
--- a/extensions.toml
+++ b/extensions.toml
@@ -2472,6 +2472,10 @@ submodule = "extensions/mistral-vibe"
 version = "2.7.5"
 path = "distribution/zed"
 
+[mjml]
+submodule = "extensions/mjml"
+version = "0.1.0"
+
 [mlir-tablegen]
 submodule = "extensions/mlir-tablegen"
 version = "0.2.1"


pataruco:
@cla-bot check

MrSubidubi (changes requested):
Thanks for this and sorry for the way too long response time!

Could you elaborate what precisely this adds on top of existing HTML support? We try to only have one definition per language as per our [extension publishing prerequisites](https://zed.dev/docs/extensions/developing-extensions#extension-publishing-prerequisites), thus a bit curious why you went with a separate extension here.

pataruco:
> Thanks for this and sorry for the way too long response time!
> 
> Could you elaborate what precisely this adds on top of existing HTML support? We try to only have one definition per language as per our [extension publishing prerequisites](https://zed.dev/docs/extensions/developing-extensions#extension-publishing-prerequisites), thus a bit curious why you went with a separate extension here.



Thanks @MrSubidubi , and don’t worry about the timing. I’m sure you have a lot of requests to get through.

That’s a good question. In short, this isn’t a second definition of HTML. It’s actually the only definition of MJML, which is a different language ([mjml.io](https://mjml.io/)). MJML uses its own `.mjml` file type, which doesn’t overlap with HTML, and has its own component model. It compiles to responsive email HTML. If you’ve ever worked with email markup, you know how tricky it can be!

The reason MJML can’t just be handled by HTML support is because of the language server. `mjml-lsp` (built on [mrml](https://github.com/jdrouet/mrml)) understands MJML’s components and can report issues that wouldn’t make sense to an HTML language server.

  * For example, there are nesting rules: `mj-text` must be inside `mj-column` or `mj-hero`, and `mj-section` must be inside `mj-body` or` mj-wrapper`.
  * Some attributes are required, like src on `mj-image` or name and href on `mj-font`.
  * It also detects unknown `mj-*` tags and suggests corrections if there’s a typo.
  * The server enforces that there’s only one mj-head and one mj-body in a file.
  * It also checks the overall structure of your MJML files.

I do reuse the `tree-sitter-html` grammar internally, since MJML is syntactically XML/HTML-shaped, which avoids the need to maintain a near-identical grammar, but that’s an implementation detail, not a claim that MJML is HTML. 

Using `.mjml` as another HTML file extension wouldn’t work either, because the HTML language definition can’t provide MJML’s validation or special highlighting.

So this follows the idea of “one definition per language.” This is the main MJML definition.  If you’d like a different structure, I’m happy to make adjustments.

MrSubidubi (approved):
Pristine explanation and I very much appreciate you providing this. Happy to merge, thanks!

PR #5161, opened 2026-03-09
Add silverstripe language support

nicolas-cusan:


nicolas-cusan:
@cla-bot check

MrSubidubi (changes requested):
Thanks for this and sorry for the slow reply!

Overall looks fine, could you elaborate why this needs a distinct language defnition, as this language seems to be PHP at its core?

nicolas-cusan:
Hi @MrSubidubi thank you for reviewing. The Silverstripe CMS has a proprietary templating language. See here: https://docs.silverstripe.org/en/6/developer_guides/templates/syntax/ 

The package only focused on the templating language, not the PHP aspect of things.

MrSubidubi (approved):
Alright, thanks for the context and sorry for another delay!

PR #5164, opened 2026-03-09
Add Vyper extension

heswithme:
Vyper language support with Tree-sitter and vyper-lsp integration.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4126,6 +4126,10 @@
 	path = extensions/vynora
 	url = https://github.com/QuarkPixel/Vynora.git
 
+[submodule "extensions/vyper"]
+	path = extensions/vyper
+	url = https://github.com/heswithme/zed-vyper.git
+
 [submodule "extensions/wakatime"]
 	path = extensions/wakatime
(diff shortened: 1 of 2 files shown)


heswithme:
Please let me know what you think and whether auto-installing vyper-lsp is appropriate, or this should be managed by the user.
Upd: I decided that for simplicity users should manually install vyper-lsp, thus reducing extension surface.

heswithme:
I've extended coverage of tree-sitter in last commits and bumped it's pin in the zed-vyper. Extension is ready for publishing from my side.

heswithme:
added alternative lsp backend and fixed highlighting

MrSubidubi (changes requested):
Thanks for this and sorry for the long delay in review!

Can you please elaborate what you are doing with searching around in the filesystem/why this is happening? Not sure we want this following our [extension publishing prerequisites](https://zed.dev/docs/extensions/developing-extensions#extension-publishing-prerequisites), thus would like some context before I review this in its entirety.

MrSubidubi:

PR #5195, opened 2026-03-12
Update shhhed-theme to v1.5.0

mkhamat:
added light variant


mkhamat:
@MrSubidubi looks like the extension is not updated in the zed

MrSubidubi:
Sigh... I missed this here, but you downgraded your version number for this update, thus, you'd have to pin this. Your next version must be higher than the highest version your extension currently has, that is 1.33.7

mkhamat:
@MrSubidubi i can bump it now in new PR, sorry missed that

PR #5197, opened 2026-03-12
Add hwa extension

yuricommits:


yuricommits:
@cla-bot check

MrSubidubi:
Thanks for this!

In the current state, this is far from being ready to be published - you are missing a lot of steps for users to be able to use the extension: In the current state, users cannot install the language server at all and even if they were able to get that installed, it would still not work with the extension in its current state.

Please take a look at how existing extensions handle this (e.g. the Svelte one, think the HTML too) and feel free  to resubmit once the extension has matured a bit. Thanks!

yuricommits:
> Thanks for this!
> 
> In the current state, this is far from being ready to be published - you are missing a lot of steps for users to be able to use the extension: In the current state, users cannot install the language server at all and even if they were able to get that installed, it would still not work with the extension in its current state.
> 
> Please take a look at how existing extensions handle this (e.g. the Svelte one, think the HTML too) and feel free to resubmit once the extension has matured a bit. Thanks!

Hey, thanks for the feedback and for taking the time to look at it! I'll check out how Svelte and HTML handle the LSP installation side of things and get it into better shape before resubmitting.

PR #5268, opened 2026-03-17
Update design-tokens to v0.2.0

bennypowers:
Updates design-tokens extension to [version 0.2.0](https://github.com/bennypowers/asimonim/releases/tag/v0.2.0).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -924,7 +924,7 @@
 
 [submodule "extensions/design-tokens"]
 	path = extensions/design-tokens
-	url = https://github.com/bennypowers/design-tokens-language-server
+	url = https://github.com/bennypowers/asimonim.git
 
 [submodule "extensions/desktop"]
 	path = extensions/desktop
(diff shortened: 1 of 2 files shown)


MrSubidubi (changes requested):
I am overall fine with this, but I'd like for us to be a bit more considerate with former users - you renamed the language server ID here (which itself is a breaking change, which I am happy to keep if you'd like this), but with https://github.com/bennypowers/asimonim/blob/b3579f8264bb648fed1fb1a23bbd495a5de90030/extensions/zed/src/asimonim.rs#L107 I think we will leave the old language server name around for people, will we not?

MrSubidubi:
That said, happy to merge this if you want to quickly follow-up here

bennypowers:
The id stays the same, only the git repo and binary name change. 0.2.0 is semver breaking in the 0. pre-release range

MrSubidubi (approved):
Nope it did not, you did not change the _extension_ ID, yet the _language server ID_ was changed. 

Anyway, will merge as-is

PR #5306, opened 2026-03-19
Added mako v0.1.0

auwi-nordic:
Adds support for Mako templates with several languages as base language.

- HTML: `.html.mako`, `.htm.mako`
- XML: `.xml.mako`, `.svd.mako`
- CSS: `.css.mako`
- C: `.h.mako`, `.c.mako`
- Text: `.mako`

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2026,6 +2026,10 @@
 	path = extensions/make
 	url = https://github.com/caius/zed-make.git
 
+[submodule "extensions/mako"]
+	path = extensions/mako
+	url = https://github.com/auwi-nordic/zed-mako
+
 [submodule "extensions/malibu"]
 	path = extensions/malibu
 	url = https://github.com/michael-andreuzza/malibu-theme-zed.git
--- a/extensions.toml
+++ b/extensions.toml
@@ -2058,6 +2058,10 @@ version = "0.4.0"
 submodule = "extensions/make"
 version = "1.1.2"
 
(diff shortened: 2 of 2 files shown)


MrSubidubi:
Apologies for the heavy delay in feedback and no response so far!

I am working through the queue currently, recall looking at this once and forgot to give feedback there - sorry for this! This would have had issues in CI (we since enabled this to auto-run, which did not yet happen for you yet), hence I did not immediately review this, which is just straight up my bad - apologies.

We since merged another mako extension whilst working through the queue that did not have these issues - I know this might be frustrating, if you are fine with it, I'd appreciate if you could contribute any improvements there.

Apologies for this, this should not have happened and we are working to improve the process here for future contrubutors. I know this won't help you at all here, still felt like mentioning.


Feel free to reach out if there's something I can help with, I still very much appreciate you taking the time to look into this! Still, apologies for the terrible experience here.

auwi-nordic:
No worries, as long as mako support is added I'm happy

auwi-nordic:
BTW, is there a recommended way of adding support for languages embedded within a template?

This extension added support for some common languages combined with Mako, e.g. HTML+Mako and C+Mako. The other extension doesn't support this.

But it's not exactly ideal to have to add support for all possible languages within a template extension. So I'm not sure if I should then create a dedicated extension for e.g. C+Mako. Any thoughts?

PR #5344, opened 2026-03-22
Add mypy language server

Zenor27:
Add `mypy` typechecker extension for Zed.

- Uses [pygls](https://github.com/openlawlibrary/pygls/) as a simple LSP (since mypy doesn't have one)
- Highlight errors in Zed
- Does not support dmypy (yet)

This is a first version of the extension, and my first extension for Zed.

https://github.com/Zenor27/zed-mypy

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2462,6 +2462,10 @@
 	path = extensions/muted
 	url = https://github.com/arifzeeshan/Muted.git
 
+[submodule "extensions/mypy"]
+	path = extensions/mypy
+	url = https://github.com/Zenor27/zed-mypy.git
+
 [submodule "extensions/naive-ui-intellisense"]
 	path = extensions/naive-ui-intellisense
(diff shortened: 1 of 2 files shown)


MrSubidubi:
Thanks for this!

Unfortunately, I'll have to close this for now - we might accept this at some point, but there are some issues with this as-is. See our 
[extension publishing prerequisites](https://zed.dev/docs/extensions/developing-extensions#extension-publishing-prerequisites) for more context, but to name a few:
- The ID does not match the requirements
- You may not include the Python language server with the extension itself
- You should not panic in the extension code

See e.g. https://github.com/zed-extensions/kotlin (or for that matter, any extension in that org) on how to handle things more gracefully and how to approach this.

That said, feel free to resubmit this with all of this fixed! Just closing this to keep the queue a bit cleaner and since there is quite a lot of work needed to both get the repo as well as this PR in shape. You are free to resubmit this!

PR #4878, opened 2026-02-16
Add cmake extension

NikitolProject:
Adds full CMake language support for Zed.

### Syntax Highlighting
Full syntax highlighting for CMakeLists.txt and .cmake files using [tree-sitter-cmake](https://github.com/uyha/tree-sitter-cmake).

### Language Server
Integrates [neocmakelsp-fast](https://github.com/NikitolProject/neocmakelsp-fast) LSP with:
- Autocompletion for commands, variables, and paths
- Path completions with directory caching
- Hover documentation
- Go-to-definition
- Signature help for CMake commands
- Diagnostics and linting
- Code formatting

The LSP binary is automatically downloaded from GitHub Releases on first use.

### Editor Features
- **Outline panel** — navigate project(), add_executable(), add_library(), functions, macros
- **Auto-indentation** — proper indentation for if/else/endif, foreach, function blocks
- **Bracket matching** — parentheses and command blocks
- **Text objects** — select inside/around functions and macros

### Build Tasks
Run directly from Command Palette or inline on targets:
- Configure (Debug/Release)
- Build All / Build Target
- Run Executable
- Run Test / Run All Tests
- Clean
- Install
- Reconfigure

## Repository
https://github.com/NikitolProject/zed-cmake

## License
MIT

PR #5348, opened 2026-03-23
Update neocmake to v1.0.0

k0tran:
Long awaited CMake extrension update.
Most of the changes was upstreamed from here https://github.com/zed-industries/extensions/pull/4878
Also fixed bug when extension would crash offline.

NikitolProject's neocmake-fast was not integrated tho. It seems that it is small fork which should be upstreamed instead.


BedRockJie:
Mark merge reminder

PR #5355, opened 2026-03-23
Add Phosphor theme

eneko-codes:
Adds **Phosphor**, a neon phosphor theme for Zed with dark and light variants.

- Repository: https://github.com/eneko-codes/phosphor-zed-theme
- License: MIT

Dark
Light

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4417,3 +4417,6 @@
 [submodule "extensions/zwirn"]
 	path = extensions/zwirn
 	url = https://codeberg.org/polymorphicengine/zwirn-zed-extension.git
+[submodule "extensions/synthwave-theme"]
+	path = extensions/synthwave-theme
+	url = https://github.com/eneko-codes/synthwave-theme.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -4449,6 +4449,6 @@
 [submodule "extensions/zwirn"]
 	path = extensions/zwirn
 	url = https://codeberg.org/polymorphicengine/zwirn-zed-extension.git
-[submodule "extensions/synthwave-theme"]
-	path = extensions/synthwave-theme
-	url = https://github.com/eneko-codes/synthwave-theme.git
+[submodule "extensions/phosphor-zed-theme"]
+	path = extensions/phosphor-zed-theme
+	url = https://github.com/eneko-codes/phosphor-zed-theme.git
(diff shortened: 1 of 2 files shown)


MrSubidubi (changes requested):
Thanks for this!

It seems that for your theme, we already have existing ports over at https://zed.dev/extensions?filter=themes&query=synth. Could you please elaborate on why we should publish your theme as a separate extension to the already existing themes and explain where your theme differs from the existing themes/why your changes do not make sense to contribute upstream? Thanks!

eneko-codes:
Thanks for the review, and fair question.

Since opening this PR I've renamed and reframed the theme as **Phosphor** (`phosphor-zed-theme`) — the new identity is closer to phosphor CRT / neon-on-deep-purple than to the Synthwave-84 lineage that the existing themes derive from. I've just pushed an update to this PR with the new submodule, the renamed `[phosphor-zed-theme]` key in `extensions.toml` (moved to its alphabetical position), and updated title/body.

Concrete differences from the three existing themes (`synthwave`, `synthwave-alpha-theme`, `synthwave84`):

- **Light variant.** Phosphor ships *Phosphor Pastel Light* alongside the dark theme. All three existing themes are dark-only.
- **Original palette, not a Synthwave-84 port.** The dark variant uses a deep purple `#1c182c` background with a cyan-forward function color (`#00fff5`), hot-pink keywords (`#ff2d95`), and yellow strings (`#fffe6e`). Synthwave Alpha, for comparison, is purple-dominant with `#aa54f9` as its main accent on a `#241b30` background. The accent strategy and contrast feel are distinct, not a re-skin.
- **Different identity / scope.** Phosphor isn't trying to be another Synthwave-84 port — it has a companion [phosphor-terminal-theme](https://github.com/eneko-codes/phosphor-terminal-theme) for Terminal.app, Ghostty, and Starship so the palette carries across tools.

For the same reasons, upstreaming wouldn't be a clean fit: the existing themes are faithful to Synthwave-84, and adding a light variant or swapping the accent strategy would change their identity rather than improve them.

(Note: there's an existing `phosphor-icons-theme` in the marketplace — that's an icon theme, different category from this color theme, so there shouldn't be a functional conflict.)

Happy to adjust further if anything still feels off.

MrSubidubi:
Sounds reasonable and thanks for the context!

However, you are missing some important aspects from our 
[extension publishing prerequisites](https://zed.dev/docs/extensions/developing-extensions#extension-publishing-prerequisites) here, notably, the ID is not allowed, and you will also need to sort everything here using `pnpm sort-extensions`. 

With that, I am gonna close this for the time being - please don't let this discourage you, please just open another PR with all that taken care of and I'll gladly take another look. Thanks!

PR #5374, opened 2026-03-24
Update npm-package-json-checker to v1.0.6

e-simpson:
### Fixed
- Settings are now applied via standard LSP configuration flow (`initialize.initializationOptions` and runtime `workspace/didChangeConfiguration`)


e-simpson:
Thank you!

PR #5375, opened 2026-03-25
Add green theme

ningfangbin:
Th theme is mainly in green color.


MrSubidubi (changes requested):
Please fix the issue as outlined by CI, otherwise, we should be good to go!

MrSubidubi (approved):
Thanks! 🟢

PR #5421, opened 2026-03-28
Update smithy to v0.0.2

Nojipiz:
- Bump smithy extension version to 0.0.3 in extensions.toml
- Update smithy submodule to latest commit on main branch.

Biggest change of this is LSP support, related issue on extension repo https://github.com/joshrutkowski/zed-smithy/issues/6

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -3513,7 +3513,7 @@ path = "extensions/zed"
 
 [smithy]
 submodule = "extensions/smithy"
-version = "0.0.2"
+version = "0.0.3"
 
 [sml]
 submodule = "extensions/sml"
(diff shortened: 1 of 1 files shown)


Nojipiz:
@cla-bot check

Nojipiz:
@cla-bot check

MrSubidubi:
You indeed can, however, you'd need to
- remove the rust toolchain file
- remove the grammars directory

Both has to happen upstream before the version bump here will succeed.

If the PRs upstream are not accepted within a reasonable timeframe, we can re-evaluate here whether we want to move the extension into zed-extensions perhaps

MrSubidubi:
Will close for now since this is not really actionable here, but happy to take another look later. Thanks for opening an upstream PR!

Worst case, we can soon discuss on forking the extension into `zed-extensions` so we can update it. Thanks!

PR #5457, opened 2026-03-31
Update objectscript to v1.3.0

hkimura-intersys:


MrSubidubi (changes requested):
Thanks! Is there any particular reason as to why the path suffixes for the objectscript language were removed?

hkimura-intersys:
> Thanks! Is there any particular reason as to why the path suffixes for the objectscript language were removed?

The `objectscript` grammar is the "playground" version of the language, and the `objectscript_udl` grammar is meant for udl files, so I pointed the path from that grammar to be udl instead.

MrSubidubi (approved):
Ah, apologies for missing that in the initial review! Thanks!

Do note that you can make a language invisible for language selection by setting `hidden = false` in the language config.

PR #5484, opened 2026-04-03
Add Popping and Locking Black theme

rgomids:
This PR introduces a new theme: **Popping and Locking Black**, a darker variant of the existing Popping and Locking theme.

## Summary

This theme is based on the original Popping and Locking port and focuses on:

- Pure black background (`#000000`) for reduced glare and better contrast
- Consistent ANSI palette aligned with the original theme
- Improved coverage across UI, terminal, and syntax tokens
- Adjusted contrast for readability in Zed's rendering model

## Key Improvements

- Full UI token coverage (borders, panels, tabs, scrollbars)
- Refined syntax highlighting:
  - clearer distinction between keywords, types, and functions
  - improved readability for comments and doc comments
- Terminal ANSI palette aligned with original theme colors
- Better selection, highlight, and active line visibility

## Design Decisions

- Maintained original color identity where possible
- Adapted tokens where Zed differs from VSCode/TextMate scopes
- Prioritized readability over strict 1:1 port fidelity

## Credits

Based on the original work:
- https://github.com/randoneering/popping-and-locking-zed-theme

## Screenshots

screen1

screen1

## Notes

- Tested on multiple file types (Go, JSON, Markdown)
- No breaking changes to existing themes
- Follows current Zed theme schema (`v0.2.0`)


MrSubidubi (changes requested):
Thanks for this!

It seems that for your theme, we already have an existing port over at https://zed.dev/extensions/popping-and-locking. Could you please elaborate on why we should publish your theme as a separate extension to the already existing theme and explain where your theme differs from the existing theme/why your changes do not make sense to contribute upstream? Thanks!

rgomids:
@MrSubidubi

Thanks for the feedback!

I’m not trying to replace the existing theme, but introduce a **separate “Black” variant**, similar to what already exists in the VSCode ecosystem:
https://marketplace.visualstudio.com/items?itemName=philsinatra.popping-and-locking-vscode-black

## Why a separate extension?

This theme is intentionally different, not an incremental improvement:

- Uses a **true black (`#000000`) background** across the UI
- Adjusts contrast and foreground tones (less washed-out look)
- Tunes terminal colors to be more vivid
- Normalizes some syntax choices (e.g. property keys in neutral white)

## Why not upstream?

These changes are **opinionated and change the visual identity** of the theme. Merging upstream would alter expectations for users of the existing port.

## Conclusion

This is a **parallel variant**, aligned with how the theme is distributed in VSCode, targeting users who prefer a pure black, higher-contrast experience.

rgomids:
@MrSubidubi 

Just to confirm - are you suggesting that instead of publishing this as a separate extension, this “Black” variant should be included as an additional theme option within the existing Popping and Locking extension (e.g. as a secondary theme in the same package)?

If so, I’m happy to adapt the contribution in that direction.

MrSubidubi:
> Just to confirm - are you suggesting that instead of publishing this as a separate extension, this “Black” variant should be included as an additional theme option within the existing Popping and Locking extension (e.g. as a secondary theme in the same package)?
>
> If so, I’m happy to adapt the contribution in that direction.

If that works for both you and the upstream maintainer, yes, I'd prefer that route. That said, should that not be the case, happy to discuss this further here

MrSubidubi:
Will close this for now given no response, but happy to look at a fresh PR should they not want it upstream. Thanks!

PR #5496, opened 2026-04-04
Add todolist

jasonxue1:


jasonxue1:
@cla-bot check

MrSubidubi:
Thanks for this and sorry for the delay in review!

From what I can see, this basically just mirrors the existing comment extension and thus is not a good fit for the registry as per 
[extension publishing prerequisites](https://zed.dev/docs/extensions/developing-extensions#extension-publishing-prerequisites). If you want better captures, you can perhaps discuss fallback captures in the upstream repo.


Feel free to reach out here should you disagree!

PR #5513, opened 2026-04-05
Add Gruvbox Material Theme as an almost 1:1 from VSCode

Mettwasser:


MrSubidubi (changes requested):
Thanks for this!

It seems that for your theme, we already have existing ports over at https://zed.dev/extensions?query=gruvbox. Could you please elaborate on why we should publish your theme as a separate extension to the already existing themes and explain where your theme differs from the existing themes/why your changes do not make sense to contribute upstream? 

Also, please make sure your extension fulfills the [extension publishing prerequisites](https://zed.dev/docs/extensions/developing-extensions#extension-publishing-prerequisites). Once you did the necessary updates, please update everything related here. Thanks!

Mettwasser:
> Thanks for this!
> 
> It seems that for your theme, we already have existing ports over at https://zed.dev/extensions?query=gruvbox. Could you please elaborate on why we should publish your theme as a separate extension to the already existing themes and explain where your theme differs from the existing themes/why your changes do not make sense to contribute upstream? 
> 
> Also, please make sure your extension fulfills the [extension publishing prerequisites](https://zed.dev/docs/extensions/developing-extensions#extension-publishing-prerequisites). Once you did the necessary updates, please update everything related here. Thanks!
> 
> 

Hey, theoretically, yes there are already Gruvbox Material themes, however none of the existing themes actually look like the "original" Gruvbox Material from the VSCode extensions. This theme aims to replicate it as an almost 1:1 (as close as the current styling system in Zed allows) replica.

However, I ran into some issues that make Zed not feel quite as good yet. For that reason I dropped Zed and won't be able to maintain the theme for now. So it's fine to close this. I will open it again once I actually switch over to Zed.

PR #5519, opened 2026-04-05
Add unity-debugger

tomires:
Enables support for debugging Unity projects
https://github.com/tomires/zed-unity-debugger


tomires:
@cla-bot check

MrSubidubi (approved):
Would love to have some context on why the port detection is needed (AFAIK there is nothing like this for other debuggers) but otherwise looks fine, thanks for this and sorry for the long wait!

PR #5554, opened 2026-04-09
Add palenight-dark-theme

stackia:
The PR adds a new theme `palenight-dark-theme`.

https://github.com/stackia/zed-palenight-dark

image


MrSubidubi (changes requested):

PR #5565, opened 2026-04-10
Add Brewfile language support

psg2:
## Summary

- Adds syntax highlighting for **Brewfile** files used by [Homebrew Bundle](https://docs.brew.sh/Brew-Bundle-and-Brewfile)
- Uses a custom tree-sitter grammar at [psg2/tree-sitter-brewfile](https://github.com/psg2/tree-sitter-brewfile)
- Supports all Brewfile keywords: `tap`, `brew`, `cask`, `mas`, `vscode`, `whalebrew`, `go`, `cargo`, `uv`, `krew`, `flatpak`, and `cask_args`
- Highlights strings, comments, symbols, numbers, booleans, and key-value arguments

## Test plan

- [x] Installed as dev extension in Zed — compiles and loads successfully
- [x] Tested against a real Brewfile with 0 parse errors
- [x] Verified syntax highlighting for all supported keywords, strings, and comments

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4453,3 +4453,6 @@
 [submodule "extensions/zwirn"]
 	path = extensions/zwirn
 	url = https://codeberg.org/polymorphicengine/zwirn-zed-extension.git
+[submodule "extensions/brewfile"]
+	path = extensions/brewfile
+	url = https://github.com/psg2/zed-brewfile.git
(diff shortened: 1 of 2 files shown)


psg2:
@cla-bot check

MrSubidubi (changes requested):
Thanks for this!

Is there a particular reason that we could not just have this as a suffix in the Ruby extension? GitHub uses that for highlighting these

psg2:
You're absolutely right, thanks for catching this. I just checked and the Ruby extension already includes `"Brewfile"` in its `path_suffixes` — I didn't realize installing the Ruby extension would cover this. Closing this PR as redundant.

PR #5569, opened 2026-04-10
Add Dexter language server

flowerett:
Adds support for the [Dexter Elixir LSP](https://github.com/remoteoss/dexter) extension for fast go-to-definition for Elixir language.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -946,6 +946,10 @@
 	path = extensions/devicetree
 	url = https://github.com/anikinmd/zed_devicetree
 
+[submodule "extensions/dexter"]
+	path = extensions/dexter
+	url = https://github.com/remoteoss/dexter-zed.git
+
 [submodule "extensions/discord-presence"]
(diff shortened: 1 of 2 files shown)


flowerett:
@cla-bot check

MrSubidubi (changes requested):
Thanks for this!

Quick question - is there a good reason for this to not be added directly to https://github.com/zed-extensions/elixir? While we do not enforce this, we have all Elixir language servers within that extension currently and a very active maintainer, thus the question here.

flowerett:
> Quick question - is there a good reason for this to not be added directly to https://github.com/zed-extensions/elixir? While we do not enforce this, we have all Elixir language servers within that extension currently and a very active maintainer, thus the question here.

Sorry, yes, there’s no need for a standalone extension, let me bundle it into the Elixir one. [PR](https://github.com/zed-extensions/elixir/pull/115)

PR #5570, opened 2026-04-10

arindampradhan:
## Summary

In both case `ctrl + S` syncs the changes in either editor views

- Add Excalidraw Preview extension for viewing .excalidraw files in a native window
- Supports .excalidraw, .excalidraw.svg, and .excalidraw.png formats
- Includes slash command for easy access
- Ships companion binary that downloads on first use

https://github.com/user-attachments/assets/37ba4f47-084f-4270-be0f-4d4096519e1f

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1146,6 +1146,10 @@
 	path = extensions/evolved-theme
 	url = https://github.com/evoL/evolved-theme.git
 
+[submodule "extensions/excalidraw-preview"]
+	path = extensions/excalidraw-preview
+	url = git@github.com:arindampradhan/excalidraw-zed-extension.git
+
 [submodule "extensions/exograph"]
 	path = extensions/exograph
 	url = https://github.com/exograph/zed-extension
--- a/extensions.toml
+++ b/extensions.toml
@@ -1162,6 +1162,11 @@ submodule = "extensions/evolved-theme"
 path = "dist/zed"
 version = "0.3.0"
 
+[excalidraw-preview]
+submodule = "extensions/excalidraw-preview"
+path = "extension"
+version = "0.1.0"
+
 [exograph]
 submodule = "extensions/exograph"
 version = "0.0.2"


arindampradhan:
> cla

I have signed the agreement

arindampradhan:
@cla-bot check

MrSubidubi:

arindampradhan:
> Thank you for this and the effort that went into this!
> 
> Unfortunately, I do think that at the core, this misuses the slash command API and it not something that we want to officially endorse at this time - slash commands for agents are intended as just that and nothing else, so we will not merge this for the time being, sorry to disappoint you here.
> 
> For this to be officially publishable, this will require the API to officially support commands from extensions, which is on our radar, yet now obviously not supported. So we will have to close this for now - thank you for your efforts though anyway.
> 
> That said, you are obviously free to use this personally and/or share with others. This does just mean that we will not share this through the official channel of our extension registry.

@MrSubidubi 

Thank you for the detailed feedback — this makes the underlying concern much clearer.

I understand that the issue is not the Excalidraw integration itself, but that using the slash command API to trigger extension behavior sets a precedent that conflicts with the intended design of slash commands as agent-only primitives.

That makes sense.

What this extension is trying to do is expose a native document preview workflow for .excalidraw files, similar in spirit to how markdown preview works today — the slash command was only used because there is currently no extension API surface to register file-type–driven native views or commands.

Extensions currently have no supported way to register “open this file type in a native companion window / preview surface”.

I would be very interested in adapting this extension to whatever shape the future API for extension-provided commands or views takes. The implementation is already structured around a companion binary and file-type handling, so replacing the slash command with an official mechanism would be straightforward once available.

If helpful, I’m also happy to treat this as an exploratory reference implementation for that API discussion, since this is a concrete example of the kind of capability extensions will want once commands/views are supported.

I’ll keep maintaining this externally in the meantime.

Thanks again for the clarification and for reviewing this.

JaimeJunr:
I'm currently seeing that Zed isn't good with AI internally; it's necessary to use the terminal inside the IDE instead of using what comes with the IDE, plus plugins aren't coming out for the tool...

PR #5603, opened 2026-04-13
Update julia to v0.1.10

piechologist:
Fix missing functions and macros in outline views. 

Fix missing highlighting for certain macros.

Update to tree-sitter-julia v0.25.0. The new grammar is significantly smaller and more performant.

BREAKING: The new grammar drops support for emojis as identifiers. This means that variable or function names etc. cannot contain emojis anymore. Of course, any values can still contain anything.


MrSubidubi (approved):
Thanks!

PR #5611, opened 2026-04-14
Update haxe to v0.4.0

Frixuu:
- Now uses a different Haxe grammar upstream.
- EDIT: Added outline support.
- EDIT: Added injections for the Comment Highlighter extension.
- Bumped required extension API for better Windows support.
- **Runs a Node script** on startup to try to discover a valid build configuration.
  - Removes friction of creating a custom settings file for many projects (default toolchain, non-monorepo).

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -1616,7 +1616,7 @@ version = "0.2.1"
 
 [haxe]
 submodule = "extensions/haxe"
-version = "0.3.1"
+version = "0.4.0"
 
 [hbuilderx-push-light]
 submodule = "extensions/hbuilderx-push-light"


MrSubidubi (changes requested):
Not really the biggest fan of the script you added here, we usually do not allow it following the [extension publishing prerequisites](https://zed.dev/docs/extensions/developing-extensions#extension-publishing-prerequisites). I think I asked this sometime, but is there a compelling reason for having this? Why is the language server not looking for the proper file? Sorry if this is a repeat question, but I tend to forget things after 400 PRs 😅

Frixuu:
Haxe's default "build system", .hxml files, [contain just command line arguments for the compiler](https://haxe.org/manual/compiler-usage-hxml.html). Unlike more typical solutions, they are freeform, can be named anything, and can include other .hxml files. For example, the [repo for the ecosystem's package manager](https://github.com/HaxeFoundation/haxelib) contains [29 of them](https://github.com/search?q=repo%3AHaxeFoundation%2Fhaxelib++language%3AHXML&type=code).

Those list imported libraries, current transpilation target, initialization macros and more, which is why the language server needs one. Currently, the user must specify one manually in `.zed/config.json` for every project they maintain, which is not an amazing experience and gives a bad first impression.

As of writing, the extension API is currently on version 0.7.0. [WorkTree](https://docs.rs/zed_extension_api/0.7.0/zed_extension_api/struct.Worktree.html) can [read a text file](https://docs.rs/zed_extension_api/0.7.0/zed_extension_api/struct.Worktree.html#method.read_text_file), which implies it can detect whether a file exists, but it needs a known path. It can also [return a root path to the workspace](https://docs.rs/zed_extension_api/0.7.0/zed_extension_api/struct.Worktree.html#method.root_path), but **cannot list entries in that path**. `std::fs` APIs are implemented, but they set the virtual root `/` to the extension's working directory `.../zed/extensions/work/*`, which means they [cannot be used to query the project directory](https://github.com/zed-industries/zed/issues/16933#issuecomment-2313089918).

Thus, I've decided to implement this directory scanning in Node, without waiting for upstream API changes. I was inspired by the [Java extension](https://github.com/zed-extensions/java), which [also used to use a Node.js script](https://github.com/zed-extensions/java/pull/71/changes#diff-067a0f194afc1987394b7259e9d78e868dbf8e0dae410128c3fae9df30d9783c) to [work around lacking API](https://github.com/zed-extensions/java/blob/2defa5febc046022131ca595c6eda477aae2475f/src/lsp.rs#L16C1-L21C56), albeit with a different surface.

> Why is the language server not looking for the proper file?

As for why isn't the language server doing this _currently_: it was developed in tandem with a VSCode extension, where this problem was solved [in the extension itself](https://github.com/vshaxe/vshaxe/blob/0081ca15e3a0944ac7c162ee54f7aefdd12cc570/src/vshaxe/HxmlDiscovery.hx).

Please let me know if this reason is compelling enough :smile: Otherwise, I'll be happy to revert this and go forward with just the grammar changes.

Frixuu:
Bump. (The Node.js part is still included; I've just taken the liberty to append some last minute grammar changes.)

MrSubidubi (approved):
Sorry for the delay - you did your research, you used a proper script and you did your best to give me context. Not happy with this totally, but fine with me here and will approve. If we can find a better way forward, would appreciate if you pursued this, but in no way insisting.

In any case, thank you very much for a very thought- and contextful explanation, sincerely appreciate it and is the primary reason I greenlight this here.

PR #5640, opened 2026-04-16
Update fish to v0.1.0

piechologist:
Fix syntax highlighting for command substitutions and options.

Update the tree-sitter-fish grammar.


MrSubidubi (approved):
Thanks!

PR #5649, opened 2026-04-17
Add hl7v2 language support

Yes25:
Adds an extension for [HL7 v2](https://en.wikipedia.org/wiki/HL7) messages (old but still widely used standard in the medical domain).

  - Syntax highlighting via [tree-sitter-hl7v2](https://github.com/Yes25/tree-sitter-hl7v2)
  - Loads the Language server ([hl7_v2_lsp](https://github.com/Yes25/hl7_v2_lsp))

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1630,6 +1630,10 @@
 	path = extensions/hivacruz-theme
 	url = https://github.com/kinoute/zed-hivacruz-theme
 
+[submodule "extensions/hl7v2"]
+	path = extensions/hl7v2
+	url = https://github.com/Yes25/zed-hl7v2-extension.git
+
 [submodule "extensions/hledger"]
 	path = extensions/hledger
 	url = https://github.com/juev/hledger-zed.git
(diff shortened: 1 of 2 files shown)


Yes25:
@cla-bot check

Yes25:
@cla-bot check

MrSubidubi (changes requested):

Overall looks good, however
- please remove the other extension from here (I'll review that PR separately)
- resolve the conflicts
- make sure the ID here and in the repo match  - they currently do not.

Also, please note that we have `github_release_by_tag_name` in the API that you may use, although that is not a requirement here.

Yes25:
Thanks for the feedback. The problems should be resolved now I think:

- removed the accidentally added fhir extension from this PR.
- changed the id to lower case to match the one in the extension repo.
- rebased with current `main` to resolve conflicts.
- code change in the extension repo to make use of the `github_release_by_tag_name`. 

And thanks for the patience. I'm fairly new to this process.

MrSubidubi (approved):
Nice, looks good! Thank you!

PR #5655, opened 2026-04-18
AngelScript

jeruin-dev:
grammar support for AngelScript


MrSubidubi:
Thanks for this! Will close for the time being since the CLA issue was left unanswered for this, happy to repoen once that was taken care of

PR #5674, opened 2026-04-19
Update unison to v0.0.9

zetashift:
EDIT: seems a few more linters have been added since the last time I submitted an update :sweat_smile: 

Updates to 0.0.9 which adds a  license and bumps some versions.


MrSubidubi (approved):
Thanks!

PR #5680, opened 2026-04-21
Update tamarin to v1.12.0

LH104729:
- Using grammar from 'tamarin-prover/tamarin-prover' instead of 'tamarin-prover/vscode-tamarin' since 'tamarin-prover/tamarin-prover' is the source of the grammar while 'tamarin-prover/vscode-tamarin' is just a local copy.
- Version number is now the same as the version of 'tamarin-prover/tamarin-prover'.


MrSubidubi (approved):
Thanks!

PR #5696, opened 2026-04-22
Add CSV Preview extension

Yings520:


Yings520:
@cla-bot check

MrSubidubi:
Thank you for this and the effor that went into this!

Unfortunately, I am gonna close this for the time being - we usually only ship one language support per extension, so you'd probably have to contribute anything that is reasonable to the existing `CSV` extension.

Furthermore, there is ongoing work over at https://github.com/zed-industries/zed/blob/3767d69d6470f028411cfb75d25adbe5548ff9ca/crates/csv_preview/src/csv_preview.rs#L30-L34 👀  So there is something on the near horizon for this, if you are down, feel free to contribute improvements in the Zed repo and give that a try yourself by building yourself and trying the feature flag! I can also add you to that if you'd like me to so that this would already be available for you (although note please that we'll not take bug reports for that feature, but you are free to contribute).

If you want to add the language server regardless, feel free to reopen another PR here with just the language server. Thanks!

PR #5697, opened 2026-04-22
Add GN language support

dsa28s:
### Summary
Adds `zed-gn-language` extension to the Zed extensions registry.

- [x] Registers the extensions/zed-gn-language submodule
- [x] Adds the extension entry to extensions.toml

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4514,6 +4514,10 @@
 	path = extensions/zed
 	url = https://github.com/zed-industries/zed.git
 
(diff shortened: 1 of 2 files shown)


dsa28s:
@cla-bot check

MrSubidubi (approved):
Thanks!

Note that
```
            if zed::current_platform().0 != zed::Os::Windows {
...
```

is not needed,as we noop on Windows already. But that can be fixed in a follow-up

PR #5726, opened 2026-04-24
Add micropython MCP server

DuckyScr:
MicroPython device bridge via MCP; auto-detects USB boards, supports mount-based live execution and push-to-flash; verified against ESP32-C6.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2402,6 +2402,10 @@
 	path = extensions/metal
 	url = https://github.com/computer-graphics-tools/metal-analyzer.git
 
+[submodule "extensions/micropython"]
+	path = extensions/micropython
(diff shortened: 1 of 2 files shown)


DuckyScr:
@cla-bot check

MrSubidubi:
Thank you for this and the effort that went into this, a big sorry from my side for the way too slow response (working on that)!

Unfortunately, I do think that at the core, this misuses the MCP server API and it not something that we want to officially endorse at this time - MCP servers for agents are intended as just that and nothing else, so we will not merge this for the time being, sorry to disappoint you here.

Also happy to discuss ideas on how to add this in a discussion within the Zed repository - do feel free to ping me there.

Thank you!

PR #5733, opened 2026-04-24
Add note language support

ciiqr:
This is the zed port of my note syntax
[vscode extension](https://marketplace.visualstudio.com/items?itemName=ciiqr.note-syntax)
[extension readme](https://github.com/ciiqr/tree-sitter-note/blob/main/zed/readme.md)

sample

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2726,6 +2726,10 @@
 	path = extensions/not-too-shabby
 	url = https://github.com/staleo/zed-nottooshabby-theme.git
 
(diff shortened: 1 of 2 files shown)


MrSubidubi (changes requested):
Thanks for this and sorry for the slow reply!

Can we pleaes change the ID to something like `ciiqr-note` or similar, since this seems to be your personal syntax? I'd like us to indicate this in the extension ID too, please.

PR #5734, opened 2026-04-24
Add nanos world Lua extension

khalilbellil:


MrSubidubi (changes requested):
Thanks for this! 

We are usually pushing towards extensions not providing too much features, although in this case, we could do it. How would you feel about this being two extensions or do you wanna keep it as one? 

Either way, I think we need to discuss the extension ID here as per [extension publishing prerequisites](https://zed.dev/docs/extensions/developing-extensions#extension-publishing-prerequisites)

khalilbellil:
@MrSubidubi
I would prefer keeping it as one extension, because it will be useful for the users. For context this extension is made for gamemodes devs in Nanos World game (a new sandbox game) and it contains a plug and play ready to go tools for them.

For the extension ID I will fix it right now sorry for the mistake.

khalilbellil:
I just realized that my extension id does not actually contains "zed" it's only the repo name. Is it an issue ? @MrSubidubi

khalilbellil:
I renamed the extension folder name so it is consistent with extension id

MrSubidubi:
Sorry for the too long delay here, I think this suffers from a very bad merge. Happy to look at this again if you were to resubmit this, as reviewing this in this PR will be hard due to GitHub and submodules.. Closing with this, feel free to resubmit. Thanks!

PR #5741, opened 2026-04-25
Update ansible to v0.3.1

kartikvashistha:
[Release notes for the core changes](https://github.com/kartikvashistha/zed-ansible/releases/tag/0.3.0).

Full diff between the currently published version and this new version: https://github.com/kartikvashistha/zed-ansible/compare/0.2.0...0.3.1

Many thanks to the following new contributors:
1. @deadnews
2. @JacKeTUs 
3. @Neo2SHYAlien

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -122,7 +122,7 @@ version = "21.0.0"
 
 [ansible]
 submodule = "extensions/ansible"
-version = "0.3.0"
+version = "0.3.1"
 
 [anthracite-theme]
 submodule = "extensions/anthracite-theme"
--- a/extensions.toml
+++ b/extensions.toml
@@ -122,7 +122,7 @@ version = "21.0.0"
 
 [ansible]
 submodule = "extensions/ansible"
-version = "0.2.0"
+version = "0.3.0"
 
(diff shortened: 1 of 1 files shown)


kartikvashistha:
I think the next re-run of the CI should fix the error - I updated the title after pushing my new release, which I'm guessing is causing the problem.

MrSubidubi (changes requested):
```rust
let zed_lsp_settings = LspSettings::for_worktree("ansible-language-server", worktree);
```

is a breaking change and will result in users needing to look into their config - are you aware of this? Happy to merge either way, but wanted to check in with you here first

kartikvashistha:
Hey @MrSubidubi - thanks for flagging this. The only reason why I didn't release this as a major release is because the extension does give a default config to the server that should server most users well from the start. 

On second thoughts, maybe a major release might be better to cover all users rather than some with this change. What is your preference here?

kartikvashistha:
Let me just create a new major release. Will open another PR shortly.

kartikvashistha:
New pr is here, ty: https://github.com/zed-industries/extensions/pull/5883

PR #5755, opened 2026-04-27
Add leander extension

SrGaabriel:
Enhanced Lean 4 support with a more native feel through LSP shenanigans (proxy server)

example

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1974,6 +1974,10 @@
 	path = extensions/lean4
 	url = https://github.com/blackbird314/zed-lean4.git
 
+[submodule "extensions/leander"]
+	path = extensions/leander
+	url = https://github.com/SrGaabriel/leander
+
 [submodule "extensions/leblackque"]
 	path = extensions/leblackque
 	url = https://github.com/gx0r/leblackque.git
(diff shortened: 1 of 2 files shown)


MrSubidubi (changes requested):
Thanks for this!

This seems to violate our [extension publishing prerequisites](https://zed.dev/docs/extensions/developing-extensions#extension-publishing-prerequisites) at first glance, since this also adds Lean4 support - can you elaborate why this is needed? Can we not land improvements upstream instead?

MrSubidubi:

SrGaabriel:
> Closing as stale, feel free to submit another PR should you decide to pick this up again. Thanks!

Hi, sorry for the delay! I’ve been busy with work. Since this PR is already closed, would you prefer that I continue the discussion here or submit a new PR with a clearer explanation? Short explanation just in case:

The main difference is that the existing Lean 4 extension connects directly to Lean’s language server. That works well, but the experience is limited by Zed not currently having an equivalent to Lean’s InfoView, so Leander works around through its own language server that acts as a proxy to the official Lean language server by intercepting information meant for the InfoView and exposes it through features Zed supports (hovers, code lenses and inlay hints for ex). This information is especially important in Lean because without an InfoView or a similar substitute, it is very difficult to keep track of the current goal state, and is a blocker that keeps a lot of people I know in other editors.

I chose not to add this behavior to the existing extension because it's a fundamentally different approach that would probably complicate it, and to keep the orthodox integration separate from this more experimental one.

PR #5760, opened 2026-04-27
Bump YANG to v0.0.2

ronharel02:
Changes: https://github.com/olitez/zed-yang/pull/1


MrSubidubi (approved):
Thanks!

Note that you do not need to gate the `zed::make_file_executable(&binary_path)` - this already noops on Windows. Just an FYI and by no means a blocker here.

PR #5765, opened 2026-04-27
Update tombi to v0.2.4

ya7010:
Updated the logic to fall back to the previous Tombi CLI if downloading the latest version fails.


MrSubidubi (approved):
Thanks!

PR #5768, opened 2026-04-27
Update carbonember theme to v1.1.1

HurrellT:
Fixed a incorrectly named theme (2 themes with the same name)


MrSubidubi (approved):
Thanks!

Also, you can safely remove the `[carbonember]` section at the bottom of your extension.toml

PR #5820, opened 2026-04-30
Add extension for supporting fhir-json messages

Yes25:
This PR adds the fhir-json extension, which provides language support for [FHIR](https://hl7.org/fhir/) (Fast Healthcare Interoperability Resources) JSON files.

FHIR is a modern standard for exchanging healthcare data that gains traction across the world. FHIR resources are commonly represented as JSON, and this extension makes working with those files easier in Zed.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1226,6 +1226,10 @@
 	path = extensions/ferret
 	url = https://github.com/Ferret-Language/ferret-language-support.git
 
+[submodule "extensions/fhir-json"]
+	path = extensions/fhir-json
+	url = https://github.com/Yes25/zed-fhir-extension.git
+
 [submodule "extensions/fiber-snippets"]
 	path = extensions/fiber-snippets
 	url = https://github.com/ayberkgezer/fiber-zed-snippets
--- a/extensions.toml
+++ b/extensions.toml
@@ -1244,6 +1244,10 @@ version = "1.0.0"
 submodule = "extensions/ferret"
 version = "0.0.11"
 
+[fhir-json]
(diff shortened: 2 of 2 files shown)


MrSubidubi (changes requested):
Thanks once more!

I think that this currently violates our [extension publishing prerequisites](https://zed.dev/docs/extensions/developing-extensions#extension-publishing-prerequisites) - can you please elaborate why we need to have a distinct language for this and cannot just ship the language server here?

Yes25:
Thanks for the feedback!

I guess the violation is that this extension doubles functionality by loading the same json-grammar again, that is already shipped in zed? The rational was, that I would like to have custom syntax highlighting that reflects some semantic concepts within fhir. Would an implementation using [Semantic Tokens](https://zed.dev/docs/extensions/languages#syntax-highlighting-with-semantic-tokens) be the correct way to achieve this?

MrSubidubi:
> I guess the violation is that this extension doubles functionality by loading the same json-grammar again, that is already shipped in zed?

Yes, that was my biggest concern, in combination with the issue that this will overwrite JSON language matching in its entirety for all files that end in `JSON`.

Now, I started work on having better matching for file types through the config, but thats still far on the horizon. Hence my question would be: How far can we get with semantic highlighting and/or could we avoid overriding JSON file matching somehow?

MrSubidubi:
Also, another sorry for a slow response, I am fighting the backlog but it certainly does feel a bit like a sisyphus task with the amount that gets in, although I almost get the feeling I might have some control back in a few weeks from now 😅

MrSubidubi:

PR #5845, opened 2026-05-01
Add drogon-csp language support

rbprsp:
Adds a Zed extension for Drogon CSP (.csp), the html + cpp template format used by the [Drogon](https://github.com/drogonframework/drogon) framework.

Tokenization comes from a small tree-sitter grammar; cpp and html are injected into the nested content.

Tags covered: `<%c++ %>, <%inc %>, <%view %>, <%layout %>, [[ ]], {% %}`.
Tested locally as a dev extension.
image

Extension: https://github.com/rbprsp/zed-drogon-csp
Grammar: https://github.com/rbprsp/tree-sitter-csp
License: MIT

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1038,6 +1038,10 @@
 	path = extensions/dream
 	url = https://github.com/arturonegrete-dev/Dream-zed
 
+[submodule "extensions/drogon-csp-language"]
+	path = extensions/drogon-csp-language
+	url = https://github.com/rbprsp/zed-drogon-csp.git
+
 [submodule "extensions/duckyscript"]
 	path = extensions/duckyscript
 	url = https://github.com/Lalolog/duckyscript-zed-extension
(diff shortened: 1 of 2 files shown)


MrSubidubi (changes requested):
Thanks for this, sorry for the late feedback!

Couple of notes:
- Please update your extension name to something more human-readable
- Do we want the `-language` suffix in the ID? Just a question, your call, but would not be needed from our side
- Please update the submodule to point at a commit present on the master-branch, please.

Thank you!

rbprsp:
Thanks for the review!

  - Name is `Drogon CSP` and ID is `drogon-csp` now.
  - Submodule now points at [32b1e2f](https://github.com/rbprsp/zed-drogon-csp/commit/32b1e2f0fa49c3bf5ae1b1eab1176a4b29f69fe4), which is on `master`.

MrSubidubi (approved):
Thanks, sorry for the delay!

PR #5876, opened 2026-05-03
Add Drupal language support

abderrahimghazali:
Adds the [Drupal Syntax Highlighting](https://github.com/abderrahimghazali/zed-drupal-syntax-highlighting) extension, which treats Drupal-specific PHP files as PHP so they get syntax highlighting: `.module`, `.install`, `.theme`, `.inc`, `.profile`, `.engine`, `.test`.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1038,6 +1038,10 @@
 	path = extensions/dream
 	url = https://github.com/arturonegrete-dev/Dream-zed
 
+[submodule "extensions/drupal-syntax-highlighting"]
+	path = extensions/drupal-syntax-highlighting
+	url = https://github.com/abderrahimghazali/zed-drupal-syntax-highlighting.git
+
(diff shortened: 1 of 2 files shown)


abderrahimghazali:
@cla-bot check

MrSubidubi (changes requested):
Thanks for this and sorry for the delayed review!

Few questions:
- Can we rename the ID to just `drupal`? 
- Curious: Could we perhaps just add the path suffixes to the PHP extension instead?

abderrahimghazali:
Yeah, suffixes on the PHP extension is probably the better call. A separate language means those files get highlighting but no Intelephense/Phpactor, since both are bound to PHP. Opened zed-extensions/php#144.

Left `.profile` out of that one, bash already claims it.

Leaving this open until that's sorted either way.

PR #5878, opened 2026-05-03
Add Nord theme

pwc3:


MrSubidubi (changes requested):
Thanks for this!

It seems that for your theme, we already have an existing port. Could you please elaborate on why we should publish your theme as a separate extension to the already existing theme and explain where your theme differs from the existing theme/why your changes do not make sense to contribute upstream? Thanks!

pwc3:
Thanks for the feedback! I did look at `zed-nord-theme` before starting `nord-zed`. The existing extension has been a useful reference, but there are meaningful gaps that drove me to build a new implementation rather than submit patches upstream.

## The core issue: Markdown (and other languages) had no syntax highlighting

The problem I started with — Markdown files showing no syntax highlighting — is caused by missing tokens in `zed-nord-theme`. It lacks `emphasis`, `emphasis.strong`, `link_text`, `link_uri`, and `title` entirely. These aren't minor omissions; they're the tokens Zed uses to render bold, italic, links, and headings in Markdown. `nord-zed` covers all of them.

## Token coverage gap

`nord-zed` defines 45 syntax tokens; `zed-nord-theme` defines 27. Tokens present in `nord-zed` but absent from `zed-nord-theme` include:

- `diff.plus` / `diff.minus` — diff syntax coloring
- `emphasis` / `emphasis.strong` — Markdown italic/bold
- `link_text` / `link_uri` — Markdown and HTML links
- `title` — Markdown headings
- `enum` — enum type coloring
- `namespace` — module/namespace coloring
- `selector` / `selector.pseudo` — CSS selectors
- `embedded` — embedded code blocks
- `predictive` / `hint` / `unreachable` — editor state tokens

## Version control and UI completeness

`zed-nord-theme` leaves all icon styling (`icon`, `icon.muted`, `icon.disabled`, etc.) as `null`, and has no `version_control.conflict_marker.ours/theirs`, no word-level diff colors (`version_control.word_added/word_deleted`), and no `search.active_match_background`. `nord-zed` covers all of these.

## Palette fidelity

`zed-nord-theme` introduces a custom teal (`#6c99a6`) for several UI elements — borders, terminal ANSI cyan, and focus indicators — that doesn't appear anywhere in the official Nord palette. `nord-zed` sticks strictly to the official Nord colors, with one intentional, documented exception (`#616e88` for subdued UI text, to avoid the low contrast of nord3 for inactive elements).

## Why not contribute upstream?

The color choices in `zed-nord-theme` aren't just incomplete — some are semantically different (e.g., constants are `#d8dee9` neutral rather than `#ebcb8b` yellow; `property` is accent cyan rather than neutral; comments are italic, which `nord-zed` deliberately avoids). These aren't bugs to fix; they're design decisions. Merging `nord-zed`'s approach upstream would mean overriding the existing maintainer's choices, which isn't appropriate.

`nord-zed` is a clean-room reimplementation that prioritizes completeness, strict palette adherence, and parity with the Nord VSCode theme. The existing extension serves users who want the light variant or prefer its design choices. Both can coexist.

MrSubidubi (changes requested):
Would it then not be sensible to note this very briefly in the description? With this, we otherwise have two basically identical themes which are from the looks of it undistinguishable from one another. Also, have you asked upstream about this?

PR #5882, opened 2026-05-04
Update nomad to v0.0.3

loczek:
Release Notes: https://github.com/loczek/zed-nomad-extension/releases/tag/v0.0.3

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -2776,7 +2776,7 @@ version = "0.1.1"
 
 [nomad]
 submodule = "extensions/nomad"
-version = "0.0.2"
+version = "0.0.3"
 
 [nord]
 submodule = "extensions/nord"


MrSubidubi (changes requested):
Thanks!

Overall looks fine, can you please give me a quick explanation why all these languages are needed and especially why they are not set to `hidden`?

loczek:
The languages are mainly to provide the [correct language id to the language server](https://github.com/loczek/nomad-ls/pull/16), since it wasn't possible to fit all of the attributes/blocks into one language without conflicts on the language server side e.g.

`nomad-acl`

```hcl
namespace "test" { // namespace is a block
    policy = [ ... ]
}
```

`nomad-variable`

```hcl
path = "..."
namespace = "global" // namespace is a string attribute here

items {
    ...
}
```

As for why the languages aren't hidden, it's because they are not used for injections and they don't have `path_suffixes` assigned because [nomad](https://developer.hashicorp.com/nomad) has no file extension conventions for these languages except the `.nomad` or `.nomad.hcl` suffixes, which are used for the `nomad-job` language. The remaining languages are intended to be configured through file type associations in a `.zed/settings.json` file, so they probably should remain visible in the language picker as well.

MrSubidubi (approved):
Alright, thanks for the context and sorry for the delay!

PR #5906, opened 2026-05-05
Update cfml to v0.2.12

ghedwards:
Release notes:

https://github.com/cfmleditor/zed-cfml/releases/tag/v0.2.12

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -575,7 +575,7 @@ version = "1.0.5"
 
 [cfml]
 submodule = "extensions/cfml"
-version = "0.2.9"
+version = "0.2.12"
 
 [chai-theme]
 submodule = "extensions/chai-theme"


MrSubidubi (changes requested):
Are you sure you want to proceed with regex queries that large? I think this would be better handled in the grammar and similarly should use `anyof` here still. 

Happy to merge if you'd like me to, but wanted to check

ghedwards:
> Are you sure you want to proceed with regex queries that large? I think this would be better handled in the grammar and similarly should use `anyof` here still.
> 
> Happy to merge if you'd like me to, but wanted to check

I think this will do for now, I'll consider that in the future, I very much appreciate the feedback, thank you.

ghedwards:
Not sure if I need to tag you @MrSubidubi

MrSubidubi (approved):
No need to ping, but all good, we try to get back quickly but the queue is curerntly winning sadly. Thanks!

PR #5944, opened 2026-05-07
Add yaml-sql extension

tinwonda:
## Summary

Adds the `yaml-sql` Zed extension to the registry.

- **Extension ID**: `yaml-sql`
- **Repository**: https://github.com/tinwonda/zed-yaml-sql-language-support
- **Version**: `0.0.1`

## What it does

Injects SQL syntax highlighting into YAML values when either:
- The value's key contains `sql` as its own token (e.g. `sql_query`, `query_sql`)
- The value is preceded by a `#language=sql` comment annotation

## Checklist

- [x] Tested locally with "Install Dev Extension" in Zed
- [x] Open-source license included (MIT)
- [x] Extension ID does not contain "zed" or "extension"
- [x] `version` in `extensions.toml` matches `extension.toml`

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4697,3 +4697,6 @@
 [submodule "extensions/zwirn"]
 	path = extensions/zwirn
 	url = https://codeberg.org/polymorphicengine/zwirn-zed-extension.git
+[submodule "extensions/yaml-sql"]
+	path = extensions/yaml-sql
+	url = https://github.com/tinwonda/zed-yaml-sql-language-support
--- a/.gitmodules
+++ b/.gitmodules
@@ -4710,6 +4710,10 @@
 	path = extensions/yaka
 	url = https://github.com/edgarkanyes/yaka.git
 
+[submodule "extensions/yaml-sql"]
+	path = extensions/yaml-sql
+	url = https://github.com/tinwonda/zed-yaml-sql-language-support
+
 [submodule "extensions/yamura"]
 	path = extensions/yamura
 	url = https://github.com/Yxmura/yamura-zed-theme.git
@@ -4841,6 +4845,3 @@
 [submodule "extensions/zwirn"]
 	path = extensions/zwirn
 	url = https://codeberg.org/polymorphicengine/zwirn-zed-extension.git
-[submodule "extensions/yaml-sql"]
-	path = extensions/yaml-sql
-	url = https://github.com/tinwonda/zed-yaml-sql-language-support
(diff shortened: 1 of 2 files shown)


tinwonda:
@cla-bot check

tinwonda:
Hi @MrSubidubi, could you help to review?

MrSubidubi (changes requested):
Sorry for the delay.

At first glance, I think this would violate our [extension publishing prerequisites](https://zed.dev/docs/extensions/developing-extensions#extension-publishing-prerequisites) - can you elaborate what queries this adds and/or why we cannot consider adding these injections upstream?

tinwonda:
> At first glance, I think this would violate our [extension publishing prerequisites](https://zed.dev/docs/extensions/developing-extensions#extension-publishing-prerequisites) - can you elaborate what queries this adds and/or why we cannot consider adding these injections upstream?

@MrSubidubi 
Thanks for the review, and sorry for the slow response.

**What this adds**

The extension adds a single file of substance: `languages/yaml/injections.scm`, which defines two SQL injection patterns:

1. **Comment-driven** — a `# language=sql` annotation explicitly opts a value into SQL highlighting:
   ```yaml
   # language=sql
   query: SELECT * FROM users WHERE id = 1
   ```

2. **Key-name-driven** — keys where `sql` appears as a standalone token (matched by `(^|[-_])sql($|[-_])`) trigger highlighting on the value:
   ```yaml
   sql_query: SELECT * FROM orders
   raw_sql: |
     INSERT INTO logs VALUES (...)
   ```

Both target plain scalars and block scalars. Quoted scalars are intentionally excluded because the YAML grammar exposes quote delimiters as part of the scalar range, which would pollute the injected SQL region.

**Why not upstream**

These are opinionated, domain-specific heuristics that would be wrong for most YAML files. The key-name pattern (`sql_query`, `query_sql`, etc.) is a convention specific to data tooling (dbt, Airflow, SQLMesh). Applying it globally in Zed's built-in YAML support would inject SQL highlighting into unrelated YAML files for all users. The extension system is the right place for opt-in, niche injection behaviors like this.

**Note on the YAML language override**

I acknowledge that the extension overrides the built-in YAML language to add these injections — the `languages/yaml/` files (highlights, brackets, outline, etc.) are carried from Zed's core YAML queries. I've cleaned up the extension since the initial submission: removed an unrelated GitHub Actions JavaScript injection that was copied from core but doesn't belong here, and removed duplicate files. Happy to discuss whether the override approach is acceptable or if there's a lighter-weight path.

tin-trinh (approved):

PR #5951, opened 2026-05-07
Update quadlet to v0.3.0

mufeedali:
Release notes:

https://github.com/mufeedali/zed-quadlet/releases/tag/v0.3.0

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -3386,7 +3386,7 @@ path = "distribution/zed"
 
 [quadlet]
 submodule = "extensions/quadlet"
-version = "0.2.3"
+version = "0.3.0"
 
 [quakec]
 submodule = "extensions/quakec"


MrSubidubi (approved):
Fine for me here, please note that we have an API which can help you with finding the proper asset ovet at https://docs.rs/zed_extension_api/latest/zed_extension_api/fn.github_release_by_tag_name.html

mufeedali:
@MrSubidubi Oh that's cool! Would `latest_github_release()` allow me to auto-update?

MrSubidubi:
Yes it would! You can then compare the released version against the local one to check whether an update is required. Let me know if you need an example on how extensions do that, I think the elixir one does it and plenty others do too

mufeedali:
Nice! Will check it out, should be able to figure it out

PR #5961, opened 2026-05-08
Add jetbrains-dark-blur-theme

Slipstream-Max:
I added a blur effect to the jetbrains-dark-theme for a more comfortable visual experience.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1882,6 +1882,10 @@
 	path = extensions/jetbrains-darcula-theme-by-bronya0
 	url = https://github.com/Bronya0/Jetbrains-Darcula-Zed-Theme.git
 
+[submodule "extensions/jetbrains-dark-blur-theme"]
+	path = extensions/jetbrains-dark-blur-theme
+	url = https://github.com/Slipstream-Max/jetbrains-dark-blur-theme.git
+
 [submodule "extensions/jetbrains-icons"]
 	path = extensions/jetbrains-icons
 	url = https://github.com/ziishaned/zed-jetbrains-icons.git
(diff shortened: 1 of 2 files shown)


Slipstream-Max:
@cla-bot check

MrSubidubi (changes requested):

Slipstream-Max:
Thank you for your reply!
I understand that you already have an existing JetBrains Dark port. Let me explain why I believe my theme should be published as a separate extension rather than merged upstream.

Where my theme differs from the existing ones:

First, the existing themes are solid and don't have any transparency or blur, which makes them look a bit rigid on modern desktops. My theme adds two different transparency levels – a lighter translucency with blur on the editor window and title bar, and a slightly more opaque status bar for better **readability** – to give a glassy look while keeping everything readable. Second, many other blur/transparency themes tend to use really flashy accent colours that **distract** from the code. I stuck with the original JetBrains Dark colour palette, so the code remains the focal point and the transparency is just a subtle enhancement, not a distraction.

Why I am not proposing to merge these changes upstream:

Upstream resistance to blur/transparency – The existing theme is intentionally solid and conservative. Adding blur would be a significant UI change that may not align with the maintainers’ design philosophy. Even if it were accepted, it would likely require a configuration flag, increasing maintenance burden for the core team.

Community precedent for separate blur variants – We already see successful examples like Catppuccin and Catppuccin Blur, which are maintained by different people. This shows that users appreciate having a choice, and that such variants can thrive as independent extensions without harming the original theme’s ecosystem.

Thank you for your time and consideration. I look forward to your thoughts. )

MrSubidubi (approved):
Well, fine with me. Thanks!

PR #5986, opened 2026-05-09
Bump txtar to v0.2.2

FollowTheProcess:
Added a bunch of new injections so more languages get recognised inside txtar blocks and tightened the grammar in the tree-sitter repo so it should be more robust now


FollowTheProcess:
Hmm hang on I'm noticing some weirdness with this occasionally when using it as a dev extension, let me figure out what's going on and I'll re-open with a fix

PR #6000, opened 2026-05-10
feat: add angular-language-server extension v1.0.0

pBouillon:
I work with Angular almost daily and originally developed this to support my own development needs. This project was initially inspired by [nathansbradshaw/zed-angular](https://github.com/nathansbradshaw/zed-angular), but the implementation drifted significantly as I learned my way through the Zed API, Rust, and Tree-sitter. Contributing the changes back would have meant opening a pull request replacing most of the existing codebase, which did not feel like the right thing to do as an external contributor.

Once I had a working version, I noticed the original author mentioned they [no longer work with Angular actively](https://github.com/nathansbradshaw/zed-angular/pull/72#issuecomment-3679636744). This encouraged me to refine this into a formal extension for others to use.

This extension offers:

- Support for Angular-flavored HTML both in standalone files and inline within TypeScript files
- Highlighting and outline support for the latest Angular tokens and control flow syntax (e.g., `@if`, `@for`)

Compared to the existing extension, this version introduces a few functional differences:

- Support for silencing specific Angular diagnostic messages directly via the extension's settings
- Prioritizes existing node packages within the repository to avoid unnecessary re-downloads, looking at a `node_module` in the project root
- Uses the Angular compatibility matrix to resolve the correct version of the Angular Language Service and TypeScript based on the local project environment.

I am aware the contribution guidelines generally discourage duplicate extensions. I hope the differences outlined above make a reasonable case for this one to be listed alongside the existing extension rather than as a replacement for it.

If this is not something the Zed team wishes to include as a separate extension at this time, I completely understand. Regardless of the outcome, thank you for your time and for the work you put into reviewing these contributions.

---

#### Completion in templates

Completion in templates

#### Diagnostics

Diagnostics

#### Hover information

Hover information

#### Go to definition

Go to definition

#### Outline

Outline

#### Syntax highlighting

Syntax highlighting

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4681,3 +4681,6 @@
 [submodule "extensions/zwirn"]
 	path = extensions/zwirn
 	url = https://codeberg.org/polymorphicengine/zwirn-zed-extension.git
+[submodule "extensions/angular-language-server"]
+	path = extensions/angular-language-server
+	url = https://github.com/pBouillon/zed-angular-language-server
--- a/extensions.toml
+++ b/extensions.toml
@@ -124,6 +124,10 @@ version = "0.1.1"
 submodule = "extensions/angular"
 version = "0.0.5"
 
+[angular-language-server]
+submodule = "extensions/angular-language-server"
+version = "1.0.0"
+
 [angular-snippets]
 submodule = "extensions/angular-snippets"
 version = "21.0.0"


pBouillon:
@cla-bot check

pBouillon:
@cla-bot check

MrSubidubi:
Hey, thanks for putting this up and apologies for the slow response time!

I had a look and saw the 

> I am aware the contribution guidelines generally discourage duplicate extensions.

which I really appreciate (since you actually took a look before submitting, so thank you!), yet still think we have to at least narrow the scope here to only include the language server and nothing else, since we still are no fans of having the same language defined twice.

As I see it, nothing would prevent this extension from working properly alongside the other extension, thus curious: Could we not just have the extension provide only the language server and nothing else?

Also, while I took my dear time to reply here, I see that upstream has a new maintainer, so might be worth to consider contributing there after all? Not insisting from our side, but perhaps some of the contributions could be upstreamed even?

Will close for now given the duplicate language definition (and our backlog, which is our issue), but happy to revisit this: Should you still think contributing upstream is unreasonable, we can definitely reconsider this and I'd be happy to take a look at a fresh PR.

Happy to answer any questions you have, in any case, thanks for taking the time to get thi together!

PR #6029, opened 2026-05-11
Bump txtar to v0.3.0

FollowTheProcess:
Tightened up the grammar so it should be more robust now :) also found and fixed a bunch of edge cases that would break highlighting

Had a bit of fun getting here hence the few versions in the commits!


MrSubidubi (approved):
Thanks!

PR #6036, opened 2026-05-12
Add Yulang language support

momota1029:
Adds a highlighting-only Yulang language extension.

The extension uses the tree-sitter-yulang grammar and intentionally does not start a language server yet.

Checks run locally:
- npm test
- npm run build


MrSubidubi (approved):
Thanks!

PR #6047, opened 2026-05-12
Add LaTeX snippets

tzdna:
LaTeX extension with TexLab language server, tree-sitter syntax highlighting, and snippets. https://github.com/Oneptica/Zed-LaTeX-LSP

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -2059,6 +2059,10 @@ version = "0.1.16"
 submodule = "extensions/latex"
 version = "0.2.3"
 
+[latex-lsp]
+submodule = "extensions/latex-lsp"
+version = "0.0.5"
+
 [latte]
 submodule = "extensions/latte"
 version = "0.1.0"
--- a/extensions.toml
+++ b/extensions.toml
@@ -2059,7 +2059,7 @@ version = "0.1.16"
 submodule = "extensions/latex"
 version = "0.2.3"
 
-[latex-lsp]
+[latex-snippets]
 submodule = "extensions/latex-lsp"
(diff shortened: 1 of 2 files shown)


MrSubidubi (changes requested):
Your extension currently violates the [extension publishing prerequisites](https://zed.dev/docs/extensions/developing-extensions#extension-publishing-prerequisites). Can you please elaborate why you chose to include the languages here as opposed to just providing the language server?

tzdna:
Thanks a lot. I think the languages section duplicated existing work from the `zed-latex` extension, and including it in my extension would violate the guidelines. Some of the work should be submitted as a PR to the zed-latex extension, but I haven't done so since it hasn't been updated for 5 months, I'll give it a try anyway. For now, I've removed these parts, and the extension currently provides only `LaTeX snippets`, a feature that the existing zed-latex extension doesn't include, so it can still be used as a standalone extension.

MrSubidubi (approved):
Much better now, thanks!

If anything, we can always re-evaluate the decision if upstream is unresponsive and move things into zed-extensions, for example.

For now, happy to go ahead with the snippets, nice addition for sure

extensions.toml
```diff
@@ -2319,6 +2319,10 @@ version = "0.4.1"
 submodule = "extensions/latex"
 version = "0.2.3"
 
+[latex-snippets]
+submodule = "extensions/latex-snippets"
+version = "0.0.5"
```
MrSubidubi:
```suggestion
version = "0.0.1"
```

PR #6055, opened 2026-05-13
Add device tree source language support

Altair-Bueno:
Although there is an existing devicetree extension, zed-dts uses a different LSP (https://github.com/igor-prusov/dts-lsp vs https://github.com/kylebonnici/dts-lsp). The latter is the same zed-dts uses, and it is the same LSP the [VSCode extension uses](https://marketplace.visualstudio.com/items?itemName=KyleMicallefBonnici.dts-lsp)

This extension is meant to bridge feature parity with VScode. The other pre-existing extension (zed-devicetree) is simpler to use, but it is not as feature-full as this one (due to the LSP implementation). As an alternative, I could modify zed-devicetree to use a different LSP, but I consider it too much of a breaking change for existing users.

image

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1050,6 +1050,10 @@
 	path = extensions/dream
 	url = https://github.com/arturonegrete-dev/Dream-zed
 
+[submodule "extensions/dts"]
+	path = extensions/dts
+	url = https://github.com/Altair-Bueno/zed-dts
+
 [submodule "extensions/duckyscript"]
 	path = extensions/duckyscript
 	url = https://github.com/Lalolog/duckyscript-zed-extension
--- a/extensions.toml
+++ b/extensions.toml
@@ -1065,6 +1065,10 @@ version = "1.0.0"
 submodule = "extensions/dream"
 version = "1.0.0"
 
+[dts]
+submodule = "extensions/dts"
+version = "0.1.0"
+
 [duckyscript]
 submodule = "extensions/duckyscript"
 version = "0.0.1"


Altair-Bueno:
@cla-bot check

MrSubidubi (changes requested):
Thanks for this!

How does this differ from https://zed.dev/extensions/devicetree, could improvements not be made to that extension? See our [extension publishing prerequisites](https://zed.dev/docs/extensions/developing-extensions#extension-publishing-prerequisites) for more info

Altair-Bueno:
> Thanks for this!
> 
> How does this differ from [zed.dev/extensions/devicetree](https://zed.dev/extensions/devicetree), could improvements not be made to that extension? See our [extension publishing prerequisites](https://zed.dev/docs/extensions/developing-extensions#extension-publishing-prerequisites) for more info

The PR description already answers that question


> Although there is an existing devicetree extension, zed-dts uses a different LSP (https://github.com/igor-prusov/dts-lsp vs https://github.com/kylebonnici/dts-lsp). The latter is the same zed-dts uses, and it is the same LSP the [VSCode extension uses](https://marketplace.visualstudio.com/items?itemName=KyleMicallefBonnici.dts-lsp)
> 
> This extension is meant to bridge feature parity with VScode. The other pre-existing extension (zed-devicetree) is simpler to use, but it is not as feature-full as this one (due to the LSP implementation). As an alternative, I could modify zed-devicetree to use a different LSP, but I consider it too much of a breaking change for existing users.

MrSubidubi:
Apologies for my lousy review, did not see that, my bad.

Regarding 
> As an alternative, I could modify zed-devicetree to use a different LSP, but I consider it too much of a breaking change for existing users.

One extension in theory can ship two language servers, so you could add it upstream without introducing much churn. 

Alternatively, we could strip this extension to just provide the language server (see our 
[extension publishing prerequisites](https://zed.dev/docs/extensions/developing-extensions#extension-publishing-prerequisites) for more info). What do you prefer?

Altair-Bueno:
> What do you prefer?

I don't have any preference for either. 

@anikinmd, you contributed the original zed_devicetree extension. Would you like to hop in to the conversation and provide your own opinion? Would you be open to supporting both LSPs under zed_devicetree, or would you rather keep your extension as it is?

anikinmd:
I don't mind switching to a richer LSP. We can keep the previous one as an option as well. 
@Altair-Bueno, what are the major improvements the new LSP will bring? Assuming, tree sitter stays the same, I don't consider it as a real breaking change.

Altair-Bueno:
It would be best to try the LSP yourself via either the VSCode extension or the Zed extension. The https://github.com/Altair-Bueno/zed-dts repo itself has the .zed/settings.json file setup and an examples folder.


Personally the feature I miss the most is node consolidation on hover. Helps a lot to navigate `include` heavy DTs.

> I don't consider it as a real breaking change.

kylebonnici/dts-lsp requires configuration, while igor-prusov/dts-lsp doesn't

MrSubidubi:
Hey everyone, just checking in - what is the current state of things here/how do we want to proceed?

anikinmd:
I'm really sorry, this completely slipped my mind.

My proposal would be to keep the current behavior as the default, while adding the new LSP as an optional secondary LSP, along with configuration templates to make setup easier.

The new LSP is very Zephyr-focused, while the original extension was written with kernel development in mind. Keeping it as an optional alternative would avoid affecting existing users while extending the usefulness of the extension. If I find a way to make the new LSP work without requiring additional configuration, I wouldn't mind switching to it as the default in the future.
@Altair-Bueno, do you agree with this approach?

Altair-Bueno:
Yes, thats ok with me @anikinmd

MrSubidubi:
Sorry for the long wait here, quite a backlog amongst other things but I'll get through it.

Given I just reviewed https://github.com/zed-industries/extensions/pull/7228, will close this for now, thank you both for your efforts! Happy to reopen this should I have missed something. Thanks!

PR #6073, opened 2026-05-14
Update verilog to v0.0.17

someone13574:
- Bumps slang-server and verible language-servers
- Moved check for local binaries to be before querying github for asset names (since the extension uses pinned language server versions)


MrSubidubi (approved):
Thanks!

PR #6100, opened 2026-05-15
Update greycat to v0.1.2

maxleiko:
Update greycat to v0.1.2 (checks for `greycat-analyzer` first, then fallback to the legacy `greycat-lang` if not found)


MrSubidubi (approved):
Thanks!

PR #6102, opened 2026-05-15
Add `One Darker` theme

albertbogusz:
Adds the [One Darker](https://codeberg.org/bogusz/one-darker) theme.

Preview

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4494,6 +4494,10 @@
 	path = extensions/valt
 	url = https://github.com/valtism/valt.git
 
+[submodule "extensions/vanta-theme"]
+	path = extensions/vanta-theme
+	url = https://codeberg.org/bogusz/vanta-zed-theme.git
+
 [submodule "extensions/vapor-theme"]
 	path = extensions/vapor-theme
 	url = https://github.com/felipetesc/vapor-theme.git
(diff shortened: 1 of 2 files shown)


MrSubidubi (commented):
Thanks for this!

It seems that for your theme, we already have an existing ports. Could you please elaborate on why we should publish your theme as a separate extension to the already existing theme and explain where your theme differs from the existing theme/why your changes do not make sense to contribute upstream? Thanks!

albertbogusz:
Hi thanks for the review!

I've tested and seen some of the different variations of a darkened One Dark theme but they didn't quite satisfy what I wanted so I wanted to make my own.

The theme is different from the existing published ones as it includes my own changes to syntax highlighting and (imho) more a consistent colour palette.

I wanted to publish it here because I thought someone might appreciate the differences. The only reason it's called One Darker is because I literally started with One Dark as a base, but it has derived quite a bit away from the original. I can supply a different name if needed.

MrSubidubi:
> I can supply a different name if needed.

If you can think of one that fits, feel free to. You can also keep the connection to OneDark in the description if you like.

albertbogusz:
Closed this abomination in favour of #6285.

PR #6108, opened 2026-05-16
Update short-giraffe-theme submodule commit

kumamaki:
## Summary

Repoints the `extensions/short-giraffe-theme` submodule from `47c8efb` to `052c64e`. Same v2.2.1 content, new commit SHA.

## Why

I rewrote the git history of [kumamaki/Short-Giraffe](https://github.com/kumamaki/Short-Giraffe) to remove personal information that had been committed in earlier releases. Every commit SHA changed as a result, including the one this registry's submodule was pinned to.

The previously pinned commit `47c8efb` is no longer reachable, so new clones of `zed-industries/extensions` would fail when initializing the `short-giraffe-theme` submodule. This PR moves the pointer to `052c64e` — the rewritten commit that the `v2.2.1` tag now resolves to.

Theme files are byte-identical to what shipped as v2.2.1, so no version bump.

Apologies for the extra work.

## Test plan

- [x] `extensions.toml` entry unchanged (`version = "2.2.1"`)
- [x] Only one file changed: the submodule gitlink
- [x] New SHA is reachable on `kumamaki/Short-Giraffe@main` and tagged `v2.2.1`


MrSubidubi (approved):
No worries, all good here.

PR #6113, opened 2026-05-17
Add FujiHaze theme

BinaryMoura:
A serene Zed IDE theme inspired by the quiet beauty of Mount Fuji and the peaceful Japanese countryside.


MrSubidubi (approved):
Thanks!

PR #6146, opened 2026-05-19
Add mc-dp-icons-theme

amqndin:


amqndin:
@cla-bot check

amqndin:
@cla-bot check

MrSubidubi (changes requested):
Thanks for this, theme looks super cool!

Please make sure to follow the [extension publishing guide](https://zed.dev/docs/extensions/developing-extensions#publishing-your-extension) in detail in order to have this published. You are missing a few steps, please update your PR accordingly. Thanks!

amqndin:
i believe i added the "few steps i skipped"! so hopefully this can now be merged.

i do want to add that i actually followed the documentation closely during publishing process, but got confused. so i want to express where the confusion came from, and suggest changes to the documentation. in order to, hopefully, stop other people from being confused about the same thing

in the documentation it says

> Add a new entry to the top-level `extensions.toml` file containing your extension

i misread extension***s***.toml as extension.toml (i hope you can see that its easy to misread that). and so, i assumed "the top-level of `extension.toml`" was referring to prepending the extension.toml of my submodule with the content shown in the documentation. you can actually see me temporarily adding that in the commit history. which i then had to remove, because during testing, that apparently made the extension fail to load.

and so, the change i want to propose is to change the wording, to be more explicit. perhaps something along the lines of

> Add a new entry to the `extensions.toml` file at the root of the `zed-industries/extensions` repo, containing your extension

that would fully erase the confusion, and looking at other pull requests im not the only one that got confused about this part.

MrSubidubi:
Makes sense, you are not the first one. Feel free to submit us a PR upstream and I'll definitely take a look!

We are also looking into more automation for this at the moment, I'll probably consider that case too.

Will review your extension again in a moment

amqndin:
> We are also looking into more automation for this at the moment

ah yes, i imagine a little interactive script would be nice. one that asks you to enter your repo, extension id, name, description you want to provide. then asks whether it is a theme or a snippet and then just does everything for you would be cool. you already have the `sort-extensions` script, so could use the same approach to add an `add-extension` script there!

extensions.toml
```diff
@@ -2377,6 +2377,10 @@ submodule = "extensions/maybe-material"
 path = "extension/"
 version = "0.1.0"
 
+[mc-dp-icons]
```
MrSubidubi:
```suggestion
[mc-dp-icons-theme]
```

MrSubidubi (approved):
That is a good suggestion, I might look into that soonish.

Otherwise looks good, thanks!

PR #6187, opened 2026-05-22
Add ASPX extension

AlexanderHeffernan:
Adds ASP.NET WebForms (ASPX) syntax highlighting with HTML and C# injections.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -222,6 +222,10 @@
 	path = extensions/asn1
 	url = https://github.com/tdanner/zed-asn1.git
 
+[submodule "extensions/aspx"]
+	path = extensions/aspx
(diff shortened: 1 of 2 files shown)


MrSubidubi (changes requested):
Thanks for this!

Given 

> Install the [C# extension](https://zed.dev/extensions/c-) too. Zed doesn't ship C# in core, and this extension injects C# into <% %>, <%= %>, and <%# %> server tags — the injection only takes effect once a language called "CSharp" is registered, which happens when the C# extension is installed.

, how would you feel about instead contributing this to the C# extension itself for an overall improved experience?

MrSubidubi:

PR #6198, opened 2026-05-22
Update opencode to v1.15.9

rekram1-node:
Updating OpenCode extension to v1.15.9

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -3133,7 +3133,7 @@ version = "0.1.0"
 [opencode]
 submodule = "extensions/opencode"
 path = "packages/extensions/zed"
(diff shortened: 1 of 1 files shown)


MrSubidubi (approved):
Thanks!

Please note: We finally got around to migrating our extensions to the ACP registry - users of OpenCode will be migrated with the Zed Stable version released on the 3rd of June. 

Starting with that date, you will not need to publish the updates here anymore. Users on older versions can still get the newer version from the ACP registry (we've had a notice for the migration for months) and users on the new version will be auto-migrated to use OpenCode as provided by the ACP registry. Thanks!

PR #6208, opened 2026-05-23
Add Giossue theme collections

Giossue:
Adds four theme-only extensions:

- `carbon-theme`
- `dark-foundry-theme`
- `glass-theme`
- `slate-ui-theme`

Each extension is published as a separate public repository and includes its own MIT license.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,3 @@
+<!-- Please confirm the checklist below, then remove this comment. -->
+
+- [ ] I've read [CONTRIBUTING.md](../CONTRIBUTING.md) and followed the relevant guidance for adding or updating my extension.
--- a/.github/workflows/actionlint.yml
+++ b/.github/workflows/actionlint.yml
@@ -18,7 +18,7 @@ jobs:
       SHA: 03d0035246f3e81f36aed592ffb4bebf33a03106
       VERSION: 1.7.7
     steps:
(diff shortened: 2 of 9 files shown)


Giossue:
@cla-bot check

MrSubidubi (changes requested):
Hey, thanks for this and sorry for the slow response!

Overall all of these looks good, however, due to my late response one conflict for a ID arose, since I merged another theme first whilst going through the queue. 

Could I kindly ask you to resolve the conflicts? Thank you!

MrSubidubi:
Ah and I'd also be slightly interested in your choice to publish this as four distinct themes

Giossue:
Thanks for the review! I resolved the ID conflict by renaming my Glass extension ID/path from `glass-theme` to `giossue-glass-theme`, while keeping the visible extension name as `Glass Theme Collection`.

About publishing these as four distinct themes: I split them because each collection has its own visual direction and identity, and keeping them separate makes each listing more focused and easier to evolve independently. That said, I’m happy to consolidate them into a single theme pack if that would be preferred for the registry.

MrSubidubi (approved):
Since you are that responsive and also were able to actually follow our documentation, which makes reviewing this that much easier, I'll happily greenlight this here as is. 

Please do us one favor and publish updates for each theme individually (you can also take a look at the community action for this) as that helps us a lot with reviewing.

Looks good now here, thanks!

Giossue:
Thanks for the review and for the guidance! I’ll keep future updates split per theme.

PR #6252, opened 2026-05-27
Add diff-tool language server

xgfone:


xgfone:
@cla-bot check

xgfone:
Thanks!

Although this was raised in #6256 rather than this one, I applied the same publishing-prerequisite fixes here as well for consistency:

- Kept the extension ID aligned with the Zed extensions registry entry: `diff-tool`.
- Removed the bundled/local LSP server code path from the extension runtime.
- The extension now resolves the LSP only through:
  - `lsp.diff-tool.binary.path`
  - `ZED_DIFF_TOOL_LSP`
  - `diff-tool-lsp` from `PATH`
  - GitHub Release download
- Updated the README to clarify that the local scripts are only for development and do not bundle the LSP server.

Please let me know if anything else should be adjusted here too.

MrSubidubi:
Thanks for this and sorry for the long wait!

I just checked - if you select two files in the project panel, then open the context menu you also get the `compare marked files` action. I do not see how this differs from that really, aside from being an external thing we already ship bundled.

Gonna close it with this, but happy to reopen after discussing should you have a different opinion on this matter. Thanks!

PR #6256, opened 2026-05-27
Add json-tool language server

xgfone:


MrSubidubi (changes requested):
Thanks for this!

Also, please remove the `bundled_server` code path, as that really will never be used and I'd prefer other people checking this not having to think about this themselves. Thanks!

xgfone:
Thanks for the review!

I’ve updated the extension according to the publishing prerequisites:

- Kept the extension ID aligned with the Zed extensions registry entry: `json-tool`.
- Removed the bundled/local LSP server code path from the extension runtime.
- The extension now resolves the LSP only through:
  - `lsp.json-tool.binary.path`
  - `ZED_JSON_TOOL_LSP`
  - `json-tool-lsp` from `PATH`
  - GitHub Release download
- Updated the README to clarify that the local scripts are only for development and do not bundle the LSP server.

Please let me know if there’s anything else you’d like adjusted.

MrSubidubi (changes requested):
Yes, please suffix or prefix your extension ID with something LSP-related, please

xgfone:
OK. It's been renamed to `json-tool-lsp`.

PS: `diff-tool-lsp` is my another PR #6252.

MrSubidubi (approved):
Looks good now, thanks!

In a future release, you can remove the gating on `make_file_executable`, since that already noops on Windows. Also, feel free to remove the ENV variable reading for the path, you should just do that with settings.

Not blocking here, so happy to merge. Thanks!

PR #6295, opened 2026-05-30
Add Papyrus SSE language support

Hexanode0x0:
Papyrus is a scripting language used in Bethesda Softworks games. This extension targets specifically the version of Papyrus used in Skyrim Special Edition (SSE).

The extension currently implements a tree-sitter parser for the language and some snippets.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -3206,6 +3206,10 @@
 	path = extensions/papercolor
 	url = https://github.com/emirror-de/papercolor-zed.git
 
+[submodule "extensions/papyrus-sse"]
+	path = extensions/papyrus-sse
+	url = https://github.com/Silent-City-Projects/zed-papyrus-sse.git
+
 [submodule "extensions/paraiso"]
 	path = extensions/paraiso
 	url = https://github.com/treffynnon/zed-Paraiso.git
--- a/extensions.toml
+++ b/extensions.toml
@@ -3252,6 +3252,10 @@ version = "1.0.0"
 submodule = "extensions/papercolor"
 version = "0.2.0"
 
+[papyrus-sse]
+submodule = "extensions/papyrus-sse"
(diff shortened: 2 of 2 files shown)


MrSubidubi (changes requested):
Hey, thanks for this and apologies for getting back to you with such a delay!

I am currently working through the backlog and also merged https://github.com/zed-industries/extensions/pull/7049 here. Given our [extension publishing prerequisites](https://zed.dev/docs/extensions/developing-extensions#extension-publishing-prerequisites), would be curious whether the other extension renders this one obsolete or whether there are differences that might make it reasonable to have both. We aim to only have one extension per language and the other extension seemed to help with the more general support for this language. Thus, went ahead and merged that one.

Yet, would be curious if there might be reasons to have both of these or whether the other extension takes care of it all.

Hope you understand our reasoning here, in any case, apologies for such a late response!

Hexanode0x0:
Wow, it sure seems like that's just a more complete version of what I wanted to do.
I've been bouncing between quite a few modding projects, so I haven't had as much time to work on this as I'd like to.
I'm glad someone else has made a quality extension, I'm excited to try it out.

The only reason to have separate extensions for Papyrus from different games would be that Papyrus differs a lot between games. In Fallout 4, Papyrus has support for structs, while in Skyrim SE it doesn't. In the Skyrim version, putting complex expressions in array braces often results in undefined behavior, this was fixed in Fallout 4, etc.. To someone who's just started with scripts, this version nuance might not be obvious.
Though it's not like this can't be handled in a single extension.

Feel free to close this PR, and thanks!

MrSubidubi:
Appreciate the honest answer here, if you feel like it, I'd not be opposed to you contributing some improvements upsteam.

Either way, thanks for the work that went into this and understanding the reasoning here! Thank you!

PR #6300, opened 2026-05-30
Update ssh-config to v0.2.0

pranavmangal:
Updates ssh-config extension to v0.2.0.

It now uses the latest [tree-sitter-ssh-config](https://github.com/tree-sitter-grammars/tree-sitter-ssh-config), which allows escaped quotes in strings.


MrSubidubi (approved):
Thanks!

PR #6301, opened 2026-05-30
Add Mojo language support

vadim-su:
Adds the Mojo language extension to the Zed extension registry.

Extension repository: https://github.com/vadim-su/zed_mojo

Includes:
- Mojo language configuration
- Tree-sitter grammar registration
- Syntax/editor queries
- LSP integration for `mojo-lsp-server`
- Snippets and runnables support

Local checks run:
- `npm run sort-extensions`
- `npm run build`
- `npm run test`
- `git diff --cached --check`

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2686,6 +2686,10 @@
 	path = extensions/modus-themes
 	url = https://github.com/vitallium/zed-modus-themes.git
 
+[submodule "extensions/mojo"]
+	path = extensions/mojo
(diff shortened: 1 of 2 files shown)


helehex:
The extension could still use some work, but it would be nice to have in the registry.

MrSubidubi (approved):
Indeed could, but looks good for now, sorry for the delay!

Note that I think you probably don't need to pass the language server the entire user environment, but not blocking here.

Thank you!

PR #6309, opened 2026-05-31
Add Local History extension

neuroborus:
Adds [Local History](https://github.com/neuroborus/zed-local-history), a filesystem-first local history/recovery extension for Zed.

Extension path: `editors/zed`
Version: `0.1.0`
Submodule: `extensions/local-history`

Local History watches file saves, stores snapshots outside the project, exposes them to the Zed Agent Panel through MCP, and can restore a previous version while creating a safety snapshot first.


neuroborus:
@cla-bot check

MrSubidubi:

PR #5848, opened 2026-05-01
Update datadog-mcp to v0.3.0

robertohuertasm:
Fixes broken extension and adds support for toolsets.

PR #6310, opened 2026-05-31
Update datadog-mcp to v0.3.0

robertohuertasm:
Fixes broken extension, adds support for toolsets and adds support for the new UK1 server.

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -1010,7 +1010,7 @@ version = "0.3.7"
 
 [datadog-mcp]
 submodule = "extensions/datadog-mcp"
-version = "0.2.1"
+version = "0.3.0"
 
 [day-shift]
 submodule = "extensions/day-shift"


robertohuertasm:
@MrSubidubi, I updated the extension and tried to avoid the issues you mentioned in the [last PR ](https://github.com/zed-industries/extensions/pull/5848). Hopefully everything is fine and we can make the extension work again. It's currently broken in the marketplace.

MrSubidubi (approved):
Thanks!

We'll probably revisit how we handle remote MCPs in the near future, but expect us to keep you posted on this. If you have any input on this, feel free to share.

robertohuertasm:
Thanks @MrSubidubi! I noticed you had remote MCP support in your configuration, but then I saw there was not API surface from the extensions yet. It would be nice for the extensions to have a way to contribute the url. Then the Datadog MCP extension would be super simple, as we would only have to manage the region and the toolsets, which are all encoded in the url itself.

I'll stay tuned to see if you open that API soon.

Thanks again for your help!

PR #6313, opened 2026-05-31
Update phpmd to v0.1.2

mikebronner:
Bumps `phpmd` to v0.1.2.

Release notes: https://github.com/mike-bronner/zed-phpmd-lsp/releases/tag/0.1.2

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -3300,7 +3300,7 @@
 
 [submodule "extensions/phpmd"]
 	path = extensions/phpmd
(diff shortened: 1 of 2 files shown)


MrSubidubi (approved):
RE

> The upstream `huacnlee/zed-extension-action` creates a new branch with a
timestamped name on every run (`update-<ext>-<unix-time>`), which means
re-running the publish step opens a duplicate PR every time instead of updating
the in-flight one.

That should not be true anymore, I fixed that some time ago.

PR #6333, opened 2026-06-03
Add VapourSynth extension

Cola-Ace:
Adds the VapourSynth extension to the registry. Checks: pnpm sort-extensions; pnpm test.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4574,6 +4574,10 @@
 	path = extensions/vapor-theme
 	url = https://github.com/felipetesc/vapor-theme.git
 
+[submodule "extensions/vapoursynth"]
+	path = extensions/vapoursynth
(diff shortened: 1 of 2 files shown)


Cola-Ace:
@cla-bot check

MrSubidubi:
Thank you for this and sorry for the way too long delay here!

Sadly, this violates our 
[extension publishing prerequisites](https://zed.dev/docs/extensions/publishing/prerequisites) in many aspects - I think eventually this would be nice to have, but currently this is way too hardcoded to Windows, so would break on many machines, as well as the scripts not being something that we want to have in extensions. 

That said, this overall sounds promising/nice to have, so would be happy to help here if possible. Currently though, due to the large backlog and this not being in a shippable state, will close, sorry. Please do reach out if you have any questions. Thank you.

PR #6336, opened 2026-06-03
Add epics language support

DerBaumBoy:
Add syntax highlighting for [EPICS](https://epics-controls.org/).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1270,6 +1270,10 @@
 	path = extensions/env
 	url = https://github.com/zarifpour/zed-env
 
+[submodule "extensions/epics"]
+	path = extensions/epics
+	url = https://gitlab.com/benji_boy/epics-zed.git
+
 [submodule "extensions/erb-snippets"]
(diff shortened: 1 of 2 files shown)


MrSubidubi (changes requested):
Overall looks fine, please update the extension ID in your repo to the one here, then update the submodule. 

Another note, do we need the broad suffix patterns of `cmd` and `proto` or can these be more tightly scoped?

DerBaumBoy:
ID is updated as requested.

Sadly EPICS is pretty firm on the file extensions. In theory proto could be renamed to protocol, if that is any better, but that may lead to usability issues with string length restrictions in EPICS.
So if the somewhat arbitrary extensions are not a reason for denial, I would like to keep them.

DerBaumBoy:
I have rechecked all extension IDs in all repos, and I don't see why the pipeline is failing. If you could give me any hint as to why, I would be thankful.

Edit: last pipeline ran before my ID update push, so that makes sense.

PR #6366, opened 2026-06-05
Add Ansible Lens extension

jan-krueger:
Ansible Lens is a language server that brings go-to-definition, find-references, hover, completion, and undefined-variable diagnostics to nested Ansible & Jinja2 variables - resolving them across group_vars, host_vars, role defaults/vars, and set_fact/register.


jan-krueger:
@cla-bot check

PR #6367, opened 2026-06-05
Update valt

valtism:
Update valt theme to 1.0.1


MrSubidubi:
Thanks for this!

This is missing important steps - please see https://zed.dev/docs/extensions/developing-extensions#updating-an-extension for a detailed guide and then resubmit. Thanks!

PR #6372, opened 2026-06-06
Update termux to v0.1.3

paveloom:


paveloom:
Seems like Zed extensions still can't use built-in grammars (see https://github.com/zed-industries/extensions/pull/2763), so I ship the [Bash grammar](https://github.com/tree-sitter/tree-sitter-bash) under the disguise of the [Ebuild grammar](https://gitlab.com/paveloom-g/forks/tree-sitter-ebuild). This is in response to [gitlab.com/paveloom-g/rust/zed-termux#1](https://gitlab.com/paveloom-g/rust/zed-termux/-/work_items/1) and [github.com/zed-industries/zed#58549](https://github.com/zed-industries/zed/issues/58549).

MrSubidubi (approved):
Yeah fair point and good workaround, thanks!

PR #6378, opened 2026-06-07
Add hex color preview

sandy3002:
image
can't find a similar extension to show the color of a hex code, so here is one

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4993,3 +4993,6 @@
 [submodule "extensions/zwirn"]
 	path = extensions/zwirn
 	url = https://codeberg.org/polymorphicengine/zwirn-zed-extension.git
+[submodule "extensions/hex-color-preview"]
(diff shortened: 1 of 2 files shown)


sandy3002:
@cla-bot check

MrSubidubi (changes requested):
We do have https://zed.dev/extensions/color-highlight which I think does precisely what you want - is that the case? You could go ahead and ask them to update their description so that this is more easily discoverable.

If you want to proceed with this, please make sure your extension fulfills the [extension publishing prerequisites](https://zed.dev/docs/extensions/developing-extensions#extension-publishing-prerequisites). Particularly, you will need to update your extension ID as outlined in the documentation. Once you did the necessary updates, please update everything related here. Thanks!

sandy3002:
Thanks for pointing out! That is exactly what i needed. Closing the pr

PR #6381, opened 2026-06-07
Add template-string-converter extension

ivan-szz:
Auto-converts quotes to backticks when typing ${} inside JS/TS strings.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4258,6 +4258,10 @@
 	path = extensions/templ
 	url = https://github.com/makifdb/zed-templ.git
 
+[submodule "extensions/template-string-converter"]
+	path = extensions/template-string-converter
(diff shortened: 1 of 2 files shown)


ivan-szz:
@cla-bot check

ivan-szz:
@cla-bot check

MrSubidubi:
Thanks for this!

However, in the current form, this will not work for users - you will need to create releases and then download one of the assets for the server to work for users, as they cannot install it with that comand. See our 
[extension publishing prerequisites](https://zed.dev/docs/extensions/developing-extensions#extension-publishing-prerequisites) which should hopefully link to some content that might help you.

Will close for now, feel free to resubmit with that taken care of. Thanks!

PR #6400, opened 2026-06-09
Add LEQL extension

vrsko:
Adds syntax highlighting for LEQL (Log Entry Query Language).
Uses Tree-sitter grammar.
Extension repository:  https://codeberg.org/vrsko/leql


MrSubidubi:
Closing since the CLA has not been signed in two weeks. Happy to take another look with that dealt with

PR #6411, opened 2026-06-10
Add Ember.js language server

sathish-pv:
Adds language server for [Ember.js](https://emberjs.com/) using [Glint](https://typed-ember.gitbook.io/glint).
Also add Syntax highlighting.

Extension Repository: https://github.com/sathish-pv/zed-ember-lsp

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1246,6 +1246,10 @@
 	path = extensions/ember
 	url = https://github.com/jylamont/zed-ember.git
 
+[submodule "extensions/ember-lsp"]
+	path = extensions/ember-lsp
+	url = https://github.com/sathish-pv/zed-ember-lsp.git
+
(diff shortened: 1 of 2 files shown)


MrSubidubi (changes requested):
Thanks for this! 

We definitely will need to update the extension ID here, but I do have a few more questions: How precisely does the language server interact with the typescript language server? If I see that correctly, could it be that we instead would want for this to use https://docs.rs/zed_extension_api/latest/zed_extension_api/trait.Extension.html#method.language_server_additional_workspace_configuration instead to configure the plugin?

sathish-pv:
I don't think language_server_additional_workspace_configuration can be used but language_server_additional_initialization_options can be utilized without running the new server. Will update my handling for GlintV2 to initialize ts lsp with the needed plugin.

What change needed in extension ID?

PR #6421, opened 2026-06-10
Update svelte-effect-runtime-language-server to v2.3.1

sandersonstabo:
## Summary
- Updates \\svelte-effect-runtime-language-server\\ from 2.3.0 to 2.3.1.
- Points the submodule at usebarekey/svelte-effect-runtime@8045e3a, which packages the Zed LSP from the v2.3.1 GitHub release asset instead of resolving the private language-server package from npm.
- Release: https://github.com/usebarekey/svelte-effect-runtime/releases/tag/v2.3.1

## Checks
- \\corepack pnpm sort-extensions\\`n- \\corepack pnpm build\\`n- \\corepack pnpm test\\ (111 tests)

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -4260,7 +4260,7 @@ version = "0.2.11"
 [svelte-effect-runtime-language-server]
 submodule = "extensions/svelte-effect-runtime-language-server"
 path = "modules/svelte-effect-runtime-zed"
-version = "2.3.0"
+version = "2.3.1"
 
 [svelte-mcp]
 submodule = "extensions/svelte-mcp"


MrSubidubi (changes requested):
Can you please elaborate why you need the download here and can not publish this to NPM instead?

sandersonstabo:
> Can you please elaborate why you need the download here and can not publish this to NPM instead?

I think that's an easier approach to fetch it from GitHub releases right now, as  we don't publish the LSP as a standalone package to NPM currently

MrSubidubi:
That is fair, but we generally discourage doing this as per our 
[extension publishing prerequisites](https://zed.dev/docs/extensions/developing-extensions#extension-publishing-prerequisites). If you want to install a language server that should be invoked through node, you should publish it to NPM and ensure that way that all the requirements needed are present.

sandersonstabo:
New PR at https://github.com/zed-industries/extensions/pull/6599

PR #6435, opened 2026-06-11
Remove `sqruff` extension

maxdeviant:
This PR removes the `sqruff` extension, as the upstream repository is gone: https://github.com/gvozdvmozgu/sqruff-zed


nyabinary:
It's still there though? Also what's the alternative?

PR #6453, opened 2026-06-12
Add SchemaLock language server

lgogolin:
Adds the **SchemaLock** extension (id `schemalock-lsp`): schema-pinned validation, completion, and hover for Kubernetes-style YAML, backed by a language server.

- **Repo:** https://github.com/schemalock/zed
- **ID:** `schemalock-lsp`
- **Version:** `0.1.0`
- **License:** Apache-2.0

The extension does **not** bundle a binary — it downloads the pinned per-platform language server from the public [`schemalock/schemalock` GitHub releases](https://github.com/schemalock/schemalock/releases) on first use (via the release API) and caches it. Registers the `schemalock` language server for YAML; unowned YAML is transparently proxied to `yaml-language-server`.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -3894,6 +3894,10 @@
 	path = extensions/scala
 	url = https://github.com/scalameta/metals-zed.git
 
+[submodule "extensions/schemalock"]
+	path = extensions/schemalock
+	url = https://github.com/schemalock/zed.git
+
 [submodule "extensions/scheme"]
 	path = extensions/scheme
 	url = https://github.com/zed-extensions/scheme.git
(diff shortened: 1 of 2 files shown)


MrSubidubi (changes requested):
Thanks for this!

Also, please use normal GitHub assets and the corresponding APIs instead, I see literally no reason why you should publish this to your own CDN.

lgogolin:
Thanks for the review! Both addressed:

1. **Extension ID** is now `schemalock-lsp` (per the publishing prerequisites) — updated the id, submodule path, and `extensions.toml` entry here.
2. **Dropped the CDN** — the binary is now downloaded from the public `schemalock/schemalock` GitHub releases via the release API. The submodule points at the commit with both changes.

MrSubidubi (approved):
Thanks!

PR #6466, opened 2026-06-13
Update maho-lsp to v0.10.0

fballiano:
Bumps `maho-lsp` to v0.10.0.

This release adds a Model Context Protocol (MCP) context server (`maho-intelligence-mcp`) for Zed's AI agent, exposing Maho's runtime configuration (class aliases, rewrites, merged config, modules, events, EAV attributes, DB schema, and more), alongside the existing LSP integration.

Release notes: https://github.com/MahoCommerce/zed/releases/tag/v0.10.0

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -2452,7 +2452,7 @@ version = "0.1.0"
 
 [maho-lsp]
 submodule = "extensions/maho-lsp"
-version = "0.9.1"
+version = "0.10.0"
 
 [mainframe-theme]
 submodule = "extensions/mainframe-theme"


MrSubidubi (changes requested):
As per our [extension publishing prerequisites](https://zed.dev/docs/extensions/developing-extensions#extension-publishing-prerequisites), I am not sure we'd want this - we are moving away from MCP extensions anyway and also do not want too many vastly different features per extension.

Happy to have your thoughts on this

fballiano:
Hi @MrSubidubi, I see MCP extensions at https://zed.dev/docs/extensions/mcp-extensions, are they going to be removed? The point of my Maho extension is to provide the full "intelligence" to the IDE when working with Maho ecommerce platform, and all the insights are available via LSP and MCP, that's why I added the support for both.

I don't think they're 2 vastly different features, I mean, they are, but they're all working toward making the developer effective when working on Maho with Zed. In my opinion having only one of them doesn't provide the full picture and separating the 2 extensions seem a bit overkill.

I don't know what to do here :-\

MrSubidubi (approved):
Will allow it for now, but please be advised that we'll change it in the future. Thanks!

PR #6468, opened 2026-06-13

yankeeinlondon:
Adds **Excalidraw Preview** (`excalidraw-preview`): previews `.excalidraw`, `.excalidraw.svg`, and `.excalidraw.png` files in a native WebView window with live reload, native export save dialogs, and persistent shape library.

- Repo: https://github.com/yankeeinlondon/excalidraw-zed-extension
- The extension manifest is in the `extension/` subdirectory (hence `path = "extension"`).
- The companion native binary is downloaded at runtime from GitHub Releases (v0.3.0), not bundled.
- LICENSE present at repo root.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1326,6 +1326,10 @@
 	path = extensions/evolved-theme
 	url = https://github.com/evoL/evolved-theme.git
 
+[submodule "extensions/excalidraw-preview"]
+	path = extensions/excalidraw-preview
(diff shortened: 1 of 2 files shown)


yankeeinlondon:
@cla-bot check

MrSubidubi:

yankeeinlondon:
I have resubmitted this without the slash command features.

PR #6488, opened 2026-06-15
Add Neon Context Server

peterkyle01:
Adds the Neon Context Server extension: MCP server for Neon serverless Postgres with 10 tools - read-only SQL, schema exploration, project/branch management, EXPLAIN plans, and masked connection strings.

Source: https://github.com/peterkyle01/neon-context-server (pinned to tag v0.1.0).
The server binary is auto-downloaded from GitHub releases (linux-x86_64, macos-arm64, macos-x86_64) via the extension wrapper.

Only NEON_API_KEY is required; project, branch, and database are auto-discovered.

Tested locally as a dev extension against live Neon databases.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2922,6 +2922,10 @@
 	path = extensions/neon-comfy-soft-themes
 	url = https://github.com/guustavocl/zed-neon-comfy-soft-themes
 
+[submodule "extensions/neon-context-server"]
+	path = extensions/neon-context-server
+	url = https://github.com/peterkyle01/neon-context-server.git
+
 [submodule "extensions/neon-cyberpunk"]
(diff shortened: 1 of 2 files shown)


MrSubidubi (changes requested):
Thanks for this!

Could you elaborate why this cannot just be added as a remote context server?

peterkyle01:
> Thanks for this!
> 
> Could you elaborate why this cannot just be added as a remote context server?

This uses a personal api key to query  neon locally from the binary, having it as a remote context server will proxy it to a third party which people are not fond of but this way it runs safely locally calling neon directly. I wanted to uphold security as one if its main reasons that's why it also does not alter data.

PR #6501, opened 2026-06-16

yankeeinlondon:
Adds **Excalidraw Preview** (`excalidraw-preview`): previews `.excalidraw`, `.excalidraw.svg`, and `.excalidraw.png` files in a native WebView window with live reload on save, native export save dialogs, and a persistent shape library.

- Repo: https://github.com/yankeeinlondon/excalidraw-zed-extension
- Manifest is in the `extension/` subdirectory (`path = "extension"`).
- The companion native binary is downloaded at runtime from GitHub Releases (`v0.4.2`), not bundled.
- LICENSE present at repo root.

### Re-submission of #6468 — slash-command API removed

This re-opens #6468, which @MrSubidubi closed because the extension exposed `/preview-excalidraw` and `/new-excalidraw` via the slash-command API (reserved for agents). Thank you for that feedback.

Both slash commands have been **removed entirely**. The extension now provides only the `excalidraw-preview` language server, which opens the preview automatically on `didOpen`/`didSave` for `.excalidraw*` files — so the open-preview flow no longer depends on slash commands. The extension's only API surface is `language_server_command`.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1338,6 +1338,10 @@
 	path = extensions/evolved-theme
 	url = https://github.com/evoL/evolved-theme.git
 
+[submodule "extensions/excalidraw-preview"]
+	path = extensions/excalidraw-preview
+	url = https://github.com/yankeeinlondon/excalidraw-zed-extension.git
+
 [submodule "extensions/exograph"]
 	path = extensions/exograph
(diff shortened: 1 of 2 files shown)


MrSubidubi:

PR #6503, opened 2026-06-16
Add Elsa language support

MrPoloGit:
Adds Elsa (elsa): syntax highlighting and completion for the Elsa (https://github.com/ucsd-progsys/elsa) lambda calculus teaching language (.lc files).

Repo: https://github.com/MrPoloGit/elsa-lang

Tree-sitter grammar: https://github.com/MrPoloGit/tree-sitter-elsa

The companion language server [elsa-lsp](https://github.com/MrPoloGit/elsa-lsp) is downloaded at runtime from GitHub Releases (v0.1.0), not bundled. It provides completion for let bindings defined in the current file.

LICENSE is present at every repo root.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1274,6 +1274,10 @@
 	path = extensions/elm
 	url = https://github.com/zed-extensions/elm.git
 
+[submodule "extensions/elsa-lang"]
+	path = extensions/elsa-lang
+	url = https://github.com/MrPoloGit/elsa-lang.git
+
 [submodule "extensions/ember"]
 	path = extensions/ember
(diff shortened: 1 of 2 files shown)


MrSubidubi (approved):
Very nice (and would have actually helped me recently 😅 )!

You may eventually replace https://github.com/MrPoloGit/elsa-lang/blob/349bdaf15d32dbc28bfebc4cb39e9e1d27564340/src/lib.rs#L58 with `zed::github_release_by_tag_name` (or something amongst those lines) which would be a bit better, but is definitely not blocking here.

Thanks!

PR #6504, opened 2026-06-16
Add ASP.NET extension

muamars:


MrSubidubi:
Thanks for this!

There have been previous attempts at this, and I'll hit you with the same fate - we would want the Razor highlighting support to live natively in the C# extension so that we can provide an optimal experience for users. This would also let you get rid of the duplicate omnisharp support, which is also against our 
[extension publishing prerequisites](https://zed.dev/docs/extensions/developing-extensions#extension-publishing-prerequisites).

I will close with that, we'd appreciate if you could shoot us a PR at https://github.com/zed-extensions/csharp. Thanks!

PR #6510, opened 2026-06-16
Add angular-ng extension

ArielJMR:
This adds **angular-ng**, an Angular Language support extension with project-aware version auto-detection.

It is a fork of the existing `angular` extension that addresses a common pain point: the original pins a fixed, modern `@angular/language-server` version, which breaks on projects using older Angular. `angular-ng` instead:

- **Auto-detects the project's Angular version** — it prefers a `@angular/language-server` already installed in the project's `node_modules` (exact match, covers monorepos), otherwise derives a compatible language-server + TypeScript pair from the project's `package.json` (Angular 13 → latest), falling back to a modern default.
- **Forwards the Angular Language Service settings** like the official VSCode extension: feature flags (`forceStrictTemplates`, `autoImports`, snippet/optional-chain completions, suppressed diagnostics), `--angularCoreVersion`, and inlay hints (on by default, configurable via `initialization_options`).

The internal language server name remains `angular`. A separate `id` (`angular-ng`) is used so it does not collide with the existing `angular` entry in the registry.

Repository: https://github.com/ArielJMR/zed-angular
License: MIT

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -138,6 +138,10 @@
 	path = extensions/angular
 	url = https://github.com/nathansbradshaw/zed-angular.git
 
+[submodule "extensions/angular-ng"]
+	path = extensions/angular-ng
+	url = https://github.com/ArielJMR/zed-angular.git
+
 [submodule "extensions/angular-snippets"]
 	path = extensions/angular-snippets
 	url = https://github.com/hudson-newey/angular-snippets-for-zed.git
--- a/extensions.toml
+++ b/extensions.toml
@@ -142,6 +142,10 @@ version = "0.1.2"
 submodule = "extensions/angular"
 version = "0.0.5"
 
+[angular-ng]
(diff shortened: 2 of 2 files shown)


ArielJMR:
@cla-bot check

MrSubidubi (changes requested):
Thanks but this definitely violates our [extension publishing prerequisites](https://zed.dev/docs/extensions/developing-extensions#extension-publishing-prerequisites). 

Can we please find a better approach for this if we want to proceed with this here or go with instead improving things upstream?

ArielJMR:
Thanks for the review — you're right, this is covered by the publishing prerequisites (improve an existing extension upstream rather than submitting a fork). I've opened a PR upstream against the original `angular` extension instead: nathansbradshaw/zed-angular#90

Closing this one. Thanks!

PR #6517, opened 2026-06-17
Update greycat to v0.1.7

maxleiko:
Bumps GreyCat to v0.1.7.

Updates the bundled tree-sitter-greycat grammar (c346a0c): `fn` is now
accepted as an attribute name (`type Foo { fn: String; }`) and as an
object-field key (`Foo { fn: 1 }`), which previously produced parse errors.


MrSubidubi (approved):
Thanks!

PR #6520, opened 2026-06-17
Add fastapi-lsp

alex-oleshkevich:
Adds [fastapi-lsp](https://github.com/alex-oleshkevich/fastapi-lsp), a language server for FastAPI and Starlette.

Provides diagnostics, completions, and hover for Python, HTML, and Jinja2 files.

Tested locally as a dev extension in Zed.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1374,6 +1374,10 @@
 	path = extensions/fantasticons-icons-theme
 	url = https://github.com/caiolandgraf/fantasticons-zed.git
 
+[submodule "extensions/fastapi-lsp"]
+	path = extensions/fastapi-lsp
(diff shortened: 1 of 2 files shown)


MrSubidubi (changes requested):
Please remove https://github.com/alex-oleshkevich/fastapi-lsp/blob/92d272aef170ee3f5901c87bd3b5e4b5c6d521bd/editors/zed/src/lib.rs#L19-L23 - users should have this configured themselves if needed so that `Worktree::which` resolves this properly. Furthermore, `std::env::var` will not work in the WASM context anyway.

Please update your submodule here after. Thank you!

alex-oleshkevich:
@MrSubidubi done

MrSubidubi (approved):
Thanks!

PR #6524, opened 2026-06-17
Add one-dark-pro-night-flat theme

LaurenceRawlings:
## Summary

Adds the **OneDark Pro Night Flat** theme for Zed, a theme ported from [One Dark Pro](https://github.com/Binaryify/OneDark-Pro).

- **Extension id:** `one-dark-pro-night-flat`
- **Repository:** https://github.com/LaurenceRawlings/zed-one-dark-pro-night-flat

## Checklist

- [x] Extension has `extension.toml` with valid metadata
- [x] Theme JSON follows the Zed theme schema
- [x] MIT license included in the extension repository
- [x] Submodule added under `extensions/one-dark-pro-night-flat` and registered in `extensions.toml`
- [x] `extensions.toml` and `.gitmodules` sorted via `pnpm sort-extensions`
- [x] Tested locally as a dev extension

## Note

The `-theme` suffix for the id was left out to match the other theme extensions surrounding it in `extensions.toml`. This can be added if required.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -3250,6 +3250,10 @@
 	path = extensions/one-dark-pro-monokai-darker
 	url = https://github.com/9ssi7/zed-one-dark-pro-monokai-darker.git
 
+[submodule "extensions/one-dark-pro-night-flat"]
+	path = extensions/one-dark-pro-night-flat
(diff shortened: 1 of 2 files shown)


MrSubidubi (changes requested):

Depending on the answer, would also require the ID update. Thanks!

PR #6526, opened 2026-06-18
Add fivem-natives extension

Tiltann:
Adds LuaLS type definitions and snippets for FiveM native functions (client, server, shared) in Zed.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -5161,3 +5161,6 @@
 [submodule "extensions/zwirn"]
 	path = extensions/zwirn
 	url = https://codeberg.org/polymorphicengine/zwirn-zed-extension.git
+[submodule "extensions/fivem-natives"]
+	path = extensions/fivem-natives
+	url = https://github.com/Tiltann/fivem-natives-zed.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1426,6 +1426,10 @@
 	path = extensions/fish
 	url = https://github.com/hasit/zed-fish.git
 
+[submodule "extensions/fivem-natives"]
+	path = extensions/fivem-natives
+	url = https://github.com/Tiltann/fivem-natives-zed.git
+
 [submodule "extensions/fjord-theme"]
 	path = extensions/fjord-theme
 	url = https://github.com/fjord-themes/fjord-zed.git
@@ -5161,6 +5165,3 @@
 [submodule "extensions/zwirn"]
 	path = extensions/zwirn
 	url = https://codeberg.org/polymorphicengine/zwirn-zed-extension.git
-[submodule "extensions/fivem-natives"]
-	path = extensions/fivem-natives
-	url = https://github.com/Tiltann/fivem-natives-zed.git
--- a/extensions.toml
+++ b/extensions.toml
@@ -1445,12 +1445,12 @@ version = "0.1.1"
 submodule = "extensions/firrtl-source-locator"
 version = "0.1.0"
 
-[fivem-natives]
-submodule = "extensions/fivem-natives"
-version = "0.1.0"
-
 [fish]
 submodule = "extensions/fish"
+version = "0.1.0"
+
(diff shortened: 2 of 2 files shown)


Tiltann:
@cla-bot check

MrSubidubi:
Hey @Tiltann, thanks for this (and all of these submissions) and apologies for the delayed feedback! I really appreciate you taking the time to get these up and the attempt to improve the ecosystem!

Unfortunately, at this time, I'll very sadly need to close all of these - not because I think these are bad or anything, but they will not work in a released environment. This was something you could not have known, and I'll make sure to finally get rid of this descrepancy between dev extensions and released extensions in the near future.

The issue being, that all of your extensions rely on the `definitions` being present - however, when we compile the extension, this will no longer be the case, as we do not copy that directory under any circumstances. In development, howver, it is present as you'd expect it to be. This happens because we treat dev extensions differently while we really should not - I'll personally make sure to fix this soon in the future.

However, this results in your extensions sadly not being ready to be published at this stage - I am sorry for that. You can get some of these up again by using our `download_file` API (https://docs.rs/zed_extension_api/latest/zed_extension_api/fn.download_file.html), so that you can make the files available again.

Aside from this biggest issue, two more things
- I think we'd need to update the IDs to align better with our 
[extension publishing prerequisites](https://zed.dev/docs/extensions/developing-extensions#extension-publishing-prerequisites) - that said, I am not even sure how precisely, since this is and was somewhat a first. But I think just `intellisense` might not be the best, but happy to hear your opinion
- I doubt that `workspace_configuration` is the right API here and curious whether we should not instead use https://docs.rs/zed_extension_api/latest/zed_extension_api/trait.Extension.html#method.language_server_additional_workspace_configuration - would be curious whether you tried that out at all.


Again, really sorry I have to close all these, but I'd love to discuss possible ways forward with you here in the thread. Thank you!

Tiltann:
> Hey @Tiltann, thanks for this (and all of these submissions) and apologies for the delayed feedback! I really appreciate you taking the time to get these up and the attempt to improve the ecosystem!
> 
> Unfortunately, at this time, I'll very sadly need to close all of these - not because I think these are bad or anything, but they will not work in a released environment. This was something you could not have known, and I'll make sure to finally get rid of this descrepancy between dev extensions and released extensions in the near future.
> 
> The issue being, that all of your extensions rely on the `definitions` being present - however, when we compile the extension, this will no longer be the case, as we do not copy that directory under any circumstances. In development, howver, it is present as you'd expect it to be. This happens because we treat dev extensions differently while we really should not - I'll personally make sure to fix this soon in the future.
> 
> However, this results in your extensions sadly not being ready to be published at this stage - I am sorry for that. You can get some of these up again by using our `download_file` API (https://docs.rs/zed_extension_api/latest/zed_extension_api/fn.download_file.html), so that you can make the files available again.
> 
> Aside from this biggest issue, two more things
> - I think we'd need to update the IDs to align better with our 
> [extension publishing prerequisites](https://zed.dev/docs/extensions/developing-extensions#extension-publishing-prerequisites) - that said, I am not even sure how precisely, since this is and was somewhat a first. But I think just `intellisense` might not be the best, but happy to hear your opinion
> - I doubt that `workspace_configuration` is the right API here and curious whether we should not instead use https://docs.rs/zed_extension_api/latest/zed_extension_api/trait.Extension.html#method.language_server_additional_workspace_configuration - would be curious whether you tried that out at all.
> 
> 
> Again, really sorry I have to close all these, but I'd love to discuss possible ways forward with you here in the thread. Thank you!



Heyo! 

About the late reply: take your time, quality over quantity and this response Shows that u value quality

Thank you for the detailed Response, this  Feedback  is awesome to hear from a beginners Standpoint. and i'll fix and reapply them when i get to it. 

This was very helpful!

MrSubidubi:
Appreciate the kind words, my pleasure! Looking forwards to that

PR #6542, opened 2026-06-19
Update yulang to v0.0.10

momota1029:
## Summary

- update the `yulang` extension submodule to `momota1029/yulang-zed@4855fb7`
- bump the registry version from `0.0.3` to `0.0.10`

## Notes

This picks up the Yulang language server integration and recent fixes so the extension defaults to launching `yulang server`.

## Checks

- `npx pnpm@11.1.0 sort-extensions`
- `npx pnpm@11.1.0 build`

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -5137,7 +5137,7 @@ version = "0.0.1"
 
 [yulang]
 submodule = "extensions/yulang"
-version = "0.0.3"
+version = "0.0.10"
 
 [zabby]
 submodule = "extensions/zabby"


momota1029:
The package check passed. The Danger failures look unrelated to this PR: both failed while fetching PR files/commits through `danger-proxy.zed.dev` with `ERR_STREAM_PREMATURE_CLOSE`.

MrSubidubi (changes requested):
Can you please 

- remove the legacy ID, since there exists no prior language server
- remove all the windows-specific sheanigans in regards to the paths? We run this within an WASM environment, so you may assume everything used there is POSIX-path style, no backslashes
- remove the `std::env::var` stuff, since this will not work. If you _really_ need env variables, please use `Worktree::shell_env` instead

Thanks!

momota1029:
Thanks! I removed the legacy language server ID, dropped the Windows-specific path handling and local env-var based fallback lookup, and now rely on the worktree environment via worktree.which("yulang"). The README was updated to match that behavior.

MrSubidubi (approved):
Thanks!

PR #6555, opened 2026-06-20
Add vercel-geist-theme

pol-cova:
Adds `vercel-geist-theme` as a new Zed theme extension.

This theme includes:

- `Vercel Geist Light`
- `Vercel Geist Dark`

The extension is added as a Git submodule under `extensions/vercel-geist-theme` and registered in `extensions.toml` with version `0.0.1`.

Theme repository:
https://github.com/pol-cova/vercel-theme

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4794,6 +4794,10 @@
 	path = extensions/vercel-carbon-theme
 	url = https://github.com/SwiftlyDaniel/zed-vercel-theme.git
 
+[submodule "extensions/vercel-geist-theme"]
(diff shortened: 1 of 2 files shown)


MrSubidubi (changes requested):
Thanks for this!

pol-cova:
Thanks for pointing this out!

I wasn’t aware there were already existing ports when I submitted these. I mainly made them for my own use, and now that I know I can keep a custom version locally, I’m happy to close the PR.

Thanks! :)

PR #6565, opened 2026-06-21
Add One Dark Vibrant theme

julian-graf:
Adds a color theme extension with bright and vibrant colors.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -3378,6 +3378,10 @@
 	path = extensions/one-dark-pro-vivid-theme
 	url = https://github.com/navinpeiris/zed-one-dark-pro-vivid.git
 
+[submodule "extensions/one-dark-vibrant-theme"]
+	path = extensions/one-dark-vibrant-theme
+	url = https://github.com/julian-graf/zed-one-dark-vibrant.git
+
 [submodule "extensions/one-hunter"]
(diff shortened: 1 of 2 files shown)


MrSubidubi (changes requested):
Thanks for this!

It seems that for your theme, we already have an existing ports. Could you please elaborate on why we should publish your theme as a separate extension to the already existing themes and explain where your theme differs from the existing themes/why your changes do not make sense to contribute upstream? 

Furthermore, please make sure your extension fulfills the [extension publishing prerequisites](https://zed.dev/docs/extensions/developing-extensions#extension-publishing-prerequisites). Particularly, you will need to update your extension ID as outlined in the documentation. Once you did the necessary updates, please update everything related here. Thanks!

julian-graf:
Not sure which other theme ports one dark vibrant over to Zed, could you point me to a specific one? I've checked out similarly named ones, but imho they seem to be different enough from this one. These extensions probably wouldn't accept contributions altering their theme that drastically.

I hope I got all the formalities right now, sorry for that inconvenience!

MrSubidubi (approved):
Yep, you are right - no one ported that theme. I defaulted to that question after reading "One Dark" 😅 

Looks indeed good now, thanks!

PR #6599, opened 2026-06-23
Update Svelte Effect Runtime to v3.0.2

sandersonstabo:
Updates the Svelte Effect Runtime Zed extension to 3.0.1.

Changes:
- Bumps the registry entry from 2.3.0 to 3.0.1.
- Moves the submodule to usebarekey/svelte-effect-runtime@8ffedd5.
- Installs the SER language server from npm instead of relying on a bundled GitHub release asset.
- Declares the Rust extension lib metadata and keeps the module-local BSD-3-Clause license.

Validation run locally:
- corepack pnpm build
- corepack pnpm test
- cargo check --target wasm32-wasip2 in modules/svelte-effect-runtime-zed
- cargo build --target wasm32-wasip2 --release in modules/svelte-effect-runtime-zed

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -4383,7 +4383,7 @@ version = "0.2.11"
 [svelte-effect-runtime-language-server]
 submodule = "extensions/svelte-effect-runtime-language-server"
 path = "modules/svelte-effect-runtime-zed"
-version = "3.0.1"
+version = "3.0.2"
 
 [svelte-mcp]
 submodule = "extensions/svelte-mcp"
--- a/extensions.toml
+++ b/extensions.toml
@@ -4383,7 +4383,7 @@ version = "0.2.11"
 [svelte-effect-runtime-language-server]
 submodule = "extensions/svelte-effect-runtime-language-server"
 path = "modules/svelte-effect-runtime-zed"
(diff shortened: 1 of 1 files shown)


MrSubidubi (commented):
Do we need `RUNTIME_ENTRYPOINT_COMPATIBILITY_SHIM`? No other extension does this and I'd rather have it removed here if possible, kept if it has a compelling reason for being there

sandersonstabo:
> Do we need `RUNTIME_ENTRYPOINT_COMPATIBILITY_SHIM`? No other extension does this and I'd rather have it removed here if possible, kept if it has a compelling reason for being there

Its left there as a compat fallback for locally built Zed extensions but I can remove it

sandersonstabo:
Updated this to v3.0.2 and removed the Zed-side runtime entrypoint shim from the submodule. The 3.0.2 language-server npm package now includes the top-level runtime/transform.js entry directly, so the extension no longer needs to create RUNTIME_ENTRYPOINT_COMPATIBILITY_SHIM locally.\n\nFresh PR checks are green.

MrSubidubi (approved):
Appreciate it, thanks!

PR #6606, opened 2026-06-24
ci: Add workaround for Danger flakiness

MrSubidubi:
We started seeing some flakiness in the danger checks and have been unable to determine the root cause in a brief analysis.

Since we currently do not have the time to dig into this deeper, this PR adds a retry policy to that job to instead retry the job on failure. 

Importantly, this will not influence "normal" danger fails - when there are danger checks failing, the step itself will succeed, instead danger will post a failing check to the check suites API.


maxdeviant:
I think we need to investigate further. See Slack for details.

MrSubidubi:
Closing as the issue is with node and fixed as of 24.18

PR #6610, opened 2026-06-24
Update slint extension

ogoffart:


MrSubidubi:
Hm the queries still do not seem to work. Please test this locally, happy to look at another update that fixes things and compiles.

PR #6563, opened 2026-06-21
Update todo-highlight-language-server to v0.2.2

shionit:
## Summary
- Updates `todo-highlight-language-server` submodule to `v0.2.2` and bumps `extensions.toml`.
- Linux binaries switched to fully-static musl targets (`*-unknown-linux-musl`), fixing launch failure on NixOS, Alpine, and other non-FHS distros that lack a system dynamic linker (follow-up to the issue reported in v0.2.1).
- Archive format changed to `.tar.gz` (macOS/Linux) / `.zip` (Windows), aligning with Zed extension best practices.
- Old version directories are cleaned up on upgrade.

## Test plan
- [x] `cargo check -p todo-highlight-extension --target wasm32-wasip1` passes
- [ ] Confirmed working on NixOS by original reporter

PR #6674, opened 2026-06-29
Update laravel to v0.6.0

mikebronner:
Bumps `laravel` to v0.6.0.

Release notes: https://github.com/mike-bronner/zed-laravel/releases/tag/0.6.0

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -2356,7 +2356,7 @@ version = "1.0.3"
 
 [laravel]
 submodule = "extensions/laravel"
-version = "0.5.0"
+version = "0.6.0"
 
 [latex]
 submodule = "extensions/latex"


MrSubidubi (changes requested):
Somewhat curious why this is needed, since this is the only language server I know of that needs a distinct binary for musl and not the biggest fan of checking the outside environment for this. Somewhat curious whether this could not be solved differently? But can worst case merge as is

mikebronner:
> Somewhat curious why this is needed, since this is the only language server I know of that needs a distinct binary for musl and not the biggest fan of checking the outside environment for this. Somewhat curious whether this could not be solved differently? But can worst case merge as is

@MrSubidubi Thanks for the review. :) This was my thinking behind it:

**Why the musl binary exists:** [mike-bronner/zed-laravel#236](https://github.com/mike-bronner/zed-laravel/issues/236) — a user working in an Alpine devcontainer (`dunglas/frankenphp:php8.5-alpine`) via Zed remote development. In that setup the remote server runs the extension and spawns the LSP *inside* the container, and the glibc-linked binary can't start on musl ("No such file or directory (os error 2)" — the glibc loader doesn't exist there). Alpine images are a mainstream base for PHP/Laravel containers, so this hits real users; the same loader problem breaks the gnu binary on NixOS. The fix in [mike-bronner/zed-laravel#237](https://github.com/mike-bronner/zed-laravel/pull/237) added musl builds alongside the gnu ones, selected at runtime.

**Why the environment probing:** `zed::current_platform()` exposes OS and arch but not libc, and `cfg!(target_env)` is useless in a wasm32 extension — so the filesystem probe was the only detection available. For what it's worth it mirrors what Zed core itself does to pick rust-analyzer's `gnu` vs `musl` release asset ([`crates/languages/src/rust.rs`](https://github.com/zed-industries/zed/blob/main/crates/languages/src/rust.rs) parses `ldd --version` and falls back to scanning `/lib` for `ld-musl-*`).

**Cleaner option?** — would it be preferable to use musl across the board and ship a single statically-linked musl binary per Linux arch and delete the detection entirely? A static musl binary runs on glibc distros, Alpine, and NixOS alike. That's exactly what Zed's own remote server does for every Linux host, and what gleam, tombi, codebook, and todo-highlight (#6563) already do for their language servers. laravel-lsp doesn't `dlopen` anything (the one reason rust-analyzer keeps a glibc build), so nothing blocks it here.

I could drop the gnu builds and the runtime detection, shipping only `laravel-lsp-linux-{x64,arm64}` as static musl (with mimalloc as the global allocator to offset musl's slower multithreaded malloc). Is that a better solution?

I spun up a PR with a possible solution: https://github.com/mike-bronner/zed-laravel/pull/258

mikebronner:
@MrSubidubi I went ahead and merged and deployed that PR. I think that's obviously the better way to go: not multiple Linux builds and no external looking into the operating system.

MrSubidubi (approved):
Sorry for the delay, but yeah, that indeed looks a lot better. Thank you!

PR #6678, opened 2026-06-29
Add inlang language server

NEKOYASAN:
Add Inlang language server support

This extension provides language-server support for Inlang and Paraglide message references, including inline hints, hovers, diagnostics, reference navigation, and code lenses.

This extension was developed inspired by [opral/sherlock](https://github.com/opral/sherlock), which is Inlang's VS Code extension. Since Sherlock does not support Zed, this extension was created to make similar inspections available in Zed.

inline hint & hovers
image

diagnostics
image

reference navigation
image

Code lens 
image

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2094,6 +2094,10 @@
 	path = extensions/inko
 	url = https://github.com/inko-lang/zed.git
 
+[submodule "extensions/inlang-language-server"]
+	path = extensions/inlang-language-server
+	url = https://github.com/NEKOYASAN/inlang-zed
+
 [submodule "extensions/intellij-light-theme"]
 	path = extensions/intellij-light-theme
 	url = https://github.com/chandruscm/intellij-light-theme-zed.git
--- a/extensions.toml
+++ b/extensions.toml
@@ -2136,6 +2136,10 @@ version = "0.0.2"
 submodule = "extensions/inko"
 version = "0.0.1"
 
+[inlang-language-server]
+submodule = "extensions/inlang-language-server"
+version = "0.0.1"
+
 [intellij-light-theme]
 submodule = "extensions/intellij-light-theme"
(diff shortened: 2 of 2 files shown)


MrSubidubi (changes requested):
Thanks for this!

NEKOYASAN:
@MrSubidubi Thank you for your review!

Sorry about that. Since this extension also provides code actions and a few editor features beyond just starting the language server, I thought the shorter `inlang` ID would be acceptable.

I’ve updated it to be more explicit for the publishing requirements: the extension ID is now `inlang-language-server`, and I also updated the related manifest entries, README settings key, and Cargo metadata to match.

MrSubidubi:
No worries, happens to many people here. Asked for the lsp-related suffix since it is just a language server at its core currently.

Curious though - could you briefly explain why you requested the other ID ealier?

NEKOYASAN:
@MrSubidubi Thank you!

I originally named it `inlang`, following the example of extensions like `oxc` (`oxlint`, `oxfmt`). Since the extension provides various language server features, including diagnostics, hovers, and inlay hints, I mistakenly thought the shorter name would be appropriate.

However, given that it remains a language server at its core, I think your point that it should include the `-language-server` suffix is completely fair.

MrSubidubi:
> Since the extension provides various language server features, including diagnostics, hovers, and inlay hints, I mistakenly thought the shorter name would be appropriate.

Makes sense - since adding these, we updated our policies to be a bit more strict as we want to keep the top level namespace free for languages mostly (as in, make it more clear what you are installing by just looking at the ID). We allow some more popular extensions for language servers to keep their top level namespace (oxlint probably being a good example), but generally try to enforce this policy nowadays.

MrSubidubi (approved):
Looks good now, thanks!

PR #6676, opened 2026-06-29
Update cem to v0.11.0

bennypowers:
Updates cem extension to [version 0.11.0](https://github.com/bennypowers/cem/releases/tag/v0.11.0).

PR #6682, opened 2026-06-30
Update cem to v0.11.0

bennypowers:
Updates the cem submodule to v0.11.0, which includes a fix for the
`zed_extension_api` 0.7.0 breaking change (`Project::worktrees()` removed,
replaced by `worktree_ids()`).

Supersedes #6676 which failed CI due to this API change.


MrSubidubi (approved):
Note that at some point soon, we will start migrating MCP extensions to use the MCP registry instead, so please be aware of that change. Thank you!

PR #6700, opened 2026-07-02
Add orng-theme

aefly:
- [X] I've read [CONTRIBUTING.md](../CONTRIBUTING.md) and followed the relevant guidance for adding or updating my extension.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -3314,6 +3314,10 @@
 	path = extensions/orng
 	url = https://github.com/elithrar/orng-zed.git
 
+[submodule "extensions/orng-theme"]
+	path = extensions/orng-theme
+	url = https://github.com/aefly/orng
+
 [submodule "extensions/osaka-jade-theme"]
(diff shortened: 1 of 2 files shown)


MrSubidubi (changes requested):
Thanks for this!

aefly:
Thanks so much for taking a look! Happy to clarify 🙂

`orng-zed` (published as `orng`) is a separate theme inspired by Cloudflare's workers.dev color scheme. I'm not affiliated with that project. `orng-theme` uses my own distinct palette and syntax highlighting scheme; it's not derived from `orng-zed`, just a different variant that happens to land on a similar "orange" name.

That said, I totally get why the naming could cause confusion sitting next to `orng` in the gallery,  happy to rename mine to something more distinct if that'd help! Let me know what you'd prefer and I'll get it updated.

MrSubidubi (approved):
Do feel free to change the name, but happy to ~~green~~ orangelight it for now. 

Thank you!

PR #6709, opened 2026-07-02
Update shader-ls to v0.2.0

cjhowe-us:
Uses the Zed LSP configuration API and adds grammars for HLSL, GLSL, and WGSL thanks to @Inflation!

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -4095,7 +4095,7 @@ version = "0.1.0"
 
 [shader-ls]
 submodule = "extensions/shader-ls"
-version = "0.1.0"
+version = "0.2.0"
 
 [shades-of-purple-theme]
 submodule = "extensions/shades-of-purple-theme"


MrSubidubi (changes requested):
This violates our [extension publishing prerequisites](https://zed.dev/docs/extensions/developing-extensions#extension-publishing-prerequisites) I believe, why do you need the langauges for the language definitions for the server to work with these? Could users not just install the existing extension?

cjhowe-us:
Yes, that's the current recommendation. I'll revert that change, then update the PR.

inflation:
This was for feature parity with the vscode extension from the lsp author. I'm fine if you'd need to get rid of the syntax.

cjhowe-us:
I've removed the grammars and updated the code. PTAL

MrSubidubi (approved):
Thanks!

PR #6732, opened 2026-07-06
Add CodeGraph MCP server

sunerpy:
Adds the **CodeGraph** extension, which provides [CodeGraph](https://github.com/sunerpy/codegraph-rust) as an MCP context server inside Zed.

CodeGraph is a deterministic tree-sitter + SQLite/FTS5 **code knowledge graph**. It parses a codebase, extracts symbols and their call/dependency relationships, and answers structural questions ("who calls X", "what breaks if I change X", "where is X defined", "how does this area work") in one query instead of dozens of grep + file reads. There is no AI/LLM/vector store inside the binary — output is fully deterministic.

The extension registers a `codegraph` context server. On launch it resolves the latest GitHub release of `sunerpy/codegraph-rust`, downloads the platform binary, and caches it under a version-stamped path (auto-updates on new releases); a user-provided `context_servers.codegraph.command` in Zed settings overrides the download.

- **Repository:** https://github.com/sunerpy/codegraph-rust
- **Extension path:** `editors/zed` (monorepo subdirectory; submodule pinned to tag `v0.25.1`)
- **License:** MIT (`editors/zed/LICENSE`)

### Checklist
- [x] Read the Extension Publishing Prerequisites and Publishing guide.
- [x] Tested locally as a dev extension before submitting.
- [x] The extension repository includes an accepted license (MIT).
- [x] No existing extension provides this functionality (searched `extensions.toml` — no `codegraph`/`code-graph` entry).
- [x] Added the submodule and an `extensions.toml` entry in alphabetical order.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -778,6 +778,10 @@
 	path = extensions/codebook
 	url = https://github.com/blopker/codebook-zed.git
 
+[submodule "extensions/codegraph"]
+	path = extensions/codegraph
+	url = https://github.com/sunerpy/codegraph-rust.git
+
 [submodule "extensions/codely-theme"]
 	path = extensions/codely-theme
 	url = https://github.com/GOI17/codely-theme-zed
--- a/.gitmodules
+++ b/.gitmodules
@@ -778,8 +778,8 @@
 	path = extensions/codebook
 	url = https://github.com/blopker/codebook-zed.git
 
-[submodule "extensions/codegraph"]
-	path = extensions/codegraph
+[submodule "extensions/codegraph-mcp"]
+	path = extensions/codegraph-mcp
 	url = https://github.com/sunerpy/codegraph-rust.git
 
 [submodule "extensions/codely-theme"]
--- a/extensions.toml
+++ b/extensions.toml
@@ -791,6 +791,11 @@ version = "0.0.5"
 submodule = "extensions/codebook"
 version = "0.2.14"
 
+[codegraph]
+submodule = "extensions/codegraph"
+path = "editors/zed"
+version = "0.1.0"
+
 [codely-theme]
 submodule = "extensions/codely-theme"
 version = "0.0.1"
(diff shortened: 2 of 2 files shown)


sunerpy:
@cla-bot check

sunerpy:
@cla-bot check

MrSubidubi (changes requested):
Thanks for this!

Also, I think it would be good if you could implement auto-cleanup for old binaries found in the environment on update, so that the cache for the extension does not grow too much. Thanks!

sunerpy:
Thanks! Both done.

**ID → `codegraph-mcp`.** The docs phrase the rule as a suffix (`-theme`, `-snippets`), and the registry already has 14 `*-mcp` entries — `azure-mcp`, `sentry-mcp`, `prisma-mcp`, `chrome-devtools-mcp` — so `-mcp` felt like the natural fit while keeping the product name first. Happy to switch to `mcp-server-codegraph` if you'd rather have the explicit `mcp-server` form.

The `[context_servers.codegraph]` key inside `extension.toml` stays as-is — it's what our installer and existing user configs already write, and it doesn't need to match the extension id (`chrome-devtools-mcp` does the same).

**Auto-cleanup.** Implemented. The extension resolves the latest release on each launch and caches it under `codegraph-<version>/`, so the work dir used to keep every version it had ever downloaded. It now removes superseded `codegraph-*` dirs, same shape as the `proto` / `test-extension` helpers in `zed-industries/zed`:

- only after a successful download, so a cache hit never reaps anything;
- prefix-scoped, and the current version is always kept;
- deletion is best-effort — an in-use `.exe` on Windows shouldn't be able to fail a launch.

**On `context_server_configuration`:** codegraph needs no keys or tokens — it indexes the local project and talks stdio — so there's nothing to prompt for. Anyone who wants a specific binary or a `--path` can override `context_servers.codegraph.command`. Glad to add an implementation for the installation-instructions surface if that's useful.

Registry side is updated here: `extensions/codegraph-mcp`, `[codegraph-mcp]`, version `0.2.1`, submodule pinned to v0.43.1. I also went back over the other prerequisites — the MIT `LICENSE` sits at the extension `path` (not just the repo root), all manifest fields are filled, and no build output is tracked.

PR #5530, opened 2026-04-06
Add RMarkdown language support

scfurl:
## Summary
- Adds R Markdown (`.Rmd`) language support for Zed
- Provides Markdown syntax highlighting for `.Rmd` files using the existing `markdown` tree-sitter grammar
- Injects R language highlighting into knitr-style ```` ```{r} ```` and ```` ```{r, opts} ```` code chunks
- Supports YAML/TOML frontmatter and fenced code block background highlighting

Closes #1120

## Extension repo
https://github.com/scfurl/rmarkdown

PR #6740, opened 2026-07-06
Update r to v0.2.6

ocsmit:
Release notes:

https://github.com/ocsmit/zed-r/releases/tag/v0.2.6

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -3837,7 +3837,7 @@ version = "0.2.2"
 
 [r]
 submodule = "extensions/r"
-version = "0.2.5"
+version = "0.2.6"
 
 [racket]
 submodule = "extensions/racket"


MrSubidubi (changes requested):
Could you elaborate what the markdown grammar is used for? If that is only for highlighting, can we not just add the suffixes upstream instead?

aymennasri:
@alexdhill @ocsmit can answer you here, also Alex it seems like someone else made a PR [here](https://github.com/zed-industries/extensions/pull/5530) outside of the extension; could you check it out and see if there's anything worth adding?

alexdhill:
Looks like his RMarkdown language is exactly the same highlighting I added: markdown with injections for the preamble and code fences.

To answer about *why* the core is markdown highlighting, I had originally tried to just use the native MD language and inject in the preamble/code fences but that didn't work. I think there was some issue trying to use an existing language as the core of another and not just as an injection.

aymennasri:
@MrSubidubi

MrSubidubi:
Question from me here would be:
- Should we not instead pursue the other extension instead of this or should it be the other way around?
- Is there any reason we cannot just update the core markdown language?

alexdhill:
I think the main reason to have it added here is that the core addition of rmarkdown is using the R code blocks - which require the R language highlighting, which is implemented in this extension. Any user who wants rmd will likely need to have this extension installed anyway. I know there was discussion at #25758 of adding R highlighting native, but AFAIK thats not on the roadmap. @MrSubidubi

PR #6765, opened 2026-07-09
Add Modern SCSS Modules extension

emilgerz:
Adds Modern SCSS Modules, a focused Zed extension that provides go-to-definition support from React CSS module usages like `styles.root` to matching selectors in `.module.scss`, `.module.sass`, and `.module.css` files.

The extension has been tested locally as a dev extension.


emilgerz:
@cla-bot check

PR #6778, opened 2026-07-10
Add Monochrome Amber theme

yusefkarim:
Adds my "Monochrome Amber" theme (https://github.com/yusefkarim/zed-monochrome-amber/tree/main)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2898,6 +2898,10 @@
 	path = extensions/monochromator-theme
 	url = https://codeberg.org/beem/monochromator.git
 
+[submodule "extensions/monochrome-amber-theme"]
+	path = extensions/monochrome-amber-theme
+	url = https://github.com/yusefkarim/zed-monochrome-amber
+
 [submodule "extensions/monokai-nebula"]
 	path = extensions/monokai-nebula
 	url = https://github.com/JacobCallahan/monokai-nebula-zed.git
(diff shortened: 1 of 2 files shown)


yusefkarim:
@cla-bot check

yusefkarim:
@cla-bot check

MrSubidubi (changes requested):
Thanks for this!

yusefkarim:
Hey @MrSubidubi,

Thanks for considering my PR. I looked at the currently available extensions and see "Amber Monochrome Monitor CRT Phospor Theme" and "Monochromator" being the two similar themes.

* Amber Monochrome Monitor CRT Phospor Theme: despite the similar name, it is quite different: foreground  colors are literally ALL amber color, background is a deep black or orange-tinted
* Monochromator: pretty dang close and it's what I used as a starting point for my theme actually. The primary difference is the background for Monochromator is very dark, where mine is lighter (easier on the eyes in my opinion). My theme also has some additional syntax highlighting for symbol and what not.
  - I don't think contributing my lighter background upstream makes much sense since it's a personal preference
  - I could upstream my additional syntax highlighting, but again not sure if that would be wanted/accepted

All-in-all, if you think this is too close to Monchromator, I'm fine just closing this without merging - I can just add it manually as I do today - no worries!

PR #6844, opened 2026-07-16
Add One Dark Pro GNOME theme

prabhjot0109:
This PR adds the One Dark Pro GNOME variant theme extension to the registry, ported from the popular VS Code theme.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -3306,6 +3306,10 @@
 	path = extensions/one-dark-pro-enhanced
 	url = https://github.com/hadez8877/one-dark-pro-enhanced.git
 
+[submodule "extensions/one-dark-pro-gnome-theme"]
+	path = extensions/one-dark-pro-gnome-theme
+	url = https://github.com/prabhjot0109/zed-one-dark-pro-gnome-theme.git
+
 [submodule "extensions/one-dark-pro-max"]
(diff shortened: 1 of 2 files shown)


prabhjot0109:
@cla-bot check

MrSubidubi (changes requested):
Thanks for this!

prabhjot0109:
To clarify — One Dark Pro isn't a single theme; it includes several official variants in VS Code, such as Vivid, Flat, Mix, Darker, and GNOME. These variants differ not only in accent/syntax colors, but also in their background and overall UI palette.

The GNOME variant uses a much darker, neutral background instead of the blue-grey `#282c34` used by the standard One Dark variant that the existing ports appear to be based on. This gives it a noticeably different overall appearance and contrast, without the stronger blue-ish cast of the standard background.

The syntax colors remain familiar One Dark colors, but the background and UI/chrome palette are different enough that this isn't simply a small modification to the existing extension.

The motivation for this PR is specifically to bring the official One Dark Pro GNOME variant, which is available in VS Code, to Zed. I see it as another distinct variant within the One Dark Pro family rather than a duplicate of the existing ports.

I can also add side-by-side screenshots comparing the existing One Dark port with the GNOME variant if that would help.

PR #6858, opened 2026-07-18
Update kulala-http

HusuSama:
Major updates for Kulala:

> [!IMPORTANT]
> Updating to the current version will fully support all the capabilities of `kulala-core`, but for requests sent via `lsp-action`, you will need to automatically create a `.kulala-cache` folder in the current project to save request history. Historical data can be automatically cleaned using `lsp-action` or scheduled for requests exceeding seven days

1. Switch to `kulala-core` and no longer use `kulala-ls`
2. Internally set up a `lsp` relay
3. Requests can be sent directly internally without relying on other third-party tools
4. Better Completion Experience

> [!CAUTION]
> The currently tracked branch will switch to version `v2`, and future updates will continue on the `v2` branch

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2329,6 +2329,7 @@
 [submodule "extensions/kulala-http"]
 	path = extensions/kulala-http
 	url = https://github.com/HusuSama/zed-kulala-http.git
+	branch = v2
 
 [submodule "extensions/kvs-cyberpunk-2077"]
 	path = extensions/kvs-cyberpunk-2077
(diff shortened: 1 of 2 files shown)


MrSubidubi:
Thanks for this and sorry for letting this sit for so long, slipped through the cracks!

However, I'll have to reject the update at this time - this violates our 
[extension publishing prerequisites](https://zed.dev/docs/extensions/developing-extensions#extension-publishing-prerequisites) in that you'll have to ship the JavaScript file through NPM if you want to use it. The proxy itself is not an issue, just how it is currently invoked. 

Please throw us a new PR with that fix and I'll promise to get back to you sooner there.


Also, did not forget about the thrift extension! I paved the way for this some time ago, howevever, it is currently blocked on an internal review for some code I readied for this. I'll ping people internally soon again, sorry for the wait there!

PR #6899, opened 2026-07-21
Update blade to v0.2.7

bajrangCoder:
Release notes:

https://github.com/bajrangCoder/zed-laravel-blade/releases/tag/v0.2.7

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -460,7 +460,7 @@ version = "0.1.1"
 
 [blade]
 submodule = "extensions/blade"
-version = "0.2.6"
+version = "0.2.7"
 
 [blade-runner-2049]
 submodule = "extensions/blade-runner-2049"


MrSubidubi (changes requested):
A bit curios why this mirrors the LSPs from the PHP extension, not complaining or anything, but would be interested in the reason behind it

bajrangCoder:
> A bit curios why this mirrors the LSPs from the PHP extension, not complaining or anything, but would be interested in the reason behind it

First, whenever something new is added to the PHP extension related to lsp, users eventually come to the Blade repository asking for the same support. And I added the Phantom LSP because it is fast, works well, and supports Blade out of the box.

Initially, I tried reusing the PHP language server. Reference: https://github.com/zed-extensions/php/issues/30

MrSubidubi:
Thanks for referencing that, was not aware. Curious though why you went against reusing the language server? 

Overall the policy we try to enforce nowadays is to reduce duplication, that said, in your case I'd be happy to make an exception. However, would like to find a way forward that works best for both sides

bajrangCoder:
What would be the best way? I'm happy to go with that.
I also don't like these repetition.

MrSubidubi:
I presume that it would probably be best if we were to add the Blade language IDs to the PHP extension? So that the language servers are associated with Blade.

We can then also tweak the defaults in the Zed config so that only one server is enabled for Blade by default. I am currently working to improve that this also can be configured in extensions, but not quite there yet.

bajrangCoder:
@MrSubidubi should I open pr in php extension or something? Cuz people's want that new PHPantom LSP in blade files([ref](https://github.com/bajrangCoder/zed-laravel-blade/issues/32#issuecomment-5260182952))

MrSubidubi:
Yeah, please. I can take a look, but am basically reviewing 24/7 currently, so if you would have the time, I'd gladly take a look at it and merge ASAP. If I am to take a look myself, it would take some more time and can't make any promises sadly, sorry.

MrSubidubi:
Hey, just checking in here, should we keep the PR open for the moment or how would you like to proceed here?

bajrangCoder:
Sorry for the delay. I’ve opened a PR on the PHP extension repo: https://github.com/zed-extensions/php/pull/149

Once that’s merged, I’ll publish the Blade extension update, which will remove the duplicate LSPs.

PR #6942, opened 2026-07-24
Update ObjectScript to v1.5.0

hkimura-intersys:
This update adds the objectscript langauge server
to the extension. Additionally, highlighting
updates were added and the commit for the
objectscript grammars was updated to the
most recent one. Details on the language server features
are described [here](https://github.com/intersystems/zed-objectscript/pull/9#issue-4675257253).

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -3340,7 +3340,7 @@ version = "1.0.1"
 
 [objectscript]
 submodule = "extensions/objectscript"
-version = "1.4.0"
+version = "1.5.0"
 
 [obsidian-sunset]
 submodule = "extensions/obsidian-sunset"


MrSubidubi (changes requested):
Any particular reason that we still ship XML with this given that you now renamed the language? Is this because you need the right LSP IDs? That can be configured via `language_ids` on a per-language-server basis in the extension.toml

hkimura-intersys:
> Any particular reason that we still ship XML with this given that you now renamed the language? Is this because you need the right LSP IDs? That can be configured via `language_ids` on a per-language-server basis in the extension.toml

I still shipped it as a part of this because it has injections added to it that are specific to ObjectScript. So, not related to the language server but I thought it was still needed for the highlighting changes to show up.

hkimura-intersys:
> > Any particular reason that we still ship XML with this given that you now renamed the language? Is this because you need the right LSP IDs? That can be configured via `language_ids` on a per-language-server basis in the extension.toml
> 
> I still shipped it as a part of this because it has injections added to it that are specific to ObjectScript. So, not related to the language server but I thought it was still needed for the highlighting changes to show up.

I would add it to the zed-xml extension instead, but it seems that repo is stale (hasn't had any of the open PRs merged in months, and is still based on a stale commit), so I wanted to make sure it got in on time.

PR #6983, opened 2026-07-27
Update caddyfile to v0.0.6

### LINKED ISSUE #6: Multiline comma-separated site addresses break highlighting
When a Caddyfile block is opened and defined by multiple comma-separated web addresses across several lines, highlighting breaks, and remains incorrect for the rest of the file.

I'm unsure if this is an issue with the extension's implementation or the upstream tree-sitter-caddyfile repo.

**Expected:**
The comma-separated web addresses on new lines would highlight correctly, the same as if they were on a single line.

**Actual:**
The first address is highlighted correctly, but everything from then on is not.

**Repoduction:**
```caddyfile
mysite.com,
myothersite.com {
    # From here to the end of the file, highlighting will be broken.
}
```

**Screenshot:**
Image of the problem.

### LINKED ISSUE #7: Regular expressions with parentheses, brackets, or pipe characters break highlighting
When a matcher is defined using a regular expression, it can break highlighting for the remainder of the file. In my testing, pairs of parentheses or brackets will cause problems, as will a lone `|` anywhere in the regex pattern.

For example, these lines:
```caddyfile
handle_path /working {
    file_server
}
	
@broken path_regexp /foo/(bar|baz).*
@also-broken path_regexp /foo/[0-9]+

handle /xyz {
    file_server
}
```

are rendered as:

Image

nusnewob:
- Fixes https://github.com/nusnewob/caddyfile-zed/issues/6
- Fixes https://github.com/nusnewob/caddyfile-zed/issues/7

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -582,7 +582,7 @@ version = "0.0.2"
 
 [caddyfile]
 submodule = "extensions/caddyfile"
-version = "0.0.5"
+version = "0.0.6"
 
 [cadence]
 submodule = "extensions/cadence"


MrSubidubi (approved):
Thanks!

PR #7006, opened 2026-07-29
Update XML to v0.1.1

sweetppro:
* Update tree-sitter-xml commit [fix(scanner): correct UB when deleting old tag Strings]
* Indent element children after opening tag
co-authored by @pgundlach


MrSubidubi (approved):
Thanks!

PR #7062, opened 2026-08-04
Add Windows Batch language support

pleahmacaka:
Windows Batch (.bat, .cmd) syntax highlighting via [`wharflab/tree-sitter-batch`](https://github.com/wharflab/tree-sitter-batch). LSP via [`rech-editor-batch`](https://github.com/RechInformatica/rech-editor-batch).

I've used this snippet for test:
```batch
@echo off
REM sample used to eyeball highlighting
setlocal enabledelayedexpansion

set "NAME=world"
set /a COUNT=1+2

if not "%NAME%"=="" (
    echo Hello %NAME% %CD% !COUNT!
) else (
    goto :fail
)

for /f "tokens=1,2" %%a in ('dir /b') do (
    call :log %%a
)

dir > nul 2>&1 && echo ok || echo failed

:log
echo [%TIME%] %~1
exit /b 0

:fail
exit /b 1
```

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -5214,6 +5214,10 @@
 	path = extensions/wikitext
 	url = https://github.com/Benjamin-Davies/zed-wikitext.git
 
+[submodule "extensions/windows-batch"]
+	path = extensions/windows-batch
+	url = https://github.com/pleahmacaka/zed-batch.git
+
 [submodule "extensions/wit"]
 	path = extensions/wit
 	url = https://github.com/valentinegb/zed-wit.git
--- a/extensions.toml
+++ b/extensions.toml
@@ -5308,6 +5308,10 @@ version = "0.1.0"
 submodule = "extensions/wikitext"
 version = "0.0.1"
 
+[windows-batch]
+submodule = "extensions/windows-batch"
+version = "0.2.0"
+
 [wit]
 submodule = "extensions/wit"
 version = "0.4.0"


ofek:
Thanks a lot! Out of curiosity, what is the process for natively adding support rather than as an extension?

pleahmacaka:
> Thanks a lot! Out of curiosity, what is the process for natively adding support rather than as an extension?

A maintainer could answer this better, but AFAIK it won't be added right now. The [language list](https://zed.dev/docs/languages) marks built-in ones with a `*`, and the list stays small on purpose: more built-in languages mean slower builds, a bigger binary, and more PRs for the team to review (zed-industries/zed#25758, zed-industries/zed#7096).

PowerShell is extension-only too, so moving Batch would really mean moving both, which would break extensions people already installed. The PowerShell maintainer should have a say there, so leaving it as is seems fine for now.

Would love to see it in core someday, though.

MrSubidubi (approved):
Another very very nice PR coming in from you here, thank you so much! Very happy to merge this!

Regarding the builtin support: These comments are also spot-on, however, we can at least improve the experience for installing the extension: There is an `extension_suggest` in the main repo (might differ slightly in name) that allows recommending extensions based on file extensions (we will improve this in the future, but not immediately).

I'd be very happy to accept a PR in the core repo to add the mapping for this extension.

In any case, that does not stop me from merging this one. Thank you so much!

pleahmacaka:
Thanks for merging! Followed up with the suggestion mapping in zed-industries/zed#62363.

PR #7066, opened 2026-08-04
Bump laravel to 0.7.1

mikebronner:
Bumps `laravel` to v0.7.1.

Release notes: https://github.com/mike-bronner/zed-laravel/releases/tag/0.7.1

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -2434,7 +2434,7 @@ version = "1.0.3"
 
 [laravel]
 submodule = "extensions/laravel"
-version = "0.6.1"
+version = "0.7.1"
 
 [laravel-official]
 submodule = "extensions/laravel-official"


mikebronner:
@MrSubidubi Before we merge this, I wanted to inquire if there is a way I can change the extension ID without breaking user settings and duplicate extension installs? I know the docs say this should not be done, I was hoping to keep the ID inline with the rest of the name refactor.

I'm trying to figure out a way to change the extension identifier from `laravel` to `laravel-ce`.

MrSubidubi:
Nope, we currently do not support changing an extensions ID after publishing. 

Please let me know if we should proceed with this PR anyway

mikebronner:
@MrSubidubi please go ahead with this PR. 
Would it be feasible to create a new extension under the same name and with a different ID? I guess it would be a one-time breaking change. I'll have a think about this. Hoping that this change will update everything else and obscure the actual id from the users.

MrSubidubi (approved):
Think I would not be a fan of it TBH, would be better if we just kept it this way for now. If we ever have the option to rename extensions we for sure can revisit this, but don't think there is much or any value in having two extensions that do precisely the same. Rather the opposite IMO

mikebronner:
> Think I would not be a fan of it TBH, would be better if we just kept it this way for now. If we ever have the option to rename extensions we for sure can revisit this, but don't think there is much or any value in having two extensions that do precisely the same. Rather the opposite IMO

Sounds good, my thinking was to retire the original and have it removed from the extensions list, then just register the new one. But that's fairly disruptive.

PR #7089, opened 2026-08-05
Add intellij-lsp extension

hlucas13:

## Description

Adds **IntelliJ LSP** — a language server extension that brings JetBrains IntelliJ IDEA's Java and Kotlin intelligence to Zed.

JetBrains [recently announced](https://blog.jetbrains.com/idea/2026/08/intellij-idea-goes-lsp/) their IntelliJ LSP server for third-party editors (currently previewing in VS Code/Cursor). This is an **unofficial community port** of that server to Zed, using the same LSP protocol.

**Zero setup required.** The extension automatically downloads and installs the IntelliJ LSP server on first launch. No scripts, no manual configuration.

Features: code completion, navigation, refactorings, inspections, and quick-fixes for Maven, Gradle, and Bazel projects. Multi-platform (macOS, Linux, Windows).

- **Repository**: https://github.com/hlucas13/intellij-lsp-zed
- **Languages**: Java, Kotlin
- **License**: MIT (extension code) / JetBrains EULA (server, downloaded at runtime — not shipped with the extension)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -5409,3 +5409,7 @@
 [submodule "extensions/zwirn"]
 	path = extensions/zwirn
 	url = https://codeberg.org/polymorphicengine/zwirn-zed-extension.git
+[submodule "extensions/intellij-lsp"]
+	path = extensions/intellij-lsp
+	url = https://github.com/hlucas13/intellij-lsp-zed
+	branch = main
--- a/extensions.toml
+++ b/extensions.toml
@@ -2148,6 +2148,10 @@ version = "0.0.1"
 submodule = "extensions/intellij-light-theme"
 version = "0.1.0"
 
+[intellij-lsp]
+submodule = "extensions/intellij-lsp"
+version = "0.1.0"
+
 [intellij-newui-theme]
 submodule = "extensions/intellij-newui-theme"
 version = "0.1.0"


MrSubidubi:
Thank you for this!

Unfortunately, we will not proceed with it at this time for the following reasons:
- https://open-vsx.org API use needs to be either approved or bought from what I could find whilst looking into this. While this might work for you using this privately, I suspect we as Zed Industries might run into issues/concurrency limits there if we were to ship this through our extension registry.
- JetBrains themselves prompt users to read and accept their EULA in other editors with the extension - we cannot just bypass this for users unprompted, as this is not an appropriate way to deal with an EULA. This is not a way to accept this after all
- While I definitely want and hope for us to support this at some point for our users, we'd need to double-check how we can ship this in a way that works for JetBrains. While I don't want to say we will only accept an extension coming from JetBrains, I suspect we might either want to pull it in zed-extensions or at least have some form of confirmation that we can ship it this way. This applies especially long-term when this will be a commercial product.

I consider especially the first two points a dealbreaker for this extension as is, and thus will close, as we will not be endorsing this extension officially at this point by publishing it. Hope you can undertand our reasoning here, in any case, thank you for taking the time to look into this and getting both the PR and extension up!

hlucas13:
Hi @MrSubidubi, thank you for the detailed review on #7089. Both points were well-taken and I've addressed them concretely.

1. **Open VSX API at runtime**: the extension no longer queries any registry API on a user's machine. The server binary is fetched from a pinned, per-version JetBrains CDN URL embedded at compile time.
2. **EULA silently bypassed**: there's now an explicit `accept_jetbrains_eula` setting gate; nothing downloads or runs without it, and it's re-checked on every launch.

The full technical breakdown of the new architecture is in the replacement PR #7130, I won't duplicate it here. On your third point (ideally a JetBrains-sanctioned integration), that's honestly outside what I can guarantee unilaterally, and the extension's README and THIRD_PARTY_NOTICES.md state the server's licensing status clearly.

Happy to answer questions on the new submission whenever you have time.

PR #7042, opened 2026-08-01
Add monokai-ignite theme

darkings:
Monokai Ignite is a blazing fast, warm amber accented Monokai theme tailored for maximum focus.

- Theme ID: `monokai-ignite`
- Version: 0.1.0
- License: MIT

PR #7092, opened 2026-08-06
Add monokai-ignite theme

darkings:
Adds the Monokai Ignite theme extension.

- Registers only `monokai-ignite-theme` at version `0.3.0`
- Adds its git submodule at commit `f5382b2c10fc4fa6d8d86c916797065dc9873a87`
- Replaces #7042, whose diff included unrelated extension version changes

Validated with `git diff --check`.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2982,8 +2982,8 @@
 	path = extensions/monochromator-theme
 	url = https://codeberg.org/beem/monochromator.git
 
-[submodule "extensions/monokai-ignite"]
-	path = extensions/monokai-ignite
+[submodule "extensions/monokai-ignite-theme"]
+	path = extensions/monokai-ignite-theme
 	url = https://github.com/darkings/monokai-ignite.git
 
 [submodule "extensions/monokai-nebula"]
--- a/.gitmodules
+++ b/.gitmodules
@@ -5409,3 +5409,6 @@
 [submodule "extensions/zwirn"]
 	path = extensions/zwirn
 	url = https://codeberg.org/polymorphicengine/zwirn-zed-extension.git
+[submodule "extensions/monokai-ignite"]
+	path = extensions/monokai-ignite
+	url = https://github.com/darkings/monokai-ignite.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -2982,6 +2982,10 @@
 	path = extensions/monochromator-theme
 	url = https://codeberg.org/beem/monochromator.git
 
+[submodule "extensions/monokai-ignite"]
+	path = extensions/monokai-ignite
+	url = https://github.com/darkings/monokai-ignite.git
+
 [submodule "extensions/monokai-nebula"]
 	path = extensions/monokai-nebula
 	url = https://github.com/JacobCallahan/monokai-nebula-zed.git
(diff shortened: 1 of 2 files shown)


MrSubidubi (changes requested):
Thanks for this!

Also, it seems that for your theme, we already have an existing ports. Could you please elaborate on why we should publish your theme as a separate extension to the already existing themes and explain where your theme differs from the existing themes/why your changes do not make sense to contribute upstream? Thanks!

darkings:
Thanks for the review! I've addressed both points.

## Extension ID

The extension ID has been updated to `monokai-ignite-theme` per the publishing prerequisites, and everything related has been updated to match:

- `extension.toml` → `id = "monokai-ignite-theme"` (submodule pinned to commit `f5382b2`)
- Submodule path → `extensions/monokai-ignite-theme`
- `extensions.toml` entry → `[monokai-ignite-theme]`
- `.gitmodules` updated accordingly

## Why a separate extension

I'm aware of the existing Monokai ports — and `monokai-pro-ce` in particular is indeed close, since both themes draw from the Monokai Pro palette. However, Monokai Ignite is not a reproduction of Monokai Pro; it is a redesign *on top of* that palette, with deliberate decisions that the existing ports intentionally do not make:

**1. Dual-accent system (the core differentiator)**

Existing Monokai ports map UI accents directly onto the syntax palette. Monokai Ignite instead separates two non-interchangeable accents:

- **Pink `#FF6188`** — reserved for syntactic emphasis: keywords, tags, errors, write highlights, `version_control.deleted`
- **Amber `#FFA000`** — reserved for UI operational emphasis: `border.focused`, `editor.active_line_number`, `link_text.hover`, `search.active_match_background`, `text.accent`, `element.active`

This keeps "what the code means" visually distinct from "what you're interacting with" — the core idea of the theme, which I didn't find in any existing port.

**2. Read/write highlight separation**

Same-symbol LSP highlights use a neutral read background (`editor.document_highlight.read_background`) versus a warm pink write background (`editor.document_highlight.write_background` = `#FF618816`), so reference flows are scannable at a glance.

**3. Palette semantics carried through the whole UI**

Syntax colors follow Monokai Pro's softer, lower-saturation palette (green `#A9DC76`, cyan `#78DCE8`, purple `#AB9DF2`, yellow `#FFD866`) rather than classic Monokai's neon tones — and the same semantic mapping extends to UI states: added = green, modified = yellow, deleted/error = pink, conflict/warning = orange `#FC9867`. Terminal ANSI colors follow the same palette for consistency, which most ports don't maintain.

**Why not contribute upstream?**

`monokai-pro-ce` is maintained as a faithful port of the original theme — its goal is to reproduce Monokai Pro as-is, so introducing the amber UI accent or the read/write separation would contradict that goal. Monokai Ignite's changes are design decisions layered on the palette (a new theme), not fixes to an existing one; a separate extension keeps both projects' goals intact.

Happy to iterate further if there's anything else to adjust!

darkings:
Updated the extension ID and PR description to reflect the current `monokai-ignite-theme` name and submodule commit. Please re-review when you have a chance. Thank you!

darkings:
@MrSubidubi

PR #13, opened 2024-02-14
Add brainfuck extension

JosephTLyons:

Issue #12, opened 2024-02-14
Markdown preview stopped working


krstp:
With latest ZED preview release `0.123.1`, the _markdown: open preview_ stopped working.

maxdeviant:
Transferring this to the main Zed repo, as it doesn't seem related to extensions.

PR #7105, opened 2026-08-06
Update docker-compose to v0.4.1

eth0net:
Release notes:

https://github.com/eth0net/zed-docker-compose/releases/tag/v0.4.1

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -1169,7 +1169,7 @@ version = "0.1.0"
 
 [docker-compose]
 submodule = "extensions/docker-compose"
-version = "0.1.0"
+version = "0.4.1"
 
 [dockerfile]
 submodule = "extensions/dockerfile"


MrSubidubi (changes requested):
Thanks for this!

Looking at this, I am somewhat wondering whether we should perhaps consider merging this with the Docker-extension short- or long-term - I don't wanna rip this out of your hands, but just throwing my thoughts here and would love to hear your opinion.

Given our [extension publishing prerequisites](https://zed.dev/docs/extensions/developing-extensions#extension-publishing-prerequisites), this extension collides both on the duplicate language definition as well as the language server stuff. Now, I definitely see the value this adds with having the language definition for docker compose, after all, the tasks are super nice and definitely worth to have (we should have a better API for this definitely, but since we don't, that makes sense).

Shipping the language server twice, however, is not something I really like - we have it in the Dockerfile extension already, see https://github.com/zed-extensions/dockerfile/blob/main/src/language_servers/docker_language_server.rs. We generally do not want to ship language servers twice, thus I am not a fan of having this here, although I totally understand why you added that. 

Now, we could update the Docker extension and have the language server be associated with the language. However, that would still require any user to install the Docker compose extension for proper compose support.

Thus, looking at this, I am somewhat thinking whether it would make sense to merge the two, but would be very curious about your thoughts.

eth0net:
I'm happy to have this merged into the Dockerfile extension if that's what you think is best :) it would make things simpler after all.

MrSubidubi:
If that works for you, happy to make it happen! Mind opening a PR against the Dockerfile extension so I can take a look there? Feel free to ping me here and I'll take a look as soon as possible.

eth0net:
@MrSubidubi opened it here: zed-extensions/dockerfile#48

It brings over the `Docker Compose` language and the service runnables, and
fixes the existing `language_ids` mapping — it sends `docker-compose`, but the
server gates Compose support on `dockercompose`, so those features were silently
inactive. Details and verification in the PR.

Happy to close this PR in favour of that one whenever you're ready — the two
would collide on the language definition if both landed. The published
`docker-compose` extension is still at 0.1.0, which predates the language
definition, so nothing conflicts in the meantime.

eth0net:
Closing in favour of zed-extensions/dockerfile#48, which brings the Docker
Compose language and its service runnables into the Dockerfile extension.

Merging this as well would leave two extensions defining a `Docker Compose`
language and shipping the same language server, which is what @MrSubidubi
flagged above. Consolidating there is the better shape — thanks for steering it
that way.

For anyone arriving from the linked issues: the fixes described in
eth0net/zed-docker-compose#6, #8, #9, #11, #12 and #13 are not lost, they move
to the Dockerfile extension via that PR.

PR #7110, opened 2026-08-07
Add Skript language support

DaisyCatTs:

Adds support for [Skript](https://github.com/SkriptLang/Skript), the scripting language used to write Minecraft server plugins. It has a large user base and no editor support in Zed today.

Skript is unusual in a way that shapes the extension: it has no context-free grammar. Every effect, condition, expression and event is a pattern string registered at runtime by Skript or one of its addons — 2,660 in core Skript alone — so nothing lexical distinguishes `set {_x} to 5` from `player is op`. The tree-sitter grammar therefore handles structure only (lines, indentation, sections, strings and their interpolation, variables, comments, and the handful of headers with a fixed shape), and a language server classifies each statement against Skript's published syntax database and returns the result as semantic tokens.

That means `semantic_tokens` needs to be `"combined"` for statements to be coloured; the README covers it, and `highlights.scm` stands on its own without it.

The language server also provides completion, hover, go-to-definition, find references, rename, formatting, inlay hints and diagnostics. It is fetched from GitHub releases, with prebuilt binaries for the five platforms Zed supports.

Tested locally as a dev extension.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4266,6 +4266,10 @@
 	path = extensions/skill-language-server
 	url = https://github.com/CyrusNuevoDia/skill-language-server.git
 
+[submodule "extensions/skript"]
+	path = extensions/skript
+	url = https://github.com/DaisyCatTs/SkriptZed.git
+
 [submodule "extensions/sl4y-theme"]
 	path = extensions/sl4y-theme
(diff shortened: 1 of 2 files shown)


MrSubidubi (approved):
Looks good, thank you!

PR #7128, opened 2026-08-08
Update mc-dp-icons-theme to v0.1.1

amqndin:
Updates `mc-dp-icons-theme` submodule to v0.1.1.


MrSubidubi:
Thanks but unfortunately, this has conflicts, you should base the bump off a fresh main.

Will close with that as it is easier to keep track of the updates that way, please resubmit another bump considering the above. Thanks!

PR #7164, opened 2026-08-10
Update Julia to v0.2.0

piechologist:
Migrate from LanguageServer.jl to JETLS.jl

This is a breaking change!  See the extension's [README](https://github.com/JuliaEditorSupport/zed-julia).

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -2383,7 +2383,7 @@ version = "0.1.0"
 
 [julia]
 submodule = "extensions/julia"
-version = "0.1.10"
+version = "0.2.0"
 
 [just]
 submodule = "extensions/just"


piechologist:
@MrSubidubi There are 2 pending checks awaiting approval (danger & package). I haven't seen this status before and want to be sure it's not related to the CI action we've added to our repo. — Sorry for being pushy: we just migrated to a new LS, and it would be nice to have the extension ready for a talk on Thursday at the annual JuliaCon.

MrSubidubi (changes requested):
No worries, just allowed CI.

Few quick questions:
- The way you download the LSP is not how we usually handle this, is there any reason we cannot just download this using our zed extension API into the extension-available dirs, then inform the local Julia runtime where its at?
- Can we make the granted capabilities less broad? Just wanting everythings seems a bit much
- Language server IDs are normally kebab-case and lowercase - any chance this would also work for you guys here?

piechologist:
Thank you for your quick response! Those are valid points. I think the way to go is to close this PR and open a new one with the changes and a new version number?

aviatesk:
@MrSubidubi Thanks for your reviews!

> The way you download the LSP is not how we usually handle this, is there any reason we cannot just download this using our zed extension API into the extension-available dirs, then inform the local Julia runtime where its at?

JETLS is implemented and distributed as a Julia package rather than as a self-contained executable. It therefore needs to be installed, resolved, and precompiled for the user's local Julia runtime. Using Julia's package manager is currently the most straightforward and reliable way to do that.

This installation model is not new to the JETLS migration. The previous version of the extension also launched Julia with a dedicated `@zed-julia` environment and installed LanguageServer.jl using `Pkg.add`, falling back to `Pkg.update` when loading it failed. The new implementation follows essentially the same package-manager-based model, but pins the JETLS revision and installs it as a Pkg app into an extension-private depot.

Downloading the JETLS source archive through Zed's `download_file` API would not remove the Julia package-manager step: we would still need to invoke Julia to resolve the environment, install dependencies, and precompile JETLS. It would mostly split the existing `Pkg.Apps.add` operation into a Zed-managed source download followed by a manually managed Julia package setup.

Unless JETLS starts publishing a ready-to-run, self-contained distribution artifact (although this would require significant progress on the juliac project currently underway on the Julia runtime side), I believe the current installation approach is reasonable. That said, we can address the broad `process:exec` capability separately and restrict it to the Julia invocations and arguments the extension actually needs.

MrSubidubi:
> I think the way to go is to close this PR and open a new one with the changes and a new version number?

For us it is sufficient if you were to just update the submodule here - we do not check whether the latest commit is a bump. That said, whatever works best for your side.

>  That said, we can address the broad process:exec capability separately and restrict it to the Julia invocations and arguments the extension actually needs.

That would be nice.


Lastly, how much extra-work would it be for you to download the script through the Zed API? I understand that dependencies need to be installed still, and am fine with that, but if we could have the script downloaded that way I'd personally prefer that for the reason that it keeps our ability to disallow downloads on the Zed side.

MrSubidubi:
Closing this as stale, happy to look at a new PR that deals with this or get an update here.

aviatesk:
> Lastly, how much extra-work would it be for you to download the script through the Zed API? I understand that dependencies need to be installed still, and am fine with that, but if we could have the script downloaded that way I'd personally prefer that for the reason that it keeps our ability to disallow downloads on the Zed side.

To keep Zed's ability to disallow downloads meaningful, the extension declares a scoped `download_file` capability for the pinned JETLS release, and before starting a managed installation with Julia's Pkg manager, it now fetches the release's version descriptor through Zed's downloader and verifies it against the hard-coded pin. If that download is rejected, the managed installation aborts — so disabling downloads on the Zed side effectively disables the extension's own network activity as well. The package installation itself still goes through Julia's package manager, which you noted is acceptable for dependencies.

PR #7167, opened 2026-08-10
Add Local Keep AI extension

laynef:
Adds the Local Keep AI extension (`local-keep-ai`, v1.0.0) as a git submodule per CONTRIBUTING.md, with a sorted `extensions.toml` entry.

Local Keep AI provides slash commands (/lk, /lk-refactor, /lk-tests, /lk-run, /lk-commit, /lk-models) that drive the `lk` CLI (`pip install local-keep-ai-cli`) for local-first AI assistance. The extension repository is MIT licensed (LICENSE at the repo root), the submodule commit is on its `main` branch, and the extension ships only extension code — the CLI is resolved from the user's PATH at runtime, never bundled.

Supersedes #6170, which predated the license file and the submodule format; both are addressed here. CLA: being signed by @laynef.

PR #7170, opened 2026-08-10
Add Local Keep AI extension

laynef:
Adds the Local Keep AI extension (`mcp-server-local-keep-ai`, v1.2.2) as a git submodule, with sorted `extensions.toml` and `.gitmodules` entries (`pnpm sort-extensions`).

The extension provides a **context server** for the Agent Panel: it launches `lk mcp` from the Local Keep AI CLI, exposing `ask` and `list_models` tools over MCP. Per the publishing prerequisites for MCP server extensions, the server is **not** bundled — the user's per-server `command` settings (path, arguments, env) are honored when set, otherwise `lk` resolves from the user's environment, and a descriptive error plus `ContextServerConfiguration` installation instructions cover the not-installed case. The extension defines no settings of its own and reads no environment outside what Zed provides.

**Testing on the exact submitted commit**, replicating this repo's CI workflow locally (linux/amd64, Node 24, pnpm 9, Rust 1.90 / `wasm32-wasip2`, pinned `zed-extension` CLI `2a00db06`), with the diff base set to current `main`:
- `pnpm install`, `pnpm build`, `pnpm test` — all pass.
- `pnpm package-extensions` — detects one changed extension, packages it; manifest reports `"provides":["context-servers"]`, `wasm_api_version: 0.7.0`.
- Runtime: full MCP session against the published CLI (v1.21.46 on PyPI) — initialize/protocol negotiation, `tools/list`, and real `tools/call` round-trips for both tools.

The extension repository is MIT licensed (LICENSE at the root), the submodule uses an HTTPS URL, and the pinned commit is the head of its `main` branch. The id follows the `mcp-server-*` convention used by MCP server extensions in this registry. CLA is signed.

Supersedes #7167/#7169 (originally submitted against the deprecated slash command API).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2890,6 +2890,10 @@
 	path = extensions/mcp-server-kubernetes
 	url = https://github.com/ferra-io/zed-mcp-server-kubernetes.git
 
+[submodule "extensions/mcp-server-local-keep-ai"]
+	path = extensions/mcp-server-local-keep-ai
+	url = https://github.com/laynef/localkeep-zed.git
+
 [submodule "extensions/mcp-server-markitdown"]
 	path = extensions/mcp-server-markitdown
 	url = https://github.com/G36maid/zed-mcp-server-markitdown
--- a/extensions.toml
+++ b/extensions.toml
@@ -2945,6 +2945,10 @@ version = "0.1.2"
 submodule = "extensions/mcp-server-kubernetes"
 version = "0.1.0"
 
+[mcp-server-local-keep-ai]
+submodule = "extensions/mcp-server-local-keep-ai"
+version = "1.2.2"
+
 [mcp-server-markitdown]
 submodule = "extensions/mcp-server-markitdown"
 version = "0.0.1"


laynef:
Fixing the merge conflict now. Please give me one second

MrSubidubi (changes requested):
Thanks for this, this is looking better
- https://github.com/laynef/localkeep-zed/blob/b52fe4b83515637f4e3e1d2c7ef12bf6dc36e23c/src/lib.rs#L36 you should search in the environment with `Worktree::which` for the binary and throw an appropriate error if no binary is found.
- You should be able to read a custom user path from the `ContextServerSettings` as these should provide the user with a configurable path already, so that you do not need the config for your server at all like this. Since a user can already configure `path` there, that should be sufficient

laynef:
> Thanks for this, this is looking better
> 
> * https://github.com/laynef/localkeep-zed/blob/b52fe4b83515637f4e3e1d2c7ef12bf6dc36e23c/src/lib.rs#L36 you should search in the environment with `Worktree::which` for the binary and throw an appropriate error if no binary is found.
> * You should be able to read a custom user path from the `ContextServerSettings` as these should provide the user with a configurable path already, so that you do not need the config for your server at all like this. Since a user can already configure `path` there, that should be sufficient

Okay thank you! I will have those changes done very soon!

laynef:
Thank you for the review — both points are addressed in v1.2.1 → v1.2.2 (submodule bumped, branch rebased on current `main`).

**Custom path now comes from `ContextServerSettings`.** The extension no longer defines any settings of its own; `context_server_command` reads the user's per-server `command` block and honors `path`, `arguments` and `env` as configured, falling back to `lk mcp` resolved from the user's environment when there is no override:

```rust
let settings = ContextServerSettings::for_project(context_server_id.as_ref(), project)?;
if let Some(command) = settings.command {
    let path = command.path.ok_or(INSTALL_HINT)?;
    return Ok(Command {
        command: path,
        args: command.arguments.unwrap_or_else(|| vec!["mcp".to_string()]),
        env: command.env.map(Vec::from_iter).unwrap_or_default(),
    });
}
```

**Missing-binary case returns a descriptive error** pointing at both remedies — `pip install local-keep-ai-cli && lk login`, or setting `command.path` for this server — and the same guidance is surfaced through `ContextServerConfiguration`'s installation instructions.

laynef:
I apolgize. I changed the title back to what you made it. Claude made the title change on accident. All the changes you asked for were made. I updated the branch so all is up to date. I hope all is done correctly as you asked. I hope all is good now. I went ahead and made all the changes you asked for in v1.2.2

PR #7172, opened 2026-08-10
Add gruvbox-poly-theme

polina4096:


MrSubidubi (changes requested):
Thanks for this!

polina4096:
> Thanks for this!
> 
> It seems that for your theme, we already have an existing ports. Could you please elaborate on why we should publish your theme as a separate extension to the already existing themes and explain where your theme differs from the existing themes/why your changes do not make sense to contribute upstream? Thanks! 

It's not a port, but my personal theme inspired by gruvbox. It's not just a copy or a color shift. I've put substantial amount of time working on it.

PR #7173, opened 2026-08-10
Update TSRX to v0.0.85

zhsama:
## Summary

- update the TSRX extension registry entry from 0.0.84 to 0.0.85
- advance the Ripple submodule to the commit containing the matching Zed extension manifest
- update the bundled language server from 0.2.208 to 0.3.118

## Why

The registry still points at TSRX 0.0.84. Ripple main already contains 0.0.85, which updates the language tooling to the current TSRX packages and includes Octane compiler auto-detection for published packages.

## Validation

- pnpm sort-extensions
- pnpm build
- pnpm test (115 tests passed)
- cargo check --manifest-path extensions/tsrx/packages/zed-plugin/Cargo.toml --target wasm32-wasip2

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -5122,7 +5122,7 @@ version = "1.1.0"
 [tsrx]
 submodule = "extensions/tsrx"
 path = "packages/zed-plugin"
-version = "0.0.84"
+version = "0.0.85"
 
 [turtle]
 submodule = "extensions/turtle"


zhsama:
@cla-bot check

MrSubidubi (changes requested):
Thanks for this!

We usually do not check for the server with a script like this - you can either just let Zed error in that case (we show a generic error there) or check with `fs::metadata`, then throw an appropriate error.

Would kindly ask you to make that update, all else looks good. Thanks!

zhsama:
Thanks! I removed the shell/cmd wrapper and switched the extension back to launching the resolved language-server executable directly in Ripple-TS/ripple#1416. Once that upstream change is merged, I’ll update this PR’s submodule pointer to the merged commit.

leonidaz:
we have a process now set up to publish the extension via a pr whenever one released. 

here's the one for version 0.0.86 https://github.com/zed-industries/extensions/pull/7200 which has all fixes in this pr + the merged pr that calls the language server directly: https://github.com/Ripple-TS/ripple/pull/1416

MrSubidubi:
Thanks! Closing in favor of the newer bump

PR #7182, opened 2026-08-11
Bump Soufflé to v0.1.1

panosdimak:
Bumps the Soufflé extension to v0.1.1.

Fixes the language server failing to start with `operation not supported on this platform`; the jar path was resolved with `std::fs::canonicalize`, which is unsupported on WASI. It now joins onto `std::env::current_dir()`.


MrSubidubi (approved):
Thank you, forgot about this during the too many reviews here 😅

PR #7211, opened 2026-08-13
Update deps-language-server to v0.1.6

bug-ops:
## Summary
- Linux now always requests the statically-linked unknown-linux-musl deps-lsp asset instead of unknown-linux-gnu, since zed_extension_api has no way to detect the host libc at runtime and a static musl binary runs on both glibc and musl distros.
- Downloaded archives are now verified against the .sha256 sidecar published on the same GitHub release before extraction. The sidecar parser handles both plain sha256sum output and Windows CertUtil output.

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -1183,7 +1183,7 @@ version = "1.10.0"
 
 [deps-language-server]
 submodule = "extensions/deps-language-server"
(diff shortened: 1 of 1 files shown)


MrSubidubi (approved):
Thanks!

PR #7228, opened 2026-08-15
Update devicetree to v0.2.0

anikinmd:

This PR bumps the devicetree extension to the v0.2.0
New version brings new LSP
https://github.com/kylebonnici/dts-lsp
Originally proposed in the PR #6055 

Initial idea was to keep two LSP servers, because new LSP requires configuration. During the development I found that it's easier to have one preconfigured. Extension ships default configuration to the LSP, that is simple enough to have a good workflow on the Linux Kernel tree. Configuration can be overridden with project settings.
Also the extension now ships snippets for the project settings to have a simple configuration for Kernel or Zephyr based projects.

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -1222,7 +1222,7 @@ version = "2.0.1"
 
 [devicetree]
 submodule = "extensions/devicetree"
-version = "0.0.1"
+version = "0.2.0"
 
 [dhall]
 submodule = "extensions/dhall"


MrSubidubi (approved):
The snippet support for the settings is something I never thought of and is unorthdox, but am happy to accept this.

Will merge albeit this is not how most language servers do their stuff, but looks good to me and looks like a good approach. 

Thank you so much for jumping in here!

PR #7246, opened 2026-08-17
Update sora-theme to v0.1.2

Aejkatappaja:
Updating an extension I maintain.

Release notes: https://github.com/Aejkatappaja/sora/releases/tag/v1.8.0

The theme picks up a raised border colour, and a fix where the editor's invisible characters were rendered from the border token instead of the one that means invisibles, so they no longer disagree with the same theme's Neovim side.

Tested locally as a dev extension from `extras/zed`.


MrSubidubi (approved):
Thanks!

PR #7287, opened 2026-08-21
Add pycharm-dracula-themes

YanivZalach:

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4062,6 +4062,10 @@
 	path = extensions/pyatlas-lsp
 	url = https://github.com/Akamine2001/zed-pyatlas.git
 
+[submodule "extensions/pycharm-dracula-themes"]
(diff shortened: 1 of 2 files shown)


YanivZalach:
@cla-bot check

MrSubidubi (changes requested):
Thanks for this!

YanivZalach:
Hello!

Mainly because the original pycharm theme is grate, and is very close to the pycharm theme.
However when doing web dev the syntax highlighting is not clear enough in my eyes.

My theme address that exact point, and as such not the exact theme for python as the original pycharm theme that the original is more slimier to. Meaning while slimier, they are not the same and offer a different experience.

Appreciate the response.

YanivZalach:
If a better name is needed to draw the difference between theme, tell me :)

PR #7318, opened 2026-08-23
Add Pale Fire theme

sahmed007:
Adds Pale Fire, a four-variant theme with broad syntax coverage and additional Ruby, Rails, and ERB testing.


sahmed007:
@cla-bot check

MrSubidubi (approved):
Thanks!

PR #7329, opened 2026-08-25
Add One Monokai theme

Sothcheat:

Adds **One Monokai** (`one-monokai-theme`), a dark theme pairing One Dark's UI palette with Monokai's syntax assignments.

Repository: https://github.com/Sothcheat/one-monokai-theme

### Provenance and license

This is an unofficial port of [azemoh/vscode-one-monokai](https://github.com/azemoh/vscode-one-monokai) by Joshua Azemoh, which is MIT licensed. The extension's `LICENSE` retains the original copyright and permission notice alongside the port's, and the README credits the original author and states that the port is unaffiliated and unendorsed.

### Notes

- Each color is derived from the TextMate scope the original theme assigns, mapped onto the tree-sitter capture Zed emits for that token, rather than eyeballed from screenshots.
- Syntax coverage spans the grammars bundled with Zed and the nvim-treesitter capture vocabulary that language extensions use, so Java, C#, Dart, Kotlin, Rust and friends are colored rather than falling through to the default foreground.
- The theme validates against `https://zed.dev/schema/themes/v0.2.0.json`.
- Tested in Zed at the submitted submodule commit (`9e7959b`), which is on `main`.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -3622,6 +3622,10 @@
 	path = extensions/one-hunter
 	url = https://github.com/teziovsky/zed-one-hunter-theme.git
 
+[submodule "extensions/one-monokai-theme"]
+	path = extensions/one-monokai-theme
+	url = https://github.com/Sothcheat/one-monokai-theme.git
+
 [submodule "extensions/onurb"]
 	path = extensions/onurb
 	url = https://github.com/brunoocrv/onurb-zed
--- a/extensions.toml
+++ b/extensions.toml
@@ -3685,6 +3685,10 @@ version = "0.1.0"
 submodule = "extensions/one-hunter"
 version = "0.0.4"
 
+[one-monokai-theme]
+submodule = "extensions/one-monokai-theme"
+version = "0.1.0"
+
 [onurb]
 submodule = "extensions/onurb"
 version = "0.1.0"


Sothcheat:
I've verified my commit email on my GitHub account, so the commits now resolve to @Sothcheat. Could the CLA check be re-run?

Sothcheat:
@cla-bot check

Sothcheat:
@cla-bot check

MrSubidubi (changes requested):
Thanks for this!

Sothcheat:
Hi @MrSubidubi!

The reason I made this theme is that there isn't an existing one that matches the exact "One Monokai" color scheme.
image

These are all the Monokai themes currently in the Extensions. All of them are their own variations of the original Monokai theme. Mine is also a variation, specifically a port of the VS Code theme originally created by azemoh: https://github.com/azemoh/vscode-one-monokai.git.

I used that theme in VS Code and loved it. When I switched to Zed, I couldn't find the exact same theme with the same color scheme for the editor, syntax, and so on. Because of that, I decided to port the theme to Zed myself. Here is a comparison between zed-one-monokai and the others I found in Zed's Extensions:

image
image

And here is how the One Monokai theme looks in VS Code:

image

I hope you can see the vision here. Thank you!

PR #7348, opened 2026-08-26
Add Yona extension

akovari:
Adds the official Yona extension, backed by `yls` and the public `tree-sitter-yona` grammar.

Validation:
- `npx tree-sitter-cli test`
- stdlib corpus parses without Tree-sitter ERROR nodes
- `cargo check` and `cargo check --target wasm32-wasip2`
- manifest/discovery smoke check


akovari:
@cla-bot check

PR #7351, opened 2026-08-26
Add Java Properties extension

peppe1337:
Adds syntax highlighting for Java `.properties` files (`java.util.Properties`).

**Why:** `.properties` files currently open as plain text. No built-in grammar under `crates/grammars/src` claims the `properties` path suffix, the `java` extension claims only `["java"]`, and the extension search API returns no results for `properties`. I did not check the `path_suffixes` of all 1442 registered extensions individually.

**Grammar:** [tree-sitter-grammars/tree-sitter-properties](https://github.com/tree-sitter-grammars/tree-sitter-properties) (MIT), pinned to `6310671b24d4e04b803577b1c675d765cbd5773b`.

**Contents:** grammar and `highlights.scm` only. No Rust code, no language server, no snippets.

**Testing:** I installed the extension at commit `8e3ed4a` into Zed 1.16.3 on Linux and opened a `.properties` file containing comments (`#` and `!`), both `=` and `:` separators, a Unicode escape and a line continuation. The status bar reads `Java Properties` and highlighting is applied. With the extension removed, the same file in the same editor reads `Unknown` and shows no highlighting.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2326,6 +2326,10 @@
 	path = extensions/java-eclipse-jdtls
 	url = https://github.com/ABckh/zed-java-language-support-jdtls
 
+[submodule "extensions/java-properties"]
+	path = extensions/java-properties
+	url = https://github.com/peppe1337/zed-java-properties.git
+
 [submodule "extensions/javascript-snippets"]
 	path = extensions/javascript-snippets
 	url = https://github.com/seekode/zed-js-snippets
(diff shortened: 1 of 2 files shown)


peppe1337:
Closing this myself.

Zed's `CONTRIBUTING.md` states: *"we don't accept contributions from autonomous agents."* This PR was produced by one. I did not read that section before opening it — I had read the same file for a different question and missed it. That is my error, and the right correction is to withdraw rather than leave it in your queue.

The `cla-signed` check is also unsigned, so it could not have been merged in any case.

Sorry for the triage cost.

PR #7358, opened 2026-08-27
Update helm to v0.0.7

cabrinha:
Release notes:

https://github.com/cabrinha/helm.zed/releases/tag/v0.0.7

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -2097,7 +2097,7 @@ version = "0.0.8"
 
 [helm]
 submodule = "extensions/helm"
-version = "0.0.6"
+version = "0.0.7"
 
 [herzha-theme]
 submodule = "extensions/herzha-theme"


kirillmandrygin:
@MrSubidubi hi! Is there a chance that it will be merged in the near future? sorry for the mention I'm just looking forward to this release...

MrSubidubi (changes requested):
Thanks for this!

- The added path suffixes seem to overlap with existing extensions, e.g. `gotmpl`. Could you elaborate on that?
- Can we please just query for _one_ settings key when it comes to the language server config, more specifically, the same ID for the language server that is specified in the extension.toml? Should that not also help with wrapping the object, since that should be more predictable/can we perhaps then not guess there but actually know what to do?
- What is the `yamlls`-thing about, and more specifically, why do we also query for the same server suffixed with `.js`?

cabrinha:
Thanks for the review! All three points are addressed in v0.0.8, which is up in #7468 (this PR is superseded by it — same submodule history plus one fix commit):

**1. Overlapping path suffixes (`gotmpl`)**

You're right. Bare `gotmpl` is claimed by the gotmpl extension (`hjr265/zed-gotmpl`), and I verified in Zed's matcher (`available_languages.rs`, post-#29716 longest-match) that a `foo.gotmpl` file scores identically for both extensions, so the winner is effectively arbitrary. Removed `gotmpl` and the unjustified `helm` suffix. Kept `tpl`, the compound `yaml.gotmpl`/`yml.gotmpl`, `helmfile.yaml`/`helmfile.yml`, and `.helmignore` — all Helm-specific with no competing claimant. Chart `*.yaml` templates stay on user `file_types` globs per https://zed.dev/docs/languages/helm, and `_helpers.tpl`/helmfile files that open with an action are still caught by `first_line_pattern`.

**2. Single settings key**

Done — `lsp_settings()` now queries only `language_server_id.as_ref()`, i.e. `lsp.helm`, the ID from `[language_servers.helm]` in extension.toml (same convention as Zed's own extensions). With exactly one known shape, nesting the flat user object under helm-ls's `helm-ls` section is exact, so the unwrap/guess helper (`unwrap_helm_ls_section`, `LSP_SETTING_KEYS`) is deleted. README now documents a single flat shape under `lsp.helm.settings`.

**3. The `yamlls` thing + `.js` suffix**

helm-ls shells out to yaml-language-server for Kubernetes schema validation; `yamlls.path` and the `YAMLLS_PATH` env var are its documented knobs (`internal/util/config.go`, default executable `yaml-language-server`). The extension resolves the absolute path so the server doesn't depend on PATH lookup inside Zed's spawned environment — an explicit user `yamlls.path` still wins via `or_insert`. The `.js` variant is dropped (it's not helm-ls's default executable name and I found no evidence for it), and the two duplicated lookups are now one `yamlls_path()` helper used for both the config injection and the env var.

Happy to close this one in favor of #7468.

MrSubidubi:
Thanks for the follow-up!

> Happy to close this one in favor of https://github.com/zed-industries/extensions/pull/7468.

Will do that for you, thanks!

PR #7364, opened 2026-08-27
Update windows-batch to v0.3.0

pleahmacaka:
Command arguments like `--id=Figma.Figma` were parsed as a single node, so flags, separators and values all got one color. The grammar now splits them, and the queries highlight each part.

| Before | After |
| --- | --- |
| before | after |

Ref pleahmacaka/zed-batch#1


MrSubidubi (approved):
Thanks!

PR #7370, opened 2026-08-27
Update laravel to v0.7.3

mikebronner:
Bumps `laravel` to v0.7.3.

Release notes: https://github.com/mike-bronner/zed-laravel/releases/tag/0.7.3

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -2599,7 +2599,7 @@ version = "1.0.3"
 
 [laravel]
 submodule = "extensions/laravel"
-version = "0.7.1"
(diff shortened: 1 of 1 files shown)


MrSubidubi:
Please remove the shellcheck stuff here completely, I do not see any good reason for this to live in this extension as this is not the intended use of that API and violates our 
[extension publishing prerequisites](https://zed.dev/docs/extensions/publishing/prerequisites).

Also, why does your language server need env-support? Curious at the very least, but the first I consider blocking. Will close due to that for now.

PR #7412, opened 2026-08-29
Update laravel to v0.7.5

mikebronner:
Bumps `laravel-ce` from 0.7.2 to v0.7.5.

Release notes: https://github.com/mike-bronner/zed-laravel/releases/tag/0.7.5
Release notes: https://github.com/mike-bronner/zed-laravel/releases/tag/0.7.4
Release notes: https://github.com/mike-bronner/zed-laravel/releases/tag/0.7.3

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -2613,7 +2613,7 @@ version = "1.0.3"
 
 [laravel]
 submodule = "extensions/laravel"
-version = "0.7.1"
+version = "0.7.5"
 
 [laravel-official]
 submodule = "extensions/laravel-official"


mikebronner:
@MrSubidubi Thanks for your feedback on #7370.

## 1. The shellcheck injection — removed completely

I completely agree, that was overstepping and I put my personal config as a default. Thanks for catching that. I was only thinking of user experience from the point of Laravel, and didn't take the time to consider how it might affect other projects and users. Thank you for the reference to the guidelines, I will keep those in mind going forward.

`LaravelExtension::language_server_additional_workspace_configuration` is gone,
along with its three private helpers (`shellcheck_suppression_enabled`,
`is_laravel_worktree`, `shellcheck_args_with_sc2034_excluded`) and their eleven
unit tests. The `shellcheck.suppressUnusedVarWarnings` setting went with them,
as did every claim in the README and docs that this extension quiets SC2034 for
you.

## 2. Why the language server needs env support

Two halves: one attach entry we could not defend and dropped, one we kept and
can.

### Dropped: `"env"`

It targeted the language name registered by third-party dotenv extensions.
Those register a bare `"env"` path suffix, which loses the length tie to bash's
`".env"` — so the entry never claimed the files it appeared to. It bought no
capability while reading, to a reviewer already looking at over-reach, as this
extension claiming another extension's language.

The attach list is now `["PHP", "Blade", "XML", "Shell Script"]`.

### Kept: `"Shell Script"`

Three facts, in causal order.

**This extension defines no languages and ships no grammars.** No `languages/`
directory, no `[languages]` section, no `[grammars]` section. The `languages`
key under `[language_servers.laravel-lsp]` is an *attach* list — "run my server
on files Zed has already classified this way." It registers nothing, overrides
nothing, and claims nothing.

**Env features dispatch on the filename, never on Zed's classification.** Every
one routes through a single predicate, `env_key_locator::is_env_file_name` (or
`path_is_env_file` for callers holding a path), matching `.env` exactly or a
`.env.` variant prefix. A buffer Zed classifies as Shell Script but does not
*name* `.env` or `.env.<suffix>` fails that gate and reaches no env feature:
no completion, no code lens, no semantic tokens, no hover, no go-to-definition.
Find-references and rename are narrower still — `.php` and `.blade.php` only.

**`Shell Script` is the only route to `.env*` files.** Zed ships no `env`
language and classifies every env file as Shell Script: bare `.env` through the
bash grammar's `path_suffixes`, and the `.env.*` variants through the
`"Shell Script": [".env.*"]` entry in Zed's own default settings. One attach
entry reaches `.env` and every variant alike, with no user configuration.

Shipping a dotenv language of our own was considered and rejected. It could win
the tie for bare `.env`, but the `.env.*` variants sit at a precedence tier only
a user's own `file_types` can override — so it would fix `.env` and leave
`.env.local` shell-linted. A split worse than the uniform status quo.

`"XML"` follows the identical filename-gated pattern, for `phpunit.xml` env-var
autocomplete. Naming it here so it is not the next surprise.

The reasoning now lives in the repo rather than in a thread: the comment above
`languages` in `extension.toml`, and in user-facing terms under "Why this
extension attaches to `Shell Script`" in `docs/environment.md`.

## What checking your question turned up

Writing that claim down meant verifying it, and it was not true of every call
site.

Env-key completion tested `path.contains(".env")` — a substring match against
the whole path — so a shell script under a directory named `.envs/` was treated
as an env file and offered env completion. Precisely the over-reach you were
probing, and it was real.

Two further sites had drifted the same way, including the step that admits an
edited buffer as an env source at all, which used a `.env` *prefix* match and so
swallowed `.envrc` — direnv's file, common in Laravel repositories. Because env
sources merge by variable name, a `DATABASE_URL` in a project's `.envrc` could
answer for the one its own `.env` declares.

All of them now route through the one predicate, and the tests drive the real
LSP handlers rather than calling the predicate directly: a unit test of an
extracted predicate proves nothing about whether the dispatch consults it.

The env surface was tightened further in this release — completion no longer
merges the language server process's own environment into the popup, and
sensitive values are masked in LSP output.

MrSubidubi (approved):
Much better, thank you.

MrSubidubi (approved):

PR #7433, opened 2026-08-31
Add cartan language support

jzrake:
Adds the cartan extension at 0.4.2: language support for cartan documents (`.cart`) — the tree-sitter grammar from cartan-lang/tree-sitter-cartan and the `cartan lsp` language server, which the extension starts from the user's PATH. The extension repository (https://github.com/cartan-lang/zed-cartan) is refreshed per release of the language, in lockstep with its version.


MrSubidubi (approved):
Thanks!

PR #7452, opened 2026-09-02
Add visual-studio-2019-dark-theme

yueyinqiu:

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -5442,6 +5442,10 @@
 	path = extensions/visual-assist-dark
 	url = https://github.com/trojanfoe/visual-assist-dark.zed.git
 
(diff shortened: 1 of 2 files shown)


yueyinqiu:
@cla-bot check

MrSubidubi (changes requested):
Thanks for this!

yueyinqiu:
Thank you for the reply!

If you are referring to `visual-studio-dark-theme`, then they are actually two different themes. `visual-studio-dark-theme` is a built-in theme in VS Code. The `visual-studio-2019-dark-theme` I submitted is a migration from https://github.com/dotnet/vscode-csharp/blob/main/themes/vs2019_dark.json. It provides a much closer experience to Visual Studio.

PR #7483, opened 2026-09-05
Update lini to v0.2.0

monfa-red:
Bumps the Lini extension to 0.2.0.

- The tree-sitter grammar gains a rule for `--name:` variable declarations; the regenerated parser (ABI 14) parses every sample in the Lini repo with zero ERROR nodes.
- `highlights.scm` regenerated from the 1.0 property ledger.
- Grammar pin in `extension.toml` moved to the commit carrying the regenerated parser.

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -2733,7 +2733,7 @@ version = "0.1.1"
 [lini]
 submodule = "extensions/lini"
 path = "editors/zed"
-version = "0.1.0"
+version = "0.2.0"
 
 [linkerscript]
 submodule = "extensions/linkerscript"


monfa-red:
Friendly ping — this one's been open since Sep 5 and all checks are green (danger, package, CLA).

It's the standard two-line bump: `version` 0.1.0 → 0.2.0 in `extensions.toml`, and the submodule pointer moved to the commit carrying that `extension.toml`. The grammar pin in `[grammars.lini]` already points at the commit whose generated parser handles every sample with zero ERROR nodes.

Users currently get a grammar from well before the language's 1.0, so it misses a fair bit of current syntax. Happy to rebase or split it differently if that'd help — thanks for maintaining this.

MrSubidubi (approved):
Could not review this the last time around since I had too bad internet for GitHub to load the diff.


Does look good now though, thanks!

PR #7493, opened 2026-09-05
Update stylelint to v3.2.0

florian-sanders:

Changes:

* **lsp:** Allow opting out of the managed language server download via `lsp.stylelint-lsp.settings.download`
  * Setting `download` to `false` makes the extension only use `binary.path` or the worktree `PATH`, and fail with an explicit error when neither provides the binary, instead of silently falling back to the extension-managed download.
  * Omitting the setting or using `true` preserves the managed fallback.

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -4941,7 +4941,7 @@ version = "0.0.1"
 
 [stylelint]
 submodule = "extensions/stylelint"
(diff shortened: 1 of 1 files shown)


MrSubidubi (approved):
Thanks!

PR #7494, opened 2026-09-05
Add aster extension

Zfinix:
Adds the **Aster** extension (id: `aster`, version 0.1.0, Apache-2.0).

- Repository: https://github.com/Zfinix/aster, submodule at `extensions/aster`
- Adds a `/aster-review` slash command to the Zed assistant panel that runs an Aster review of the current changes and inserts the findings
- Builds clean against the Zed extension API (wasm target, verified locally)


Zfinix:
@cla-bot check

PR #7530, opened 2026-09-09
Add Inkit extension

waterblower:
Adds Inkit (`inkit`) 0.3.0 for the Ink narrative scripting language.

Includes an original Tree-sitter grammar, highlighting, outline, definition/reference navigation, completion, and local image hover previews. The Rust adapter downloads platform-specific `ink-lsp` binaries from https://github.com/waterblower/inkit/releases; no server executable is bundled in the extension. The repository includes an MIT license and a public, pinned grammar source.

There is already an `ink` extension in the registry. But this project does not seem active and is missing some features that I need.

- [x] I've read [the contribution guidelines](https://github.com/zed-industries/extensions/blob/main/CONTRIBUTING.md). The overlap and remaining release-installation validation are disclosed above.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2298,6 +2298,10 @@
 	path = extensions/ink
 	url = https://github.com/yuna0x0/zed-ink.git
 
+[submodule "extensions/inkit"]
+	path = extensions/inkit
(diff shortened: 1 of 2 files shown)


waterblower:
@cla-bot check

PR #7543, opened 2026-09-10
Add Notion theme

amantibrewal310:
Adds **Notion**, a light and dark theme built from Notion's own colour palette (its text and icon colours and interface surfaces), mapped onto Zed's syntax and UI slots.

Every colour is lifted until it clears WCAG AA against the editor background (5.4:1 for syntax, 4.6:1 for comments and hints), so it stays readable at small sizes. Both themes define every style key Zed currently reads and all 102 syntax keys.

Repository: https://github.com/amantibrewal310/zed-notion-theme
Licence: MIT

Tested manually in Zed as a dev extension at the submodule commit.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -3534,6 +3534,10 @@
 	path = extensions/noted-theme
 	url = https://github.com/sergeevalera/noted-theme.git
 
+[submodule "extensions/notion-theme"]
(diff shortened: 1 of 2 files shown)


amantibrewal310:
@cla-bot check

PR #7554, opened 2026-09-11
Update `eh-web-themes` to v1.0.0

huhuhuhuheh:
It seems like 1.0 was disliked by the marketplace, so it is now 1.0.0

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -1417,7 +1417,7 @@ version = "0.0.6"
 
 [eh-web-themes]
 submodule = "extensions/eh-web-themes"
-version = "1.0"
+version = "1.0.0"
 
 [eiffel-theme]
 submodule = "extensions/eiffel-theme"


MrSubidubi (approved):
Thank you so much for spotting this! This just helped a few other users as well. 

I updated the logic just now so that this no longer can and will happen for future users. Still, apologies for the inconvenience this caused you!

Also, 

Image

huhuhuhuheh:
I agree with the image lol

I only figured out when it wasn't appearing in the extension browser

MrSubidubi:
Yeah, thanks so much for raising this and looking into it!

One other extension we could already help due to your findings, and two others I opened a PR against. Found even a fifth one that was stale at this point and had to remove it. Work like yours helps us immensively with things like this, thank you so much!

huhuhuhuheh:
> Yeah, thanks so much for raising this and looking into it!
> 
> One other extension we could already help due to your findings, and two others I opened a PR against. Found even a fifth one that was stale at this point and had to remove it. Work like yours helps us immensively with things like this, thank you so much!

I guess I figured out why i suppose, if u also don't mind telling what PR numbers are those cuz of my findings?

MrSubidubi:
https://github.com/zed-industries/extensions/pull/7619/changes#diff-dfcd8afcb372a75a1eb1bdc0af57d36a5e0a727dbca99350779abb21333a298fR23 contains the list of affected extensions, with yours included, we already fixed three of those.

PR #7566, opened 2026-09-11
Update deps-language-server to v1.0.0

bug-ops:
## Summary
- Syncs with a deps-lsp release verification fix: the extraction step now verifies the archive that is actually extracted, instead of a second unverified download
- Documents deps-lsp 0.12-0.14 additions: GitHub Actions and GitLab CI/CD ecosystem coverage, license hover with SPDX allow/deny-list diagnostics, and pnpm-lock.yaml lock file support
- Bumps the version in `extensions.toml` from 0.1.8 to 1.0.0

### DIFF
--- a/extensions.toml
+++ b/extensions.toml
@@ -1242,7 +1242,7 @@ version = "1.11.0"
 
 [deps-language-server]
 submodule = "extensions/deps-language-server"
-version = "0.1.8"
+version = "1.0.0"
 
 [depsy-lsp]
 submodule = "extensions/depsy-lsp"


MrSubidubi (changes requested):
Why is this reimplementing something that we ship bundled as part of Zed?

bug-ops:
> Why is this reimplementing something that we ship bundled as part of Zed?

`zed::download_file` bundles "download" and "extract" as one opaque step with no way to extract bytes that were already fetched and verified, so every call is a fresh HTTP request.

We need to checksum-verify the archive against the .sha256 sidecar published with each `deps-lsp` release before anything gets executed. Doing that with the bundled helper alone means calling `download_file` twice: 
- once Uncompressed to hash;
- once with GzipTar/Zip to extract
And those are two independent network fetches. 

That's exactly the bug I hit and fixed (bug-ops/deps-zed#14): the verified copy gets discarded, and the binary that actually gets extracted and made executable comes from a second, unverified fetch. A compromised CDN edge or MITM proxy between the two requests can serve honest bytes to the check and a backdoored binary to the extraction - TOCTOU on the integrity gate.

The fix is to download the archive once, hash and verify those exact in-memory bytes, then extract from that same buffer, which requires local tar.gz/zip extraction, since the bundled API doesn't expose an "extract already-downloaded bytes" step. It also lets us reject zip-slip entries ourselves regardless of the host Zed version, and avoid tar::Archive::unpack()'s set_permissions call, which isn't implemented under Zed's wasm32-wasip2 extension host.

PR #7582, opened 2026-09-13
Update depsy to v2.1.0

mpiton:
## Summary

- Update the `depsy-lsp` extension to v2.1.0
- Point the `extensions/depsy-lsp` submodule at the v2.1.0 release commit

Minor release. The manifest only changes its version. The extension now downloads the language-server release that matches its own version from the direct release-asset URL, so it no longer calls `api.github.com`, and once the binary is on disk it makes no network request at start-up. Downloads are checksum-verified in a staging directory before they replace anything, and when GitHub is unreachable the extension falls back to the newest release already downloaded. Full notes: https://github.com/mpiton/zed-depsy/blob/v2.1.0/CHANGELOG.md

## Type

chore

## Validation

- `pnpm build`
- `pnpm test` (115 passed)
- `pnpm sort-extensions` (no changes)
- `cargo build --release --target wasm32-wasip2 --locked` in `depsy-zed` at the tagged commit
- Release assets published for all five targets at https://github.com/mpiton/zed-depsy/releases/tag/v2.1.0


MrSubidubi (changes requested):
Why do you not use the `github_release_by_tag_name`, if I may ask?

mpiton:
> Why do you not use the `github_release_by_tag_name`, if I may ask?

Sure. It's on purpose, and it's the reason 2.1.0 exists.

`github_release_by_tag_name` goes through `api.github.com` (`/repos/{repo}/releases/tags/{tag}`). Unless the user has `GITHUB_TOKEN` in Zed's environment, that's the unauthenticated limit of 60 requests an hour per IP, and it needs one more host to be reachable. That's what broke in mpiton/zed-depsy#400: the extension called `latest_github_release` on every start, the call failed on a corporate network, and the language server didn't start even though the binary was already on disk. The `releases/download/` URLs don't count against the API limit, and zed-industries/zed#9483 is about that same 403.

The other reason is that the lookup has a known answer. The tag is `v{CARGO_PKG_VERSION}` and the asset names are fixed by our release workflow, so the `download_url` the API would return is the string the extension already builds (the API hands back `browser_download_url`, which is that URL). An extra request and an extra failure mode, for nothing new. If an asset is missing the download fails like any other error and the extension falls back to the newest release it has on disk.

Downloads and the checksum fetch still go through Zed's HTTP client (`download_file` and `http_client::fetch`), so proxy settings apply as before.

If you'd rather have the API call for consistency with other extensions, I can do it, but it means a 2.1.1 release and a new PR here, and I'd want to keep the direct URL. Is there something on the Zed side I'm missing, like plans to cache or proxy the GitHub calls?

MrSubidubi (approved):
Thanks for the explanation, fine with it for now.

> like plans to cache or proxy the GitHub calls?

The former is very much something I have in my head and want to look into sometime soonish - we cannot enforce all extensions to handle this and thus have to get a bit tricky/hacky sadly, and I think I am starting to get somewhere where I like the direction, so might give it a shot soonish.

MrSubidubi:
Also thanks for linking the issue, will see if I find the time to enter the discussion there.

PR #7385, opened 2026-08-28
Add depsy language server

mpiton:
## Summary

- Add the `depsy` extension (mpiton/zed-depsy at v2.0.0)

Depsy is the extension previously published as `dependi`, renamed at the request of the dependi.io maintainers, who are taking over the `dependi` id with their own extension (mpiton/zed-depsy#377). Same features as before: dependency version hints, vulnerability diagnostics and update code actions for Cargo.toml, package.json, requirements.txt, pyproject.toml, go.mod, composer.json, pubspec.yaml, csproj, Gemfile and pom.xml.

The existing `dependi` entry stays for now. dependi 1.11.0 (#7362) shows a startup notice pointing users here; a follow-up PR will remove `dependi` once users have had time to migrate.

## Type

feat

## Validation

- `pnpm build`
- `pnpm test`
- `pnpm sort-extensions` (no changes)
- `cargo build --release --target wasm32-wasip2 --locked` in `depsy-zed` at the tagged commit

- [x] I've read [the contribution guidelines](https://github.com/zed-industries/extensions/blob/main/CONTRIBUTING.md) and followed the relevant guidance for adding or updating my extension.

PR #7592, opened 2026-09-14
Remove `dependi` extension

mpiton:
This PR removes the `dependi` extension. I'm its author.

`dependi` was renamed to Depsy at the request of the Dependi maintainers (dependi.io), who want the `dependi` id for their own Zed extension. Discussion and agreement: https://github.com/mpiton/zed-depsy/issues/377

- Depsy is already in the registry as `depsy-lsp` (#7385).
- The last `dependi` release, 1.11.0 (#7362), warns users on start-up that the extension moved to Depsy.
- The Dependi team will follow up with a PR that adds their extension under the `dependi` id.

The changes are what `pnpm remove-extension dependi` produces: the `[dependi]` entry in `extensions.toml`, the `extensions/dependi` submodule and its `.gitmodules` section.

Heads-up for the follow-up PR: `dependi` 1.11.0 and earlier versions are already in the blob store, and `package-extensions` skips a version that is already published, so the new extension should start above 1.11.0.

## Validation

- `pnpm build`
- `pnpm test` (119 passed)
- `pnpm sort-extensions` (no changes)


MrSubidubi (approved):
Thanks as well as thanks for being that cooperative with the other people! 🙂

PR #7622, opened 2026-09-16
Update gitignore-templates to v2026.9.1

hjr265:

---

Bumping version as advised here: https://github.com/hjr265/zed-gitignore-templates/pull/1


MrSubidubi (approved):
Thank you for the quick action!

Also, please note that I believe we support multiline-snippets by just having multiple entries in that body list, which might make those a bit more readable whilst editing.

Thanks!

PR #7630, opened 2026-09-16
Add Stvena Live extension

nccapo:
Adds **Stvena Live**, the Zed side of [Stvena](https://github.com/nccapo/stvena) — a terminal tool that runs Codex or Claude Code and lets you accept or reject each change block the agent makes.

- Extension repository: https://github.com/nccapo/zed-stvena (MIT)
- Submodule pinned at `e826fb4`, on `main`

### What it does

The extension itself is a launcher: its only job is to start `stvena editor-lsp --ide zed`, a language server built into the Stvena binary the user already has on `PATH`. Everything the user sees is standard LSP from that server:

| Surface | Mechanism |
| --- | --- |
| `✓ Accept` / `✗ Reject` above a change block | code lens + `workspace/executeCommand` |
| `✎ agent edit`, `👁 read (claude)`, `◆ reviewing` labels | inlay hints |
| One Information diagnostic per unreviewed change | `publishDiagnostics` |
| Review in Stvena, Add to context, Ask the agent, Accept/Reject file, Pause/Resume | code actions, kind `source.stvena` |
| `following · 3 unreviewed · 1 rejection waiting for turn end` | one long-lived `$/progress` item |

It declares no capabilities: it downloads nothing and bundles nothing, it only resolves `stvena` with `worktree.which` and fails with an install hint when it is absent. The server talks to the running Stvena through small JSON descriptors in the project's Git directory; it never writes to the working tree.

### On the extension ID

`stvena` is kebab-safe and contains neither "zed" nor "extension". I did not use the `-language-server` suffix because this is not a language server for a language — it provides no completions, diagnostics from a compiler, or anything language-specific. LSP is simply the only interface Zed exposes that can draw this surface, which is also why the extension registers one server across many languages rather than one language. Happy to rename if you would rather it read differently.

### Testing

Tested as a local extension in **Zed 1.19.2** (macOS, Apple Silicon) at this submodule commit, with `stvena` [v0.4.0-preview.2](https://github.com/nccapo/stvena/releases/tag/v0.4.0-preview.2) on `PATH` — that is the release that carries `editor-lsp`, and the README says so. Zed requests code lenses, inlay hints and code actions, re-requests them after `workspace/codeLens/refresh` and `workspace/inlayHint/refresh`, and renders all of the above.

One honest limitation, documented in the README: Stvena also wants to open the file the agent just touched without taking focus, which needs `window/showDocument`. Zed 1.19 answers `Unrecognized method`, so the server checks the client capability first, says so once in the language server log, and marks the locations in place instead of navigating. It will start navigating with no change here if Zed adds the handler.

Two Zed settings are off by default and gate two of the surfaces — `"code_lens": "on"` and `"inlay_hints": { "enabled": true }` — which the README states up front.


nccapo:
@cla-bot check

PR #7641, opened 2026-09-17
Rename vantadark-theme to slatehollow-theme

ams-sth:
## Summary
- Renamed the extension from Vantadark/Vantabright to Slatehollow (Dark/Light)
- Repository moved to https://github.com/ams-sth/slatehollow
- Replaced the `vantadark-theme` submodule with `slatehollow-theme`, pointed at the latest commit
- Bumped version to 3.0.0 to reflect the rebrand

## Test plan
- [x] Ran `pnpm sort-extensions` to keep `extensions.toml`/`.gitmodules` sorted
- [x] Verified the new submodule resolves and pins to the extension's latest commit


MrSubidubi:
Sorry but will close this unless given a compelling reason - we do not support renaming of extension IDs, since that is a unique ID for your extension you will have to stick with.

Happy to look at a PR for just the renames. Thank you.

ams-sth:
Thanks for the quick review and the pointer makes sense, so I opened #7690 which keeps the id as `vantadark-theme` and only updates the repo URL/version.

PR #7650, opened 2026-09-17
Add crate-helper language server

hereparvezali:
- [+] I've read [the contribution guidelines](https://github.com/zed-industries/extensions/blob/main/CONTRIBUTING.md) and followed the relevant guidance for adding or updating my extension.


hereparvezali:
@cla-bot check

MrSubidubi (changes requested):
Thanks for this!

Please make sure your extension fulfills the [extension publishing prerequisites](https://zed.dev/docs/extensions/publishing/prerequisites). Particularly, you will need to update your extension ID as outlined in the documentation. Once you did the necessary updates, please update everything related here. Thanks! 

Also, `std::env::var` will not work, so please remove it. I think the `which` check should be sufficient anyway, so perhaps it would be best to get rid of the entire manual probing of the .cargo directory IMO.

PR #7656, opened 2026-09-18
Add gin-csv

seahsky:
Adds **gin-csv**: delimited files with per-column colouring, and sort, filter, align, compact and pad as code actions.

Four dialects (comma, tab, pipe, semicolon) are registered as four languages, each with its own tree-sitter parser generated from one shared factory. A Rust language server provides the operations and is downloaded from this repository's releases rather than bundled.

## On overlap with the existing CSV extensions

`csv` and `rainbow-csv` both colour columns. Neither can sort, filter, align, compact or pad, and no CSV language server in any editor implements those — they need a language server, and neither extension has one. That gap is the reason this exists rather than a patch to either.

Two further differences, both checked rather than assumed:

- `rainbow-csv` packs all four delimiters into one parser, so a bare comma inside an unquoted semicolon-delimited cell starts a new colour mid-row. gin-csv compiles one parser per delimiter.
- An empty cell shifts the colours of the cells after it in the published prior art. gin-csv treats an empty field as the absence of a field, so it does not.

The languages are named `Delimited (comma)` and so on, and the grammars are prefixed `gin_csv_`, so nothing collides with either existing extension. I verified this on a machine with all three installed: all register, none overwrites another.

## Testing

Manually tested in Zed 1.20.2 at the submitted commit: all four grammars compile and colour, the language selector switches dialect, sort reorders and undoes in one step, and filter writes and opens a sibling result that is itself a readable file of the same dialect.

The server has 124 tests at the LSP protocol boundary and over the incremental document store, the grammars have 60 tree-sitter corpus cases across the four dialects, and CI runs fmt, clippy and both suites.

## Notes

- MIT, at the extension root.
- The server is fetched from GitHub releases; the `download_file` capability is scoped to this project's own release path.
- Five platforms are published per release: macOS and Linux on both architectures, and Windows x86_64.


seahsky:
@cla-bot check

PR #7665, opened 2026-09-18
Update mach to v0.9.1

octalide:
Bumps the mach extension to v0.9.1 (submodule at cbdde67, tag v0.9.1).

- The extension installs the mach-lsp release whose linked compiler the project's `[project].mach` range accepts, chosen through mach-lsp's RELEASES.json.
- Release archives are extracted by Zed through `download_file`; the extension no longer unpacks them or checks `SHA256SUMS` itself, and no longer depends on `flate2`, `tar`, `zip`, or `sha2`.
- README documents the server's `initialization_options` and how compiler compatibility is reported on `mach.toml`.
- Attribution moves to Briar Systems LLC.


MrSubidubi (changes requested):
Why does this implement its own semver and not use the semver crate? Similarly, why do you implement the unpacking yourself?

octalide:
Thanks for the look.

**Unpacking:** fair point, conceded. The only reason was a `SHA256SUMS` check, which needs the archive bytes that `download_file` with `GzipTar`/`Zip` never exposes. But the sums are published in the same release as the asset, so the check only guarded against a corrupted download. v0.9.1 (now on this PR) extracts through `download_file` and drops `flate2`, `tar`, `zip`, and `sha2`.

**Semver:** this one is deliberate. The ranges come from `mach.toml`, and Mach's range grammar ([manifest.md, "Version ranges"](https://github.com/briar-systems/mach/blob/main/doc/language/manifest.md#version-ranges)) is not Cargo's. The compiler pins it with its own tests, and the `semver` crate diverges on cases that matter for picking a server version:

| range | Mach (missing component is 0) | `semver` crate |
|---|---|---|
| `>1.2` | `>1.2.0`, admits 1.2.1 | `>=1.3.0`, excludes 1.2.1 |
| `<=1.2` | `<=1.2.0`, excludes 1.2.1 | `<1.3.0`, admits 1.2.1 |
| `=1.2` | refused | admits any 1.2.x |
| `1.2`, `*`, `1.*`, `^1 \|\| ^2` | refused | accepted |

Using the crate would install the wrong mls for any range with a partial `>`/`<=` bound. The compiler is written in Mach, so there is no crate to share. `src/semver.rs` is the smallest thing that implements that grammar, with tests against the doc's tables.

MrSubidubi (approved):
Thanks for the explanation, happy to greenlight this now.

PR #7690, opened 2026-09-22
Update vantadark-theme repository URL and version

ams-sth:
Follow-up to #7641, which I closed after @MrSubidubi correctly pointed out that extension IDs shouldn't be renamed.

## Summary
- Extension id stays `vantadark-theme` (unchanged)
- Repository moved to https://github.com/ams-sth/slatehollow (display name is now "Slatehollow", still a dark/light theme pair)
- Updated the `.gitmodules` URL and submodule pin to the new repository's latest commit
- Bumped version to 3.0.0 to reflect the rebrand

## Test plan
- [x] Ran `pnpm sort-extensions` — no further changes needed
- [x] Diff against `main` is limited to the URL, version, and submodule pin (no id/path renames)


MrSubidubi (approved):
This one works for me, thank you!

PR #7706, opened 2026-09-22
Update whim-lang to 0.14.0

azjezz:


azjezz:
I wonder if there's a way to update the grammar without having to update the extension? Almost all updates to whim-lang have been just bumping the version + updating the grammar. for a language that is still evolving, syntax will keep on changing. It would be nice to have a way to keep things up to date?

MrSubidubi (approved):
Cannot do that, as we need to build the grammar here for the appropriate target.

There is a community action linked in our docs you can use to automate the version bumping if that helps.

FWIW, happy to approve these, so no worries from our side. Thanks!

PR #7709, opened 2026-09-22
Update carve to v0.1.3

dereuromark:

Updates the Carve entry and submodule to the released v0.1.3. The registry is still on v0.1.1, so this carries v0.1.2 along with it.

Both releases move highlighting onto the current Tree-sitter grammar: a braced span closes on its own form, a bare marker no longer pairs across a table cell boundary, a forced opener of an open kind stays literal, and the reserved include directive is highlighted by part.


MrSubidubi (approved):
Thanks!

PR #7710, opened 2026-09-22
Add delphi13 extension

Alexl-git:
**Adds `delphi13`: Delphi / Object Pascal and DFM form files**

This adds tree-sitter support for Delphi (`.pas`, `.dpr`, `.dpk`, `.inc`) and for
DFM/FMX form files (`.dfm`, `.fmx`).

**On the existing `pascal` extension.** I want to be upfront that one exists, and
that I have opened an issue on it offering this work rather than going around it.
Two things make this a different submission rather than a duplicate:

1. **DFM/FMX form files have no coverage in the registry at all.** A Delphi form
   is two files -- a `.pas` unit and a `.dfm` component tree -- and the second is
   where the entire UI lives. Nothing currently highlights or outlines them. This
   is the part I would most like to land regardless of what happens to the rest.
2. **This grammar targets Delphi 13 specifically**, including inline variable
   declarations, generic constraints, attributes, class and record helpers,
   anonymous methods, and the `{$IFDEF}`-saturated source that real Delphi
   codebases are made of.

I am entirely happy for this to be resolved by merging the Delphi half into
`pascal` instead, if that is what the maintainers prefer -- the queries and the
grammar are MIT and the offer is already on the table there.

**Measured coverage.** Over the 2,218 `.pas` files of the RAD Studio 37.0
RTL/VCL/FMX/FireDAC source tree, the grammar as this extension uses it -- one
parser, raw source -- parses **2183/2217 = 98.47%** with zero ERROR or MISSING
nodes. The residue is conditional compilation that splits a single statement
across mutually exclusive `{$IFDEF}` arms, which no single-grammar parser can
resolve; paired with a preprocessor (outside Zed's model) the same grammar family
reaches 2217/2217. For DFM: 8,976 text-format form files, 99.18%, with every
failure traced to malformed or truncated input rather than a grammar defect.

The method and corpus are documented at
https://github.com/Alexl-git/tree-sitter-delphi13 so the numbers can be checked.

**Queries provided:** `highlights`, `brackets`, `outline`, `indents`,
`injections` and `textobjects` for both languages. Injections highlight SQL
embedded in string literals and inline `asm` blocks -- both extremely common in
Delphi codebases.

**Grammars:** MIT, forked from `Isopod/tree-sitter-pascal` with attribution
preserved in `tree-sitter.json`; the DFM grammar is original.


Alexl-git:
@cla-bot check

Alexl-git:
@cla-bot check

PR #7712, opened 2026-09-23
Add Ruby snippets

vitallium:
Hi! This PR adds a new extension with code snippets for the Ruby language.
The motivation is quite simple: in https://github.com/zed-extensions/ruby/issues/202 it was requested to disable snippets since most people do not use it. Since Zed does not provide such functionality out of the box, we decided to publish them as a separate extension. I hope this is OK. Thanks!


MrSubidubi (approved):
I am happy to merge this, but would also be happy to talk about whether we want to make this configurable in the settings somewhere - curious about what you think

vitallium:
> I am happy to merge this, but would also be happy to talk about whether we want to make this configurable in the settings somewhere - curious about what you think

I am happy to talk. I think we can merge this PR and then I can work on a proposal of a such setting. WDYT? THanks!

MrSubidubi:
Sounds good, thanks!

PR #7719, opened 2026-09-23
Update cfml to v0.2.29

ghedwards:
Release notes:

https://github.com/cfmleditor/zed-cfml/releases/tag/v0.2.29


MrSubidubi (changes requested):
We are usually not a fan of these long task-commands, as they tend to do more harm than good (and thus are also currently working on a better API for this). Mind giving some input as to how badly needed those are and/or whether we cannot just leave it to users to install the helper binary?

ghedwards:
> We are usually not a fan of these long task-commands, as they tend to do more harm than good (and thus are also currently working on a better API for this). Mind giving some input as to how badly needed those are and/or whether we cannot just leave it to users to install the helper binary?

The binary is the LSP server ( and its internals ), its critical to the extension and not commonly installed by users. Long running tasks are mostly just because processing of some of this information is too costly just to do during startup / open file, etc. Is there something specific you were referring to I could help address ?

PR #7721, opened 2026-09-23
Add Shades of Stoicism theme

ddh4r4m:
Adds **Shades of Stoicism**, a family of ten themes — eight dark, two light.

Repository: https://github.com/ddh4r4m/zed-shades-of-stoicism (MIT)

The themes are generated from one palette system rather than hand-written, so all ten are structurally identical. Two things the generator enforces at build time:

- **Every syntax colour in a theme sits at the same OKLCH lightness**, so a line of code reads as an even band of text instead of pulling the eye toward whichever token happens to be palest. Diagnostics are deliberately exempt.
- **Every colour that carries code clears 4.5:1** against its background, and comments clear 3:1. The build refuses to emit a theme that fails either.

Covers all 46 syntax tokens One Dark sets, plus finer tree-sitter captures, and the full dim/bright ANSI rows.

Tested locally as a dev extension. Validated against the v0.2.0 schema.

A VS Code edition built from the same palettes is [on the marketplace](https://marketplace.visualstudio.com/items?itemName=ddh4r4m.shadesai).


ddh4r4m:
@cla-bot check

MrSubidubi (approved):
Thanks!

